PR #472, opened 2012-08-08
Maven project file for JUnit as OSGi bundle

Tibor17:
Upon discussion #368, I created maven project file.
Other e.g. maven runtime environment, and script running pom.xml may come if necessary.


lonniev:
One year later, Happy New Year, and is there available an OSGi bundle at a public p2 site that has both junit 4 and hamcrest all included? This doesn't have to be bleeding edge recent, anything within ~2 years would probably be fine.

(I can bundle up my own bundle but seek the convenience of leaving this outside the scope of what my project has to actively manage.)

hwellmann:
There are OSGi bundles for JUnit on Maven Central:

    <dependency>
        <groupId>org.ops4j.pax.tipi</groupId>
        <artifactId>org.ops4j.pax.tipi.junit</artifactId>
        <version>4.12.0.1</version>
    </dependency>

    <dependency>
        <groupId>org.ops4j.pax.tipi</groupId>
        <artifactId>org.ops4j.pax.tipi.hamcrest.core</artifactId>
        <version>1.3.0.1</version>
    </dependency>

No p2 support and not hamcrest-all, but maybe that helps.

lonniev:
@hwellmann thanks--but, no, the key need is for the existence of the hamcrest-all packages as an OSGi bundle if only within a public p2 site. For any situation outside that scope, it is trivial to construct a local solution. The point is to avoid having to construct that local solution so that I and my team can remain focused--or, rather, return to a focus--on developing business logic versus maintaining packages for third-party capabilities.

(For the meantime, I took the pragmatic approach of simply adding yet another maven sub-module Tycho eclipse-plugin project to our project to bundle up the necessary hamcrest-all packages and of pushing that to our internal maven and p2 sites. Sigh.)

PR #566, opened 2012-12-05
Enables Inheritance on Category.java by adding @Inherited

gaffa:
As discussed in https://github.com/KentBeck/junit/pull/563 @dsaff suggested to try and add @Inherited. This pull request introduces that behaviour.


bemnet4u:
I know this is an old post but curious if @ayazlakdawala  figured out what is causing this and a fix? I am seeing the same error.

kcooney:
@bemnet4u could you provide a minimal, complete and verifiable example (see https://stackoverflow.com/help/mcve) ?

Also since 4.12 has already been released, you might get more eyes on your problem if you post to http://stackoverflow.com

PR #572, opened 2012-12-12
Ensuring no-generic-type-parms validator called/tested for theories

pholser:


quenio:
It's really unfortunate that 4.12 no longer supports generic params on theory methods.

There are some generic tests, such as `CompareToTheory<T extends Comparable<T>>` that are no longer supported. See: https://www.smartics.eu/smartics-test-utils/

PR #615, opened 2013-01-20
Add option for user to specify runner for parameterized tests

mc1arke:
Test that are run with the Parameterized suite currently default to a specific runner. This change adds a @ParameterizedRunWith annotation that allows a user to specify which runner the suite should use for each Test case so the user can over-ride specific functionality within the runner.


mc1arke:
Closing, as this seems to be possible through other options now, and I have no particular requirement for this feature just now.

PR #763, opened 2013-11-14
Allow custom test runners to create their own TestClasses and customize the scanning of annotations.

motlin:
This is to allow the creation of a test runner that looks for @Test annotations on virtual extension methods in (Java 8) interfaces, as discussed on [this closed pull-request](https://github.com/junit-team/junit/pull/760).

You can see the test runner I built on these extensions [here](https://github.com/motlin/JUnit-Java-8-Runner). It's a work in progress.

### DIFF
--- a/src/main/java/org/junit/runners/ParentRunner.java
+++ b/src/main/java/org/junit/runners/ParentRunner.java
@@ -76,13 +76,17 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
             // do nothing
         }
     };
-    
+
     /**
      * Constructs a new {@code ParentRunner} that will run {@code @TestClass}
      */
     protected ParentRunner(Class<?> testClass) throws InitializationError {
-        fTestClass = new TestClass(testClass);
+        fTestClass = createTestClass(testClass);
         validate();
+    }
+
+    protected TestClass createTestClass(Class<?> testClass) {
+        return new TestClass(testClass);
     }
 
     //
@@ -218,7 +222,7 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
      * </ol>
      * </li>
      * </ol>
-     * 
+     *
      * @return {@code Statement}
      */
     protected Statement classBlock(final RunNotifier notifier) {
--- a/src/main/java/org/junit/runners/model/TestClass.java
+++ b/src/main/java/org/junit/runners/model/TestClass.java
@@ -47,6 +47,13 @@ public class TestClass {
         Map<Class<? extends Annotation>, List<FrameworkField>> fieldsForAnnotations =
                 new LinkedHashMap<Class<? extends Annotation>, List<FrameworkField>>();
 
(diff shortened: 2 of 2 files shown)


kcooney:
@motlin did your custom test runner extend from `BlockJUnit4ClassRunner` or something else? I want to make sure that #1449 doesn't break you.

motlin:
@kcooney Thanks so much for checking. I'm having a bit of trouble following the new commit. If it helps, here is our custom test runner: https://github.com/motlin/eclipse-collections/tree/master/junit-trait-runner/src/main/java/org/eclipse/collections/impl/test/junit

The only customization in the runner is to override the implementation of TestClass. And TestClass is overridden to change the behavior of scanAnnotatedMembers, so that it scans interface (default) methods that are annotated and call addToAnnotationLists on those. If there's a newer way to achieve the same goal, I'm happy to upgrade.

kcooney:
@motlin that will continue to work after I merge that pull.

Here's how someone would write that class in 4.13:

```
public class Java8Runner extends BlockJUnit4ClassRunner {

    /**
     * Creates a Java8Runner to run {@code testClass}.
     * Usually called via reflection.
     *
     * @throws InitializationError if the test class is malformed.
     */
    public Java8Runner(Class<?> testClass) throws InitializationError {
        super(new Java8TestClass(testClass));
    }
}
```

We will be deprecating `ParentRunner.createTestClass()` since it won't be called for subclasses of `BlockJUnit4ClassRunner` that use the constructor that takes in a `TestClass`. The new way is cleaner (and probably what we should have done to resolve your need in the first place),

motlin:
@kcooney Sounds good, thanks.

I still feel like our test runner's behavior would be good default behavior in JUnit 4. I haven't tried to promote our test runner and I realize this style of testing isn't common. But there's also little downside. The extra reflection code won't take much time since most tests don't implement any interfaces.

We have a big and growing test suite [here](https://github.com/eclipse/eclipse-collections/tree/master/unit-tests-java8/src/test/java/org/eclipse/collections/test)

One nice example that shows the benefit is [UnmodifiableListTestCase](https://github.com/eclipse/eclipse-collections/blob/master/unit-tests-java8/src/test/java/org/eclipse/collections/test/list/UnmodifiableListTestCase.java) which inherits the tests for unmodifiable collections and for lists.

The concrete tests that are classes are always small and usually implement one method which instantiates the collection under test, like [UnmodifiableMutableListTest](https://github.com/eclipse/eclipse-collections/blob/master/unit-tests-java8/src/test/java/org/eclipse/collections/test/list/mutable/UnmodifiableMutableListTest.java).

Many, many tests live at the top of our trait hierarchy in [RichIterableTestCase](https://github.com/eclipse/eclipse-collections/blob/master/unit-tests-java8/src/test/java/org/eclipse/collections/test/RichIterableTestCase.java).

kcooney:
@motlin JUnit 5 will support defining tests in default methods.

Issue #727, opened 2013-08-23
Timeout exceptions should have stack trace from stuck thread


chrisvest:
It would be very helpful if tests with timeouts (I tend to use `@Test(timeout = ...)`) failed with exceptions that had a stack trace from the thread that was stuck or slow.

If a reference to the executing thread is saved (storing this information somewhere could be the first thing it does), then `Thread.getStackTrace()` will be available for inspection. The exception itself, however, would need to be special, because exceptions don't normally work with a _saved_ array of stack frames, but instead appears to call into native code to generate the stack frame array every time it is needed.

Issue #463, opened 2012-07-16
Add ability for timeout to print full thread stack dump


toddlipcon:
When using JUnit to run integration-style tests which involve several threads, or to test multi-threaded components, a deadlock or livelock bug can often result in a test timeout. The current behavior only provides the stack trace of the main test thread, which is not always sufficient to understand the true cause of the deadlock (the main thread may just be blocked on another worker thread).

I'd like to add the ability to dump all of the thread stacks to stderr when a timeout is triggered. Does this seem like a reasonable default behavior, or something best enabled via some kind of property/configuration?

junit-builds:
JUnit 4 is now in maintenance mode.

At this point, only critical bugs and security issues will be fixed.

The team has therefore decided to close this issue.

PR #768, opened 2013-11-22
#463: Add ability for timeout to print full thread stack dump

reinholdfuereder:
Although originally this started based on #727: Fail on timeout displays stack of stuck thread …

This is a first naive attempt that (a) optionally adds information from ThreadMXBean#findMonitorDeadlockedThreads and (b) always adds full thread dump to exception message

---

@dsaff and #727: Since the Timeout#createTimeoutException is now (?) in FailOnTimeout#createTimeoutException (statement) it would require to (1) subclass FailOnTimeout to override FailOnTimeout#createTimeoutException and (2) subclass Timeout to override Timeout#apply to create and return an instance of the new FailOnTimeout sublcass from before. I had hoped for more convenience...

Besides that I also stumbled over the fact that quite some interesting thread dump information (locks, monitors, synchronizers) can only be retrieved from ThreadMXBean from Java >= 1.6.

Thus, what do you think of the following naive and admittedly very ugly thoughts, when I want to have this extended thread dump?
* Is any poor mans plugin/extension attempt via an non-elegant reflection based class creation (e.g. there is a system property with the full qualified name of the dumping class) thinkable?
** => actually this is roughly implemented in #770
* Another ugly idea would be making JUnit a multi-project with an optional 1.6 compatible part, where this extended thread dump class can be placed. Of course this then could be packaged in whatever ugly style in a single JAR and again via system property + reflection plugged in... But this gets extremely dirty...

### DIFF
--- a/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
+++ b/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
@@ -1,8 +1,12 @@
 package org.junit.internal.runners.statements;
 
 import java.lang.management.ManagementFactory;
+import java.lang.management.ThreadInfo;
 import java.lang.management.ThreadMXBean;
-import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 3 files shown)


kcooney:

PR #770, opened 2013-11-22
Full thread stack dump on timeout with custom handler

reinholdfuereder:
This is an attempt to start addressing: #463 Add ability for timeout to print full thread stack dump
- Another attempt that allows to "install" an additional custom timeout handler that can optionally also add a further exception to the test timeout exception

=> Extending the initial rudimentary attempt of #768 to allow "installing" an additional custom timeout handler (that can optionally also add a further exception to the test timeout exception) by a globally set custom timeout handler based on a system property, that is instantiated via reflection.

This custom timeout handler is also desirable to overcome the Java 1.5 compatibility requirement of JUnit, so any default timeout handler bundled with JUnit needs to adhere to this compatibility too: but quite some interesting thread dump information (locks, monitors, synchronizers) can only be retrieved from ThreadMXBean from Java >= 1.6, I think.

(Note one self test still fails in Ant context, but not in Eclipse though -- why is that the case? can it be the parallel usage of a test case timeout via annotation attribute and the global test class timeout via rule kind of interfere? but why is it then working in Eclipse?)

Deadlocked threads of self tests are not surviving anymore (fixed in a subsequent commit; in contrast to some infinite loop running threads from other self tests...
=> In order to fix that and allow full thread stack dump in custom handler, the interrupting of the test thread was moved slightly down.

What do you think of that approach?

### DIFF
--- a/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
+++ b/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
@@ -10,6 +10,7 @@ import java.util.concurrent.FutureTask;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
+import org.junit.rules.Timeout;
 import org.junit.rules.TimeoutHandler;
 import org.junit.runners.model.MultipleFailureException;
 import org.junit.runners.model.Statement;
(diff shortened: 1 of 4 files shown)


kcooney:

PR #778, opened 2013-12-02
Add expectRootCause to expected exception rule, fixes #714

### LINKED ISSUE #714: Introduce ExpectedException.expectRootCause
Right now you can only check/expect the direct cause of an exception that is thrown in a test method.
I have an arquillian test where I need to check the root(!) cause (a ConstraintViolationException) which "is on 4th level" (wrapped three times by the container).

So an new Method "public void expectRootCause(Matcher<? extends Throwable> expectedCause) {" would come in handy!

pimterry:
Adds an expectRootCause test, which tests a matcher against the root cause of the thrown exception, rather than just the first-level contained cause matched by expectCause.

I've also updated the expectCause docs & errors to refer to 'immediate cause', to try and differentiate the two cases.

### DIFF
--- a/src/main/java/org/junit/internal/matchers/ThrowableCauseMatcher.java
+++ b/src/main/java/org/junit/internal/matchers/ThrowableCauseMatcher.java
@@ -15,7 +15,7 @@ public class ThrowableCauseMatcher<T extends Throwable> extends
     }
 
     public void describeTo(Description description) {
-        description.appendText("exception with cause ");
+        description.appendText("exception with immediate cause ");
         description.appendDescriptionOf(fMatcher);
     }
 
@@ -26,7 +26,7 @@ public class ThrowableCauseMatcher<T extends Throwable> extends
 
     @Override
     protected void describeMismatchSafely(T item, Description description) {
-        description.appendText("cause ");
+        description.appendText("immediate cause ");
         fMatcher.describeMismatch(item.getCause(), description);
     }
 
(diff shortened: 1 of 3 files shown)


JavaWebinar:
Are there any workaround for root cause available?

kcooney:
@JavaWebinar you can write your own `expectRootCause()` method.

JavaWebinar:
@kcooney Thanks! 
Until JUnit 13 I have done it in old manner:
```
        try {
            ...
            Assert.fail("Expected " + exceptionClass.getName());
        } catch (Exception e) {
            Assert.assertThat(getRootCause(e), instanceOf(exceptionClass));
        }
```

TWiStErRob:
@kcooney `expectThrows` is not much different than what we already had in 4.12: via `ExpectedException.expect(Matcher)` we could do any assertion we wanted.

IMO the problem here is that we have to write (copy-paste) the while loop (or matcher) in every single project we use JUnit in instead of quickly writing `expect(hasRootCause(instanceOf(blah.class)))`. Maybe this matcher needs to be put in hamcrest?

panchenko:
One could use a library for this, e.g.
* [guava](https://google.github.io/guava/releases/23.0/api/docs/com/google/common/base/Throwables.html#getRootCause-java.lang.Throwable-)
* [commons-lang](https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/exception/ExceptionUtils.html#getRootCause(java.lang.Throwable))

kcooney:
@TWiStErRob `expectThrows` is very different. You specify what code is expected to throw via a lambda, so 1) it is more clear and 2) you can make other assertions afterwards (ex: verify the state of the subject under test wasn't changed by the method that threw the exception).

I find that assertion frameworks that provide a DSL are better fits for doing assertions like this. For instance,  Truth has a good API for making assertions on throwables.

PR #869, opened 2014-04-08
JUnit4TestAdapter with a class with no test methods should result in a Test with no children

clumsy:
wrapping several initialisation errors into one
not triggering notifier in case test is not run

Signed-off-by: Alexander Jipa <alexander.jipa@gmail.com>


kcooney:
(doing a pass at closing old issues)

Given 1) JUnit 4-style tests were introduced a long time ago 2) JUnit 5 is around the corner and 3) this only affects an error case (a test case with no tests), I'm not sure it's worth fixing this

PR #882, opened 2014-04-22
Extending and Modifying sorting feature of JUnit4.

fshepherd:
Adding SortWith Annotation to JUnit4 such that test methods defined in a
test class or test classes defined in a test suite can be executed in a
predefined order.

There are four sorting methods available currently for choosing. They are
default, jvm, name ascending and random. First three methods works exactly
as same as their ancestors defined in MethodSorter class. Random sorting
method is added due to the request of adding randomness to tests
execution. A test class annotated with random sorting method executes its
test methods in a random order. A test suites annotated with random
sorting method executes its test classes in a random order.

User still could specify sorting method manually by designating a sorting
request. Also the manually specified sorting method will override all
declarative sorting method defined previously.

Currently I disabled MethodSorter class temporarily because I think if
this patch was accepted, it would no longer be useful.

BTW this is my first attempt to be part of an open source project hence I
might omit something important. Please spare me in advance if I did
something wrong.

### DIFF
--- a/src/main/java/junit/framework/JUnit4TestAdapter.java
+++ b/src/main/java/junit/framework/JUnit4TestAdapter.java
@@ -81,6 +81,7 @@ public class JUnit4TestAdapter implements Test, Filterable, Sortable, Describabl
     }
 
     public void sort(Sorter sorter) {
-        sorter.apply(fRunner);
+        if(sorter!=null)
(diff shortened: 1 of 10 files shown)


kcooney:
I sent out pull #1130 which adds `@SortWith`.

PR #905, opened 2014-05-03
junit as osgi bundle

Tibor17:
This PR
+ generates new osgi manifest headers
+ Maven build uses the same Java Hamcrest dependency as before
+ manifest implementation entries moved from maven-jar-plugin to maven-bundle-plugin
+ added integration tests which say that the osgi bundle has correct configuration


Tibor17:
I guess we were waiting quite long for this PR.
New release 5.0 with major version change sounds like the last chance.

marcphilipp:
Last chance? Why's that?

marcphilipp:
@Tibor17 Have you seen @kcooney's question above: https://github.com/junit-team/junit/pull/905#discussion_r12327207? I have the same concern. Maybe we should export all packages and don't have any private packages at all.

Tibor17:
@marcphilipp 
Thx for remiding me, I'll answer his question.

pom.xml
```diff
@@ -211,6 +218,44 @@
...
+                        <Bundle-SymbolicName>org.${project.artifactId}</Bundle-SymbolicName>
+                        <Bundle-Name>${project.name}</Bundle-Name>
+                        <Bundle-Description>${project.description}</Bundle-Description>
+                        <Bundle-Vendor>${project.organization.name}</Bundle-Vendor>
+                        <Bundle-DocURL>${project.organization.url}</Bundle-DocURL>
+                        <Bundle-Version>${project.version}</Bundle-Version>
+                        <Bundle-License>Eclipse Public License 1.0</Bundle-License>
+                        <Export-Package>
+                            !org.junit.internal.*,
+                            !org.junit.experimental.theories.internal.*,
```
Tibor17: OSGI works on package level - not class level.
So if any class like this has internal purposes, you already added in "internal" package.
Then internal packages is design problem because such internal classes have to be public anyway and therefore accessible on non-osgi based applications.

kcooney: Yes, it's rather unfortunate that we had public APIs that used internal classes.

At the very least, I don't think we can exclude org.junit.external

Tibor17: Hm, `org.junit.external`? There's no such package, see pom.xml again.
You mean maybe `org.junit.experimental.theories.internal`.

kcooney: sorry, typo. I meant we should include `org.junit.internal` because the classes under there are often used externally. I don't have a strong opinion about `org.junit.experimental.theories.internal`

Tibor17: We should make all packages exported and remove private-packages. The reason is that 
+ `org.junit.experimental.theories.internal.SpecificDataPointsSupplier` is used in public `@FromDataPoints`.
+ `org.junit.experimental.theories.internal.Assignments` is used in protected methods in `Theories`
Not nice, but has to be done :(

marcphilipp:
@Tibor17 Do you have time to "make all packages exported and remove private-packages"?

Tibor17:
@marcphilipp 
Yes I will make it, don't worry.

Tibor17:
The branch `junit5` is already on JDK6 supporting SPI.
I am glad to remove my internal Service Loader (SPI) in this PR and use the default one from JDK.
Are we ready to move this PR to `junit5` branch?

Tibor17:
What about to set Milestone JUNIT 5?
OSGi frameworks need Java 6 SPI and we need the frameworks in our unit tests.

kcooney:

I propose we close this. Java has decided to go with Jigsaw as its modularization solution. If the OSGi community still wants OSGi bundles, they can continue to create and release them outside of JUnit

marcphilipp:
Agreed.

Issue #80, opened 2010-02-09
Add SPI support (Service Provider Interface, from Java 6) to enable external RunListener


rliesenfeld:
The @RunWith annotation has been widely successful in enabling the integration of other testing tools with JUnit. For example, the following testing libraries (among others) use it: jMock, Mockito, JMockit, PowerMock, Unitils, Spring Test.

However, this mechanism requires the annotation be applied to each and every test class, and does not support the use of more than one custom runner for the same test class.

I would like to propose a simple enhancement to JUnit's ability to integrate with other testing tools: add support for the automatic discovery of org.junit.runner.notification.RunListener implementations, through the standard Service Provider API available in Java SE 6 (the java.util.ServiceLoader class).

I am the creator of the JMockit testing toolking (http://code.google.com/p/jmockit), which already implements sophisticated integration with the JUnit test runner. However, to provide the best user experience I need JMockit to get initialized before JUnit starts running the first test. I have achieved this, but not in a clean way. The mechanism I propose would provide such a clean solution, which I believe might also be useful to other testing libraries which integrate with JUnit.

The desired SPI support can be achieved with the addition of the following four lines of code to the constructor of org.junit.runner.JUnitCore (alternatively, it could be done inside the RunNotifier class):

```
if ("1.6 1.7".contains(System.getProperty("java.specification.version"))) {
    ServiceLoader<RunListener> serviceLoader = ServiceLoader.load(RunListener.class);
    for (RunListener listener : serviceLoader)
        fNotifier.addListener(listener);
}
```

The code above requires Java 6, but it would not interfere with test suites compiled for and running under Java 5 (or older).

Thanks

kriegaex:
Just found this old ticket by chance and can only say how appalled I am to find out how long ago such a simple, universally useful feature which is now marketed as a great achievement in JUnit 5 has not just been ignored but actively blocked with poor pseudo arguments by people who should know better.

Given the still big user base of JUnit 4, I would still be happy to see this backported into 4.13 (or 4.14 then) before really freezing development.

Sorry for being opinionated, but I have reasons for my opinion.

junit-builds:
JUnit 4 is now in maintenance mode.

At this point, only critical bugs and security issues will be fixed.

The team has therefore decided to close this issue.

PR #923, opened 2014-05-30

baev:


src/main/java/org/junit/internal/ServiceLoaderWrapper.java
```diff
@@ -0,0 +1,73 @@
...
+    private boolean checkHasNextSafely(Iterator iterator) {
+        try {
+            /* Throw a ServiceConfigurationError if a provider-configuration file violates the specified format,
+            or if it names a provider class that cannot be found and instantiated, or if the result of
+            instantiating the class is not assignable to the service type, or if any other kind of exception
+            or error is thrown as the next provider is located and instantiated.
+            @see http://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html#iterator()
+            */
+            return iterator.hasNext();
+        } catch (Exception e) {
```
baev: Also we should catch `ServiceConfigurationError` because it is instance of `Error`

baev:
oops, I forgot about this one and lose it via force push. So there are commit (f54758fecef54eb08b61d3c1abaa3e9cc87a7339), i'll send new PR soon

PR #949, opened 2014-07-07
support for assertEquals([String message, ] CharSequence, CharSequence)

vorburger:
This contribution fixes a problem I've run into when using assertEquals on two CharSequence.

It basically already works, of course - BUT one is left wondering why e.g. the Eclipse (and presumably other IDE as well) JUnit view's Result Comparison pop-up does not work on double-clicking a failure.. this is because, currently, assertEquals on two CharSequence throws only a generic AssertionFailedError and not the ComparisonFailure expected by such UIs.

PS: As the pom.xml specifies jdk.version 1.5 and <compilerVersion>1.5</compilerVersion>,  I was assuming, hopefully correctly, that introducing a @since 1.4 CharSequence was not a problem for backward compat.

### DIFF
--- a/src/main/java/junit/framework/Assert.java
+++ b/src/main/java/junit/framework/Assert.java
@@ -112,9 +112,8 @@ public class Assert {
      * 
      * @author Michael Vorburger
      */
-    @SuppressWarnings("deprecation")
     public static void assertEquals(String message, CharSequence expected, CharSequence actual) {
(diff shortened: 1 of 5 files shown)


kcooney:

Since we didn't hear back from @vorburger I'm closing this. If someone feels passionate about the issue, feel free to send a new pull that addresses the issues.

PR #969, opened 2014-07-30
handle the {index} parameter like a native MessageFormat argument.

melchisedek:
For example that allows to specified leading zero by using
{index,number,0000}


adrianosimoes:
Should the conflicts be fixed, or maybe it's too late and we should just close this pull request?

PR #1004, opened 2014-10-13
SortWith annotation for TestClass with default implementations

SpOOnman:
This pull request provides @SortWith annotation for TestClasses. It takes a Class as a value and uses it to sort methods within test class. It allows users to write their own implementations. Three built-in implementations are available: default, name ascending and JVM. @SortWith replaces @FixMethodOrder annotation.

I've decided to write a new pull request like @kcooney pointed out in #998. There is another pull request with @SortWith: #882, but it went too far and it is stuck with no ongoing work. I've read all comments and I hope this implementation would be accepted.


marcphilipp:
@kcooney Can this be closed for now?

kcooney:
Yes.

kcooney:
I sent out pull #1130 which adds `@SortWith`.

Issue #669, opened 2013-04-25
Cutting on stack call


stanl3y:
This is what I get when I try to run my tests in terminal. Is it possible to cut the stack call to say, three or four lines from the top ?

There was 1 failure:
1) matrixToString(GeneralMatrixTest)
org.junit.ComparisonFailure: expected:<[1.000 
2.000 ]> but was:<[ 1.000
 2.000
]>
    at org.junit.Assert.assertEquals(Assert.java:115)
    at org.junit.Assert.assertEquals(Assert.java:144)
    at GeneralMatrixTest.matrixToString(GeneralMatrixTest.java:52)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:616)
    at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
    at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
    at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
    at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
    at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
    at org.junit.rules.ExpectedException$ExpectedExceptionStatement.evaluate(ExpectedException.java:168)
    at org.junit.rules.RunRules.evaluate(RunRules.java:20)
    at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
    at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
    at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:50)
    at org.junit.runners.ParentRunner$3.run(ParentRunner.java:238)
    at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:63)
    at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:236)
    at org.junit.runners.ParentRunner.access$000(ParentRunner.java:53)
[...]

PR #1018, opened 2014-11-02
#669 Cutting on stack call

rochapaulo:
The class Packages represents the packages that must be filtered when the stacktrace will be printed, an instance of packages are passed to the Tracefilter, that do the filtering....

- If the user pass -filterstack argument, the stack will be filtered with some "default packages".
- If they pass -filterstack [pkg1, pkg2, pkg3], the pkg1, pkg2, pkg3 will be filtered from the trace (-filterstack [java.lang, junit.framework, ...]

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -348,7 +348,6 @@
                     <reportSet>
                         <reports>
                             <report>index</report>
-                            <report>summary</report>
                             <report>dependency-info</report>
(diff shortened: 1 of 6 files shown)


kcooney:

I'm closing this. It only affects JUnit3-style tests run via the JUnit3 main method.

PR #1047, opened 2014-12-18
Remove deprecated Hamcrest matchers

marcphilipp:
These matchers have been deprecated since 4.11 which was released more than two years ago.


kcooney:
Merged into junit5 branch

PR #1048, opened 2014-12-18
Remove deprecated JUnitSystem.exit() method

marcphilipp:
Do we need to wait two years to remove this?

### DIFF
--- a/src/main/java/org/junit/internal/JUnitSystem.java
+++ b/src/main/java/org/junit/internal/JUnitSystem.java
@@ -4,11 +4,6 @@ import java.io.PrintStream;
 
 public interface JUnitSystem {
 
-    /**
-     * Will be removed in the next major release
-     */
-    @Deprecated
-    void exit(int code);
-
     PrintStream out();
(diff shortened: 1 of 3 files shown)


kcooney:
LGTM

@junit-team/junit-committers any objections?

BTW, I think having two maintainers agree to merge something is generally sufficient (so if one maintainer makes a pull, another can merge it). For removing APIs, however, I would prefer if we have at least three maintainers agree to merge.

marcphilipp:
:+1: I agree to the 3/4 guideline. I had the same gut feeling.

dsaff:
Looks good to me!

stefanbirkner:
LGTM. And I agree to the 2/4 and 3/4 guidelines.

marcphilipp:
This was deprecated in 4.12. According to our discussed deprecation policy we have to wait at least a year to remove it, haven't we?

kcooney:
Yes, we should wait to remove it. Let's close this for now.

PR #1049, opened 2014-12-18
Delete JUnit4ClassRunner and its dependencies

marcphilipp:
`JUnit4ClassRunner` was deprecated in JUnit 4.3 (six years and 10 major releases ago) with a comment saying "This may disappear as soon as 1 April 2009". We started having some problems with running the tests in JDK 7, and we decided to delete the class and its support classes. Although we try very hard to  
`JUnit4ClassRunner` didn't support `Rule`s, it wasn't designed to be extensible, and it was in an internal package. Please use `BlockJUnit4ClassRunner` instead.


kcooney:
Merged into junit5 branch

PR #1055, opened 2014-12-29
A new method to expect a thrown exception

seanjreilly:
There are generally two techniques for testing that a method throws an exception in a certain scenario.

The original technique is to use a try catch block:

```java
@Test
public void someTest() throws Exception {
    //some setup here
    try {
        doSomethingThatThrows();
        fail("expected a NullPointerException");
    } catch (NullPointerException e) {
        //optionally assert something about the exception
    }
}
```

This method is generally not preferred because it is verbose, and error-prone. Forgetting the `fail` statement is a really subtle failure case that can be easy to miss.

The newest method is to use the `expected` element of the `@Test` annotation:

```java
@Test(expected=NullPointerException.class)
public void someTest() throws Exception {
    //some setup here

    doSomethingThatThrows();
}
```

This method is less verbose, and much less complicated. However, it also has two significant drawbacks that are easy to overlook:

1. You cannot make any assertions about the exception once you encounter it.
2. If the specified exception is thrown *anywhere* in the test, the test will pass, even if the test isn't thrown from the expected statement.

In the above example, if a NullPointerException is thrown in the ```//some setup here`` block, the test will still pass!

This patch adds an additional way of testing for thrown exceptions:

```java
@Test(expected=NullPointerException.class)
public void someTest() throws Exception {
    //some setup here

    NullPointerException e = expect(NullPointerException.class, () -> doSomethingThatThrows());

    //optionally assert something about the exception
}
```

This mitigates the weaknesses of using the `expected` element of the `@Test` annotation: the exception must be thrown by calling `doSomethingThatThrows()`, or the test will fail; and the test can make assertions about the NullPointerException once it occurs.

This patch has been coded in such a way that it will compile and work with anonymous interface implementations in Java 5 or higher, although Java 8 lambda expressions are needed for the syntax to really be an improvement over the original pattern.

If no exceptions is thrown by the provided code block, then an `AssertionError` is thrown. If an exception other than the expected exception is thrown, then it will be rethrown by the `expect` method, which will cause the test to fail.

### DIFF
--- a/src/main/java/org/junit/ExceptionSupplier.java
+++ b/src/main/java/org/junit/ExceptionSupplier.java
@@ -1,0 +1,13 @@
+package org.junit;
+
+/**
+ * A functional interface used to call code that is expected to throw an exception
+ * @param <T> The type of exception expected to be thrown by {@link ExceptionSupplier#get()}
+ */
+public interface ExceptionSupplier<T extends Exception> {
+    /**
+     * Execute code that is expected to throw an exception
+     * @throws T the type of exception expected to be thrown
+     */
+    void get() throws T, Exception;
+}
--- a/src/main/java/org/junit/ExceptionSupplier.java
+++ b/src/main/java/org/junit/ExceptionSupplier.java
@@ -4,7 +4,7 @@ package org.junit;
  * A functional interface used to call code that is expected to throw an exception
  * @param <T> The type of exception expected to be thrown by {@link ExceptionSupplier#get()}
  */
-public interface ExceptionSupplier<T extends Exception> {
+public interface ExceptionSupplier<T extends Throwable> {
     /**
      * Execute code that is expected to throw an exception
      * @throws T the type of exception expected to be thrown
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -991,14 +991,16 @@ public class Assert {
      * @return the expected exception, if it is thrown
      * @throws Exception any exception <i>other</i> than the expected exception
      */
-    public static <T extends Exception> T expect(Class<T> exceptionClass, ExceptionSupplier<T> supplier) throws Exception {
+    public static <T extends Throwable> T expect(Class<T> exceptionClass, ExceptionSupplier<T> supplier) throws Exception {
         try {
             supplier.get();
-        } catch (Exception e) {
-            if (exceptionClass.isInstance(e)) {
-                return exceptionClass.cast(e);
+        } catch (Throwable t) {
+            if (exceptionClass.isInstance(t)) {
+                return exceptionClass.cast(t);
             }
-            throw e;
+            AssertionError error = new AssertionError("Expected a " + exceptionClass.getName() + " to be thrown, but caught a " + t.getClass().getName() + " instead");
+            error.initCause(t);
+            throw error;
         }
         throw new AssertionError("Expected a " + exceptionClass.getName() + " to be thrown");
     }
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -955,4 +955,51 @@ public class Assert {
             Matcher<? super T> matcher) {
         MatcherAssert.assertThat(reason, actual, matcher);
     }
+
+    /**
+     * Executes a code block, expecting an exception to be thrown.
+     *
+     * If the expected exception is thrown, it is caught and returned.
+     * If an exception is not thrown, an AssertionError will be thrown.
+     * If an unexpected exception is thrown, it will not be caught.
+     *
+     * This is a Java 8 convenience method that encapsulates this pattern:
+     *
+     * <pre>
+     *     try {
+     *         doSomething();
+     *         fail("expected an Exception")
+     *     } catch (Exception e) {
+     *         //assert something about the exception
+     *     }
+     * </pre>
+     *
+     * and replaces it with:
+     *
+     * <pre>
+     *     Exception e = expect(Exception.class, () -&gt; doSomething());
+     *     ///assert something about the exception
+     * </pre>
+     *
+     * This is sometimes preferable to using <pre>@Test(expected=...)</pre>, because
+     * the exact line that should throw the exception is obvious, and you can make
+     * assertions about the encountered exception (has a certain cause or a certain message, etc).
+     *
+     * @param exceptionClass the expected exception class
+     * @param supplier a lambda expression (or anonymous interface implementation) that is expected to throw an Exception
+     * @param <T> The expected type of exception
+     * @return the expected exception, if it is thrown
+     * @throws Exception any exception <i>other</i> than the expected exception
+     */
+    public static <T extends Exception> T expect(Class<T> exceptionClass, ExceptionSupplier<T> supplier) throws Exception {
+        try {
+            supplier.get();
+        } catch (Exception e) {
+            if (exceptionClass.isInstance(e)) {
+                return exceptionClass.cast(e);
+            }
+            throw e;
+        }
+        throw new AssertionError("Expected a " + exceptionClass.getName() + " to be thrown");
+    }
 }
--- a/src/test/java/org/junit/tests/assertion/AssertionTest.java
+++ b/src/test/java/org/junit/tests/assertion/AssertionTest.java
@@ -694,8 +694,56 @@ public class AssertionTest {
                 }
             });
             fail("the unexpected exception should be thrown");
-        } catch(NullPointerException e) {
-            assertSame(testException, e);
+        } catch(AssertionError e) {
+            assertEquals("Expected a java.sql.SQLException to be thrown, but caught a java.lang.NullPointerException instead", e.getMessage());
+            assertSame(testException, e.getCause());
+        }
+    }
+
+    @Test
+    public void expectWithACodeBlockThatThrowsTheCorrectError() throws Exception {
+        final Error expectedError = new Error("The exception");
+        ExceptionSupplier<Error> supplier = new ExceptionSupplier<Error>() {
+            public void get() throws Exception {
+                throw expectedError;
+            }
+        };
+
+        Error actual = Assert.expect(Error.class, supplier);
+
+        assertSame(expectedError, actual);
+    }
+
+    @Test
+    public void expectWithACodeBlockThatDoesNotThrowAnError() throws Exception {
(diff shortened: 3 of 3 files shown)


src/main/java/org/junit/ExceptionSupplier.java
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * A functional interface used to call code that is expected to throw an exception
+ * @param <T> The type of exception expected to be thrown by {@link ExceptionSupplier#get()}
+ */
+public interface ExceptionSupplier<T extends Exception> {
+    /**
+     * Execute code that is expected to throw an exception
+     * @throws T the type of exception expected to be thrown
+     */
+    void get() throws T, Exception;
```
marcphilipp: Does this method have to declare both `T` and `Exception`?

marcphilipp:
@seanjreilly I like this idea a lot! It's a small method. Having it in JUnit instead of another library would be beneficial to a lot more developers. 

BTW ScalaTest has the exact same thing:

```scala
val thrown = intercept[Exception] {
  pizza.boom
}
assert(thrown.getMessage === "Boom!")
```

Source: http://alvinalexander.com/scala/scalatest-testing-expected-exceptions-intercept-testfailedexception

I'm not sure which name is better: `expect` or `intercept`, though. I'm also not sure if `Assert` is the best location to put it.

kcooney:
I like `expect` better than `intercept`, but perhaps because it just sounds familiar to me used in `Mockito`.

If we think we've seen enough implementations of this in a wild to come up with a good API, then I have no objections. I personally like the idea of returning the exception thrown, so users can use their favorite assertion framework (vs us spending a lot of time building a DSL for assertions on exceptions). Don't get me wrong, I love a well designed DSL, but I'd rather leave it to the maintainers of Fest, Hamcrest and Truth to come up with APIs that work well with their APIs and design philosophies.

stefanbirkner:
I think there's a common understanding on how to test for exceptions in Java 8. It needs a method that takes a statement (arbitrary piece of code) as input and returns the exception that is thrown by that statement. Differences between the implementations are:

* Return a Throwable or return an exception whose type is specified by the caller.
* Throw an AssertionError or return null if the statment doesn't throw an exception.
* The method's name.

All solutions have in common that they allocate the verification of the exception to an assertion library.

I agree with @marcphillip that it is benefical to a lot of developers if such an method is part of JUnit, but I propose a second way of doing this. Let us create an independent library junit-exception for this feature and add it as dependency to JUnit. This has some benefits:

* It can be used by TestNG users, too (without having a dependency to JUnit itself).
* It can be released independently from JUnit.

kcooney:
A junit-exception library with only one or two methods doesn't seem all that useful.

As far as the options, I prefer to return an an exception whose type is specified by the caller, throw an AssertionError if the statement doesn't throw an exception (or throws a different exception) and call the name `expect` or `expectException`.

Tibor17:
@seanjreilly 
We were struggling with junit committers quite long as we wanted to get a pure assertion method into well known tramework without rellying on vallado, etc. As a result I will definitely use my private because every time the PR discussion started and was quite positive, nothing was propagated in master.

marcphilipp:
@seanjreilly @Tibor17 We, the maintainers, are about to start an Java 8 based JUnit add-on in the upcoming months. Current working title is "JUnit Lambda". We'll post more about it soon.

Having said that, I think "JUnit Lambda" would be the place this method belongs. To use it you want to be able to use lambdas. Thus, you'll need Java 8.

kcooney:
@Tibor17 14 days isn't quite long. Are you referring to a different pull request?

In any case, if we seem a bit conservative in changing JUnit, it's because once we add a public API we likely need to support it forever. Getting the API right sometimes requires some time.

NarendraPathai:
@kcooney I was working separately on this. Does something like this make sense? It utilizes `Runnable` and `Supplier` interfaces.
```java
public class Test {
  @Rule public AssertableExpectedException exception = AssertableExpectedException.none();

  @Test
  public void test() {
      exception.expect(Exception.class)
            .when(() -> {classUnderTest.methodThrowingException();})
            .thenAssertThat(() -> {return classUnderState.getState();},
                    CoreMatchers.equalTo(-1)); 
  }
}

Tibor17:
@marcphilipp 
In that case we should look at what @lukaseder  said and use @ `FunctionalInterface` and `Consumer`.
We have create several PRs for this, so I prefer continuing in our previous google/gmail discussion with several concepts for Java 8.

dsaff:
Just to further stir the pot, what about this class?

public static class Result<T> {
  T success;
  Throwable failure;

  public static Result<T> of(Callable<T> callable) {...}

  public <T extends Exception> T asFailure(Class<T> type) {...}
}

With some 

Then with a convenience method on Result, the top exception becomes:

public void someTest() throws Exception {
    //some setup here

    NullPointerException e = Result.of(() -> doSomethingThatThrows()).asFailure(NullPointerException.class);

    //optionally assert something about the exception
}

I could see this going into JUnit 5: I like that it doesn't require extensions to existing classes, and also doesn't need static imports to make sense.

Tibor17:
The `of` method on static import does not say that much about the arg.
I think we should use onlt built-in functiona interfaces from Java 8, Better is
`Object = when( () -> {} ).asReturn()`
`Throwable t = when( () -> {} ).asThrows()`
`assertThat( t, instanceOf( NPE.class ) )`
Post assertion is advantage because we don't need to implement whatever kinds of assertion combinations the user wants, simply `t` can be asserted afterwards - I don't care how - not part of the framework.

dsaff:
@Tibor17, I'm a little confused.  Callable is built-in (java.util.concurrent.Callable).  Are you preferring a different built-in that was added in Java 8?

Tibor17:
@dsaff 
`Built-in Functional Interfaces` http://winterbe.com/posts/2014/03/16/java-8-tutorial/

dsaff:
Right.  From that post:

"The JDK 1.8 API contains many built-in functional interfaces. Some of them are well known from older versions of Java like Comparator or Runnable. Those existing interfaces are extended to enable Lambda support via the @FunctionalInterface annotation."

Just checking, Callable is a @FunctionalInterface:

http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Callable.html

So it seems to be a decent one to go with.  What do you think?

kcooney:
If we add an API for this, let's get it into  JUnit 4.13  and 5.0 and not have JUnit require JDK 1.8

@NarendraPathai I would prefer not having a complex DSL for assertions on exceptions as a part of JUnit. It is too hard to get it right. I would prefer we keep the API simple (David's API is borderline IMHO, but okay). Let's leave it to Hamcrest/Fest/Truth to create nice, competing assertion APIs.

marcphilipp:
@kcooney We could, technically, get this API to only require JDK 6. But I don't see any benefit in it. IMHO you have to use Java 8 in order to use it without hurting your eyes: :smirk:

stefanbirkner:
@dsaff Callable doesn't allow you to throw Throwable.

Tibor17:
:) Can you guys talk aside and make consistent decision avout release 4.13/5.0. This looks like you are not synchronized. Preferably create a branch with jdk 8 prototype for these purposes.

marcphilipp:
@kcooney My suggestion was to get this (at some point) into [junit-lambda](https://github.com/junit-team/junit-lambda), not "junit-core". Thus, JUnit wouldn't require JDK 8 but junit-lambda would as we've been discussing via email.

kcooney:
@marcphilipp I agree this wouldn't be a useful API to users that don't have Java 8, but that doesn't mean that the API itself needs to depend on Java 8 APIs. And I could see people finding this API useful even if they wouldn't use any features of a JUnit Lambda library. It's certainly harder to use incorrectly than ExpectedException

Let's initially include it in "JUnit lambda", and as the shape of "JUnit lambda" gets clearer, we can discuss whether to release this there or as part of JUnit 5.

kcooney:
@Tibor17 the maintainers are completely on the same page about 4.13 and 5.0. 5.0 will have breaking API changes, while 4.13, if it is released, would not.

src/main/java/org/junit/ExceptionSupplier.java
seanjreilly: It has to throw `Exception` just in case it calls a method that declares a checked exception but isn't expected to throw it. Strictly speaking, the `T` isn't necessary, but having it declares the semantic expectations much better. If the method didn't throw `T`, then the interface would have a type parameter that it doesn't use — and the type parameter really helps explain what's going to happen when calling the expect method.

seanjreilly:
@kcooney — your suggestions make perfect sense. I've omitted `@FunctionalInterface` only so that this code will compile with Java 5+ (although as you say it's not really useful unless you have Java 8). If you're putting into a Junit Lambda library that requires Java 8 then I can definitely add `@FunctionalInterface`.

Per your older comments, the one difference from your preference is that if an unexpected exception type is encountered, that exception is thrown itself, instead of an AssertionError. I felt that it gives a clearer indication of what's happening in an IDE (in IntelliJ you get a different icon for the failure, for example). If you feel strongly about that, I'm happy to change this (the unexpected exception would be a cause of the AssertionError anyway, so it's not too different).

I've added an improvements to the method that allow it to expect any kind of Throwable (Exception or Error subclass) — your suggestion of having an unexpectedException wrapped in an AssertionError made that much more elegant, so I've incorporated that change as well.

With respect to junit lambda, are you happy to merge this pull request there yourself, or is there a separate project that you'd like me to make a pull request to?

kcooney:
@seanjreilly having unexpected exception simply being rethrown sounds reasonable.

We aren't quite ready to accept pull requests for JUnit Lambda (which is why I was originally suggesting targeting JUnit 5)

marcphilipp:
Proposal:
- We open an issue at junit-lambda and reference this pull request.
- We close this pull request.
- When we are ready to accept pull request for JUnit Lambda we'll either merge it in ourselves or ask @seanjreilly to submit another pull request.

What do you think?

kcooney:
I agree with Marc's proposal

jlink:
Since we have no idea (yet) of how the internals of JUnit lamda will work,
I'd go with Marc's proposal and delay accepting pull requests for lamda
until we are ready.

2015-01-15 2:44 GMT+01:00 Kevin Cooney <notifications@github.com>:

marcphilipp:
@seanjreilly Thanks again for your pull request bwe will reconsider it when we're ready.

PR #1057, opened 2014-12-29
Parameterized reuses testclass

### LINKED ISSUE #1046: Possible to reuse TestClass instances in parameterized tests?
`TestClass` instances aren't reused for parameterized tests, which leads to repeated class scanning and noticeable memory allocations (if you happen to run a couple of thousand parameterized test cases).

For example in my case I've got about 12.000 parameter entries and the test class itself has five methods marked with annotations and seven fields marked with annotations. Every `TestClass` instance uses two `LinkedHashMap` instances, and every `Field` and `Method` instance uses a single `LinkedHashMap` instance (cf. `java.lang.reflect.Field#declaredAnnotations` resp. `java.lang.reflect.Method#declaredAnnotations`). That means each `TestClass` instance references about 14 `LinkedHashMap` instances. With 12.000 parameterized tests, I've ended up with about 168.000 `LinkedHashMap` instances (*). :cold_sweat:

That makes me wonder why `BlockJUnit4ClassRunnerWithParameters` doesn't reuse the existing `TestClass` instance [here](https://github.com/junit-team/junit/blob/master/src/main/java/org/junit/runners/parameterized/BlockJUnit4ClassRunnerWithParameters.java#L27). Is there any technical reason which prevents reusing the instance?

As a workaround I'm currently sub-classing `BlockJUnit4ClassRunnerWithParameters` in order to override the `createTestClass` method.

(*) Actually it was over 300.000 instances, so I'm still missing a factor of two somewhere.

stefanbirkner:
The `BlockJUnit4ClassRunnerWithParameters` created a new instance of `TestClass` for each parameter set. This lead to repeated class scanning and noticeable memory allocation. Reusing the `TestClass` avoids theses side effects. `TestClass` is reused by the Theories runner, too, because it is now possible.

Fixes #1046.

### DIFF
--- a/src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
+++ b/src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
@@ -29,6 +29,7 @@ import org.junit.runners.model.FrameworkMethod;
 import org.junit.runners.model.InitializationError;
 import org.junit.runners.model.MultipleFailureException;
 import org.junit.runners.model.Statement;
+import org.junit.runners.model.TestClass;
 
 /**
  * Implements the JUnit 4 standard test case class model, as defined by the
@@ -63,6 +64,16 @@ public class BlockJUnit4ClassRunner extends ParentRunner<FrameworkMethod> {
      */
     public BlockJUnit4ClassRunner(Class<?> klass) throws InitializationError {
         super(klass);
+    }
+
+    /**
+     * Creates a BlockJUnit4ClassRunner to run {@code testClass}.
+     *
+     * @throws InitializationError if the test class is malformed.
+     * @since 4.13
+     */
+    public BlockJUnit4ClassRunner(TestClass testClass) throws InitializationError {
+        super(testClass);
     }
 
     //
--- a/src/main/java/org/junit/runners/ParentRunner.java
+++ b/src/main/java/org/junit/runners/ParentRunner.java
@@ -84,6 +84,20 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
         validate();
     }
 
+    /**
+     * Constructs a new {@code ParentRunner} that will run the {@code TestClass}.
+     * @since 4.13
+     */
+    protected ParentRunner(TestClass testClass) throws InitializationError {
+        this.testClass = testClass;
+        validate();
+    }
+
+    /**
+     * @deprecated Please use {@link #ParentRunner(org.junit.runners.model.TestClass)}.
+     * @since 4.12
+     */
+    @Deprecated
     protected TestClass createTestClass(Class<?> testClass) {
         return new TestClass(testClass);
     }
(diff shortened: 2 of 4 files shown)


kcooney:

This sounds like it fixes a real performance issue. A better fix seems to be adding a static factory method to `TestClass` that does memoization. Anyone want to take this on?

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -84,6 +84,20 @@ protected ParentRunner(Class<?> testClass) throws InitializationError {
...
+    protected ParentRunner(TestClass testClass) throws InitializationError {
+        this.testClass = testClass;
+        validate();
+    }
+
+    /**
+     * @deprecated Please use {@link #ParentRunner(org.junit.runners.model.TestClass)}.
+     * @since 4.12
+     */
+    @Deprecated
```
kcooney: I was going through open issues seeing what can/should be done in preparation of 4.13, and I noticed that this change wouldn't break people that extend `Theories`. It could break people that extend `BlockJUnit4ClassRunnerWithParameters and override `createTestClass()`.

I did some research. The `createTestClass()` method was so a custom runner could return a sub-class of `TestClass`  (see https://github.com/junit-team/junit4/pull/760#issuecomment-27836350 for the context).
(As an aside, calling non-static methods from the constructor should be avod

So for this change to break someone, all of the following would have to be true:

1. You would need to use `@ParametersRunnerFactory`
2. Your `ParametersRunnerFactory` implementation would have to create a subclass of `BlockJUnit4ClassRunnerWithParameters`
3. Your subclass of `BlockJUnit4ClassRunnerWithParameters` would have to override `createTestClass()` and return a custom subclass

Note that this could only happen in projects that use JUnit 4.12, as these APIs didn't exist before that.

It's unfortunate that we decided to create `createTestClass()`. The changes proposed in *this* pull would also allow a custom runner to use a custom sub-class of `TestClass`. Also, `createTestClass()` uses the anti-pattern of having a class override a method in the base class that is called in the parent constructor.

In any case, upon further reflection I'm leaning toward merging this pull, with one suggested change.

src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
```diff
@@ -65,6 +66,16 @@ public BlockJUnit4ClassRunner(Class<?> klass) throws InitializationError {
         super(klass);
     }
 
+    /**
+     * Creates a BlockJUnit4ClassRunner to run {@code testClass}.
+     *
+     * @throws InitializationError if the test class is malformed.
+     * @since 4.13
+     */
+    public BlockJUnit4ClassRunner(TestClass testClass) throws InitializationError {
```
kcooney: Could we please make this protected?

kcooney:
Closing in favor of #1449

PR #1060, opened 2015-01-04
Change method getFilteredChildren's visibility to protected

sduxzz:
Change method getFilteredChildren's visibility to protected, so this method can be called by subclasses

### DIFF
--- a/src/main/java/org/junit/runners/ParentRunner.java
+++ b/src/main/java/org/junit/runners/ParentRunner.java
@@ -407,6 +407,18 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
         }
     }
 
+    protected Collection<T> getFilteredChildren() {
+        if (filteredChildren == null) {
+            synchronized (childrenLock) {
+                if (filteredChildren == null) {
+                    filteredChildren = Collections
+                            .unmodifiableCollection(getChildren());
+                }
+            }
+        }
+        return filteredChildren;
+    }
+
     //
     // Private implementation
     //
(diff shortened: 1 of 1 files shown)


kcooney:
Can you give a use case for why subclasses would need to call this?

sduxzz:
hi, I would like to override childrenInvoker method, so I can redefine the process how children will be run. But, runChildren and getFilteredChildren method are all private, so I can't do that. Example code:
```java
public class SomeRunner<T> extends ParentRunner<T> {
      @Override
    protected Statement childrenInvoker(final RunNotifier notifier) {
        List<T> filteredChildren = new ArrayList<T>(getFilteredChildren()); // Cant't access getFilteredChildren()
        return new ChristStatement<T>(getTestClass(), filteredChildren, notifier, this, executor());
    }
}
 ```

kcooney:
@sduxzz could you solve your problem by overriding `runChild()`? Can you give more details about what you are trying to do? Making private methods protected makes the class harder to maintain, so it's good to know what problem(s) you are trying to solve first.

sduxzz:
@kcooney Thank you for your reply. I will try to use your advice to solve my problems.  : ) 
By the way, can junit support running test cases concurrently in future? In other words, can users use something to tell junit framework run these test cases sequently and  runother test cases concurrently  in future?

sduxzz:
@kcooney Thank you for your reply. I will try to use your advice to solve my problems. : )

kcooney:
@sduxzz did the advice help, or do we still need this pull request?

sduxzz:
@kcooney  yes,  thank you for your advice,  and this pull request can be closed.

PR #1068, opened 2015-01-14
I added a constructor that takes in both a message and a cause. 

calen-legaspi:
It's just a call to this(new Exception(message, cause)).

### DIFF
--- a/src/main/java/org/junit/runners/model/InitializationError.java
+++ b/src/main/java/org/junit/runners/model/InitializationError.java
@@ -44,4 +44,8 @@ public class InitializationError extends Exception {
     public List<Throwable> getCauses() {
         return fErrors;
     }
+    
+    public InitializationError(String message, Throwable cause) {
+        this(new Exception(message, cause));        
+    }
 }


kcooney:
I don't like the idea of yet another constructor that creates a generic exception.

Instead, would it make sense for your code to either pass the cause to the constructor that takes in an exception, or  create a specific exception with a message and cause and pass that to the constructor that takes an exception?

kcooney:
I guess my real issue is that if I used this constuctor, I would expect it to set the message and cause of the exception to the passed-in parameters, but it sets neither.

marcphilipp:
@calen-legaspi What's your use case for this constructor?

marcphilipp:
I'm against this change. I don't see a reason for the extra constructor. If you need to construct such an `InitializationError` you can call `new InitializationError(new Exception(message, cause))`.

Stephan202:
I'm with @marcphilipp; the additional API surface doesn't seem worth it.

calen-legaspi:
Apologies for the late response, there was a long holiday here.

I was motivated to create this constructor because I wanted to patch Concordion. I encountered a [Concordion issue] (https://github.com/concordion/concordion/issues/100) where the stack trace just returned: "Failed to initialize ConcordionRunner" without returning the cause. Concordion just did a "throw new InitializationError("Failed to initialize ConcordionRunner");" without storing the cause. I wanted to store the cause so that it would be easier to figure out future issues.

kcooney:
I am actually not against adding the constructor (a constructor taking a Throwable and a string is quite common), but the behavior of the current implementation is unexpected.

What would happen if Concordian threw a RuntimeException with a message and cause?

PR #1069, opened 2015-01-14
Upgrade maven-surefire-plugin to latest version

hansjoachim:
Saw that maven-surefire-plugin could be upgraded to a newer version. Verified `mvn clean install` still runs successfully.


marcphilipp:
Thanks!

PR #1071, opened 2015-01-18
klass to testClass

jorgeejgonzalez:
Changed the name of the constructor variable klass to reveal its true purpose as it is defined in ParentRunner


kcooney:
Merged. Thanks!

PR #1072, opened 2015-01-19
Patch on top of pull 1065

kcooney:
Here is my proposed fix for the regression that I noticed in pull #1065

Do not merge this pull request; we'll continue to track it in 1065. I'm only sending this here because 1) it's easier than discussing the code changes abstractly, and 2) to ensure that the code builds with Travis

### DIFF
--- a/src/main/java/org/junit/experimental/max/MaxCore.java
+++ b/src/main/java/org/junit/experimental/max/MaxCore.java
@@ -114,7 +114,7 @@ public class MaxCore {
                     return new Suite((Class<?>) null, runners) {
                     };
                 } catch (InitializationError e) {
-                    return new ErrorReportingRunner(null, e);
+                    return new ErrorReportingRunner((Class<?>) null, e);
                 }
             }
         };
--- a/src/main/java/org/junit/internal/runners/ErrorReportingRunner.java
+++ b/src/main/java/org/junit/internal/runners/ErrorReportingRunner.java
@@ -13,28 +13,39 @@ import org.junit.runners.model.InitializationError;
 public class ErrorReportingRunner extends Runner {
     private final List<Throwable> causes;
 
-    private final String classNames;
+    // Only one of the following two fields will be non-null.
+    private final Class<?> testClass;
+    private final String displayName;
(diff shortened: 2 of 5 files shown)


stefanbirkner:
LGTM

src/main/java/org/junit/experimental/max/MaxCore.java
```diff
@@ -114,7 +114,7 @@ public Runner getRunner() {
                     return new Suite((Class<?>) null, runners) {
                     };
                 } catch (InitializationError e) {
-                    return new ErrorReportingRunner(null, e);
+                    return new ErrorReportingRunner((Class<?>) null, e);
```
marcphilipp: This is a possible regression for external uses of `ErrorReportingRunner`, too? Better to use a "named constructor", a.k.a. factory method?

src/main/java/org/junit/internal/runners/ErrorReportingRunner.java
```diff
@@ -13,19 +13,39 @@
...
+    /**
+     * @since 4.13
+     */
+    public ErrorReportingRunner(String displayName, Throwable cause) {
+        if (displayName == null) {
+            throw new NullPointerException("Display name cannot be null");
+        }
+        testClass = null;
+        this.displayName = displayName;
         causes = getCauses(cause);
```
marcphilipp: The three lines above and the last three lines of the first constructor look very similar and could be extracted to a third, private constructor that takes three arguments.

src/main/java/org/junit/experimental/max/MaxCore.java
kcooney: Does it matter that `ErrorReportingRunner` is in an internal package?

src/main/java/org/junit/internal/runners/ErrorReportingRunner.java
kcooney: While that would lead to a (slight) decrease in code duplication, I think it would sacrifice clarity. It is extremely clear what the constructors do

Issue #1066, opened 2015-01-10
Exceptions thrown from methodBlock() result in the failing method being unrooted in test reports


sbrannen:
The introduction of the `runLeaf()` method in `BlockJUnit4ClassRunner` in JUnit 4.9 introduces a regression with regard to exception handling. 

Consult the following versions to see the difference:

1. [BlockJUnit4ClassRunner (4.8)](https://github.com/junit-team/junit/blob/r4.8/src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java)
1. [BlockJUnit4ClassRunner (4.9](https://github.com/junit-team/junit/blob/r4.9/src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java)

As stated in the Javadoc for `methodBlock()`, `methodBlock()` _can be overridden in subclasses, either by overriding this method, or the implementations creating each sub-statement._

Thus, although other changes to the JUnit codebase itself in the JUnit 4.9 release do not result in an exception being thrown from `methodBlock()`, custom modifications to `methodBlock()` or the methods it invokes may in fact throw exceptions. In such cases, exceptions thrown from methodBlock() cause the current test execution to abort immediately. As a result, the failing test method is **_unrooted_** in test reports, and subsequent test methods are **never invoked**. Furthermore, RunListeners registered with JUnit are **not notified**.

The root cause is that the invocation of `methodBlock()` is no longer executed within a try-catch block as was the case in previous versions of JUnit.

A fix for this regression can be seen in recent changes to the [SpringJUnit4ClassRunner](https://github.com/spring-projects/spring-framework/commit/b81c522ee1624f3d5f170fc3468f4600956950bd) from the Spring Framework.

See also: [SPR-12613](https://jira.spring.io/browse/SPR-12613) in Spring's JIRA issue tracker.

kcooney:
Thanks for the report. Since this bug was introduced so long ago, I'm guessing we won't do a bug-fix release for 4.12, and instead try to get a fix into 4.13

As a work-around, have your implementation of `methodBlock()` do the` try` ... `catch`, and if an exception is thrown, return a Statement that rethrows the exception.

sbrannen:
@kcooney, did you look at the ` SpringJUnit4ClassRunner` link above?

It already does exactly what you suggested (except in the `runChild()` method in order to maintain the previous behavior). ;)

kcooney:
@sbrannen I was just leaving the comment about the work-around in case someone finds this thread.

[...]

PR #1082, opened 2015-02-15
Ensure exceptions from methodBlock() don't result in unrooted tests

sbrannen:
The introduction of the runLeaf() method in BlockJUnit4ClassRunner in
JUnit 4.9 introduced a regression with regard to exception handling.

Specifically, the invocation of methodBlock() is no longer executed
within a try-catch block as was the case in previous versions of JUnit.

Custom modifications to methodBlock() or the methods it invokes may in
fact throw exceptions. In such cases, exceptions thrown from
methodBlock() cause the current test execution to abort immediately. As
a result, the failing test method is unrooted in test reports, and
subsequent test methods are never invoked. Furthermore, RunListeners
registered with JUnit are not notified.

This commit addresses this issue by wrapping the invocation of
methodBlock() within a try-catch block. If an exception is not thrown,
the resulting Statement is passed to runLeaf(). If an exception is
thrown, it is wrapped in a Fail statement which is passed to runLeaf().

Issue: #1066

### DIFF
--- a/src/test/java/org/junit/runners/CustomBlockJUnit4ClassRunnerTest.java
+++ b/src/test/java/org/junit/runners/CustomBlockJUnit4ClassRunnerTest.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.runners;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.junit.Ignore;
+import org.junit.Test;
+import org.junit.runner.Description;
+import org.junit.runner.notification.Failure;
+import org.junit.runner.notification.RunListener;
+import org.junit.runner.notification.RunNotifier;
+import org.junit.runners.model.FrameworkMethod;
+import org.junit.runners.model.InitializationError;
+import org.junit.runners.model.Statement;
+
+/**
+ * Tests that verify proper behavior for custom runners that extend
+ * {@link BlockJUnit4ClassRunner}.
+ *
+ * @author Sam Brannen
+ * @since 4.13
+ */
+public class CustomBlockJUnit4ClassRunnerTest {
+
+	@Test
+	public void exceptionsFromMethodBlockMustNotResultInUnrootedTests() throws Exception {
+		TrackingRunListener listener = new TrackingRunListener();
+		RunNotifier notifier = new RunNotifier();
+		notifier.addListener(listener);
+
+		new CustomBlockJUnit4ClassRunner(CustomBlockJUnit4ClassRunnerTestCase.class).run(notifier);
+		assertEquals("tests started.", 2, listener.testStartedCount.get());
+		assertEquals("tests failed.", 1, listener.testFailureCount.get());
+		assertEquals("tests finished.", 2, listener.testFinishedCount.get());
+	}
+
+
+	@Ignore("This test case is run manually by the enclosing test class")
+	public static class CustomBlockJUnit4ClassRunnerTestCase {
+		@Test public void shouldPass() { /* no-op */ }
+		@Test public void throwException() { /* no-op */ }
+	}
+
+	/**
+	 * Custom extension of {@link BlockJUnit4ClassRunner} that always throws
+	 * an exception from the {@code methodBlock()} if a test method is named
+	 * exactly {@code "throwException"}.
+	 */
+	private static class CustomBlockJUnit4ClassRunner extends BlockJUnit4ClassRunner {
+
+		CustomBlockJUnit4ClassRunner(Class<?> testClass) throws InitializationError {
+			super(testClass);
+		}
+
+		@Override
+		protected Statement methodBlock(FrameworkMethod method) {
+			if ("throwException".equals(method.getName())) {
+				throw new RuntimeException("throwException() test method invoked");
+			}
+			return super.methodBlock(method);
+		}
+	}
+
+	/**
+	 * Simple {@link RunListener} that tracks the number of times that
+	 * certain callbacks are invoked.
+	 */
+	private static class TrackingRunListener extends RunListener {
+
+		final AtomicInteger testStartedCount = new AtomicInteger();
+		final AtomicInteger testFailureCount = new AtomicInteger();
+		final AtomicInteger testFinishedCount = new AtomicInteger();
+
+
+		@Override
+		public void testStarted(Description description) throws Exception {
+			testStartedCount.incrementAndGet();
+		}
+
+		@Override
+		public void testFailure(Failure failure) throws Exception {
+			testFailureCount.incrementAndGet();
+		}
+
+		@Override
+		public void testFinished(Description description) throws Exception {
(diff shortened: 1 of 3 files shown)


kcooney:
Thanks? Could you add a test?

sbrannen:
Sure. I just introduced `CustomBlockJUnit4ClassRunnerTest` and squashed the commits.

src/test/java/org/junit/runners/CustomBlockJUnit4ClassRunnerTest.java
```diff
@@ -0,0 +1,148 @@
...
+			}
+			return super.withBefores(method, target, statement);
+		}
+	}
+
+	/**
+	 * Simple {@link RunListener} which tracks how many times certain JUnit callback
+	 * methods were called: only intended for the integration test suite.
+	 * <p>This code has been copied directly from the Spring Framework.
+	 * @author Sam Brannen
```
marcphilipp: Are you the author of this class in the Spring Framework? Otherwise, this might be a bit fishy license-wise, right?

sbrannen: I am in fact the author of the code in question, but as far as I understand that is irrelevant since the ASL license header is included in the file along with the `@author` tags.

Now, I am _not_ a lawyer, but I don't think there is any issue here since:

1. The original ASL 2.0 license header is present in the file I submitted.
2. The code in question is not distributed per se with JUnit JARs; it resides in the test suite.

kcooney:
I think you manually need to add the test to the test suite in order for maven to run the test

sbrannen:
OK. I added `CustomBlockJUnit4ClassRunnerTest` to the `AllTests` suite.

src/test/java/org/junit/runners/CustomBlockJUnit4ClassRunnerTest.java
```diff
@@ -0,0 +1,148 @@
+/*
+ * Copyright 2002-2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
kcooney: JUnit isn't licensed under the Apache License. Please remove this entire comment block

src/test/java/org/junit/runners/CustomBlockJUnit4ClassRunnerTest.java
```diff
@@ -0,0 +1,148 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.junit.runners;
+
+import static org.junit.Assert.*;
```
kcooney: We don't use wildcard imports in JUnit

src/test/java/org/junit/runners/CustomBlockJUnit4ClassRunnerTest.java
```diff
@@ -0,0 +1,148 @@
...
+		new CustomBlockJUnit4ClassRunner(testClass).run(notifier);
+		assertEquals("Number of tests started for [" + testClass + "].", 2, listener.getTestStartedCount());
+		assertEquals("Number of tests ignored for [" + testClass + "].", 0, listener.getTestIgnoredCount());
+		assertEquals("Number of assumption failures for [" + testClass + "].", 0, listener.getTestAssumptionFailureCount());
+		assertEquals("Number of test failures for [" + testClass + "].", 1, listener.getTestFailureCount());
+		assertEquals("Number of tests finished for [" + testClass + "].", 2, listener.getTestFinishedCount());
+	}
+
+
+	@Ignore("This test case is run manually by the enclosing test class")
```
kcooney: I don't think you need the `@Ignore`

src/test/java/org/junit/runners/CustomBlockJUnit4ClassRunnerTest.java
kcooney: But JUnit isn't released under ASL. Could we use a simpler runner where each of the methods appends a hard-coded string to an `ArrayList`, then you can do assertions on the list.

src/test/java/org/junit/runners/CustomBlockJUnit4ClassRunnerTest.java
sbrannen: This has been replaced with a license header compliant with the Eclipse Foundation.

src/test/java/org/junit/runners/CustomBlockJUnit4ClassRunnerTest.java
sbrannen: Seriously?

Not even for `org.junit.Assert` _within_ the test suite?

Unique, but okay.

Replaced with single static import.

src/test/java/org/junit/runners/CustomBlockJUnit4ClassRunnerTest.java
sbrannen: Define "need". ;)

When executing tests via `org.junit.tests.AllTests`, you are correct: the `@Ignore` is unnecessary.

However, when executing tests within an IDE such as Eclipse, it is very useful to have such _internal_ test classes disabled by default. In that manner, there are no false negatives when executing all tests within a given package, class, etc. Otherwise, such tests that always fail (intentionally) will produce undesirable noise and distract developers from the task at hand.

src/test/java/org/junit/runners/CustomBlockJUnit4ClassRunnerTest.java
sbrannen: I have rewritten the `TrackingRunListener` from scratch. Any remaining similarities to code in Spring's testing suite are therefore coincidental due to the nature of the task at hand.

src/test/java/org/junit/runners/CustomBlockJUnit4ClassRunnerTest.java
kcooney: We don't have copyright and licenses in any of the other files that I'm aware of. I worry that someone in the future will see this, think it's an error, and change it to the EPL license.

There's very little code you are copying from the Spring code. Let's just rework it so we don't need to add license info

src/test/java/org/junit/runners/CustomBlockJUnit4ClassRunnerTest.java
kcooney: When I run tests in Eclipse, it doesn't run nested classes. We use nested classes in many of our tests, and I haven't seen us use `@Ignore` (and, in fact, some of the tests depend on there not being `@Ignore` present)

kcooney:
Thanks for the changes. I think the remaining (quite minor) issues can be resolved when we merge this. I'll try to find time Tuesday night to do that.

marcphilipp:
Which Tuesday? ;-)

kcooney:
Merged. Thanks! Sorry for the delay; both home and work life have been busy for me

sbrannen:
Great! Thanks for merging it.

kcooney:
@sbrannen do you mind updating the JUnit 4.13 release notes to mention this pull?

sbrannen:
Mentioned this pull request in the release notes [here](https://github.com/junit-team/junit/wiki/4.13-Release-Notes#pull-request-1082-ensure-exceptions-from-blockjunit4classrunnermethodblock-dont-result-in-unrooted-tests).

PR #1086, opened 2015-02-28
Changed line endings to LF

hansjoachim:
SonarQube had problems scanning this file because it believed it got
multiple blame annotations for the same line.
This seemed to be the only file with CR line endings, so I changed it
to LF which is consistent with the files in the same directory.


hansjoachim:
GitHub displays the diff in a quite odd way, so whoever reviewing this might want to look at `git log -p` instead...

kcooney:
Thanks!

PR #1087, opened 2015-03-01
Use Timeout.millis(long) in javadoc for Timeout class

mustafau:
Javadoc for Timeout class should use Timeout.millis(long) instead of deprecated Timeout(int) constructor


marcphilipp:
Thanks!

Issue #1088, opened 2015-03-04
Add naive Fibonacci implementation to the example in Parameterized's javadoc.


j--baker:
Make it possible to copy, paste, and run the code.
Make it 100% clear what is happening by being concrete.

kcooney:
@j--baker That would make the Javadoc quite long without adding much benefit. What if we added a link to http://en.wikipedia.org/wiki/Fibonacci_number instead?

PR #1089, opened 2015-03-04
Add naive addition implementation to the example in Parameterized's Javadoc.

j--baker:
Just a quick update to make example more cut & paste friendly.

### DIFF
--- a/src/main/java/org/junit/runners/Parameterized.java
+++ b/src/main/java/org/junit/runners/Parameterized.java
@@ -27,6 +27,14 @@ import org.junit.runners.parameterized.TestWithParameters;
  * <pre>
  * &#064;RunWith(Parameterized.class)
  * public class FibonacciTest {
+ *     private static class Fibonacci {
+ *         public static final int compute(int input) {
+ *             if (input < 0) { throw new IllegalArgumentException(); }
+ *             if (input >= 2) { return compute(input - 2) + compute(input - 1); }
+ *             return input;
+ *         }
+ *     }
+ *
  *     &#064;Parameters(name= &quot;{index}: fib[{0}]={1}&quot;)
  *     public static Iterable&lt;Object[]&gt; data() {
  *         return Arrays.asList(new Object[][] { { 0, 0 }, { 1, 1 }, { 2, 1 },
--- a/src/main/java/org/junit/runners/Parameterized.java
+++ b/src/main/java/org/junit/runners/Parameterized.java
@@ -25,33 +25,45 @@ import org.junit.runners.parameterized.TestWithParameters;
  * <p>
  * For example, to test a Fibonacci function, write:
  * <pre>
+ * public class IntAdder {
+ *     public static int add(int addend1, int addend2) {
+ *         return addend1 + addend2;
+ *     }
+ * }
+ *
+ * &#064;RunWith(Parameterized.class)
+ * public class IntAdditionTest {
+ *     &#064;Parameters()
+ *     public static Iterable&lt;Object[]&gt; data() {
+ *         return Arrays.asList(new Object[][] {
+ *             { Integer.MIN_VALUE, Integer.MIN_VALUE, 0 }, { Integer.MIN_VALUE, Integer.MIN_VALUE + 1, 1 },
+ *             { 0, 0, 0 }, { 1, 1, 2 }, { 2, 1, 3 }, { 3, 2, 5 },
+ *             { 4, 3, 7 }, { 5, 5, 10 }, { 6, 8, 14 },
+ *             { Integer.MAX_VALUE, Integer.MIN_VALUE, -1 }, { Integer.MAX_VALUE, Integer.MIN_VALUE + 1, 0 },
+ *             { Integer.MAX_VALUE, 1, Integer.MIN_VALUE }, { Integer.MAX_VALUE, 2, Integer.MIN_VALUE + 1 },
+ *             { Integer.MAX_VALUE, Integer.MAX_VALUE - 1, -3 }, { Integer.MAX_VALUE, Integer.MAX_VALUE, -2 } });
+ *     }
+ * 
+ *     private final int addend1;
+ *     private final int addend2;
+ *     private final int sum;
+ * 
+ *     public IntAdditionTest(final int addend1, final int addend2, final int sum) {
+ *         this.addend1 = addend1;
+ *         this.addend2 = addend2;
+ *         this.sum = sum;
+ *     }
+ * 
+ *     &#064;Test
+ *     public void test() {
+ *         assertEquals(sum, IntAdder.add(addend1, addend2));
+ *     }
+ * }
  * public class Fibonacci {
  *     public static int compute(int input) {
  *         if (input &lt; 0) { throw new IllegalArgumentException(); }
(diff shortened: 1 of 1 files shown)


kcooney:
I think having the code under test be a nested class is actually more confusing than having no example.

As I said in the bug, can you just add a link to either Wikipedia or a page on the JUnit wiki?

j--baker:
How about just showing the Fibonacci class as a non-nested class?

Having a simple test case that shows all the moving pieces and works (after adding imports) after copying and pasting is my concern.  I'm aiming for a "batteries included" example.   :battery: :)

kcooney:
@j--baker we can wait to see what the other @junit-team/junit-committers think, but I would prefer not to add extra space in the Javadoc to provide a Fibonacci implementation, but I'm fine with a link.

stefanbirkner:
How about replacing the Fibonacci example with a simple example that just calculates sums?

marcphilipp:
+1

j--baker:
No new posts in a day, so I went with @stefanbirkner 's suggestion (:heavy_plus_sign: addition test).  Thanks!

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -23,34 +23,55 @@
  * When running a parameterized test class, instances are created for the
  * cross-product of the test methods and the test data elements.
  * <p>
- * For example, to test a Fibonacci function, write:
+ * For example, this class has a method that adds two integers and returns the sum.
```
kcooney: nit: end this sentence with a colon

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -23,34 +23,55 @@
...
  * <pre>
+ * &#047;* This is the class being tested. *&#047;
```
kcooney: How about simply "Example class that adds two integers"? You already implied it was the code being tested above

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -23,34 +23,55 @@
...
+ *         return addend1 + addend2;
+ *     }
+ * }
+ * </pre>
+ * The following is a test for the above example class.
+ * For each element returned from the <code>&#064;Parameters</code> annotated method,
+ * an instance of the <code>Parameterized</code> test class is created and
+ * its <code>&#064;Test</code> annotated method (e.g. <code>test()</code>) is invoked.
+ * <pre>
+ * &#047;* This is is the unit test for {&#064;link IntAdder}. *&#047;
```
kcooney: How about "Tests for {@link IntAdder}."

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -23,34 +23,55 @@
...
  * &#064;RunWith(Parameterized.class)
- * public class FibonacciTest {
- *     &#064;Parameters(name= &quot;{index}: fib[{0}]={1}&quot;)
+ * public class IntAdderTest {
+ *     &#047;* Each of these is used to create a new {&#064;link IntAdderTest} instance. *&#047;
+ *     &#064;Parameters(name= &quot;{index}: {0} + {1} = {2}&quot;)
```
kcooney: style-nit: our new style is to put spaces before and after the `=` (see https://google-styleguide.googlecode.com/svn/trunk/javaguide.html)

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -23,34 +23,55 @@
...
  *     public static Iterable&lt;Object[]&gt; data() {
- *         return Arrays.asList(new Object[][] { { 0, 0 }, { 1, 1 }, { 2, 1 },
- *                 { 3, 2 }, { 4, 3 }, { 5, 5 }, { 6, 8 } });
+ *         return Arrays.asList(new Object[][] {
+ *             { Integer.MIN_VALUE, Integer.MIN_VALUE, 0 }, { Integer.MIN_VALUE, Integer.MIN_VALUE + 1, 1 },
```
kcooney: I would suggest you don't need so many data points for the reader to get the idea. I think we should remove the examples that cause it to overflow, because it might confuse users that don't know how addition with overflow works.

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -23,34 +23,55 @@
...
- *
- *     public FibonacciTest(int input, int expected) {
- *         fInput= input;
- *         fExpected= expected;
+ * 
+ *     private final int addend1;
+ *     private final int addend2;
+ *     private final int sum;
+ * 
+ *     public IntAdderTest(final int addend1, final int addend2, final int sum) {
```
kcooney: nit: remove the use of "final" in the parameters; it just adds noise

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -67,45 +88,58 @@
  * <dd>...</dd>
  * </dl>
  * <p>
- * In the example given above, the <code>Parameterized</code> runner creates
- * names like <code>[1: fib(3)=2]</code>. If you don't use the name parameter,
- * then the current parameter index is used as name.
+ * In the example given above, the <code>Parameterized</code> annotation
```
kcooney: I personally find the previous wording of this section easier to read and I think it provides enough information for people to follow.

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -117,7 +151,7 @@
  * <pre>
  * &#064;Parameters
  * public static Iterable&lt;? extends Object&gt; data() {
- * 	return Arrays.asList(&quot;first test&quot;, &quot;second test&quot;);
+ * 	   return Arrays.asList(&quot;first test&quot;, &quot;second test&quot;);
```
kcooney: please leave the formatting of the code the way it was

kcooney:
Thanks for the improvements!

FYI: don't worry about squashing commits at this time; the github review tool is actually harder to use once you squash commits. We can squash them on our side when we commit.

j--baker:
@kcooney - Thanks for the quick and thorough review (and your patience)!
Please review the latest updates.
(Also, thanks for mentioning commit squashing.)

src/main/java/org/junit/runners/Parameterized.java
kcooney: "is is". Also (and I know I'm nit picking) but JavaDoc generally doesn't start with "This". I suggest simply "Tests for {&#064;link IntAdder}." or "Unit tests for {&#064;link IntAdder}."

marcphilipp:
Linking to #1088.

marcphilipp:
LGTM!

@junit-team/junit-committers Can you please take a look?

stefanbirkner:
LGTM, but I think that we should avoid an additional class for the example. Just using something that is already there like the + operator should be enough. It could look like https://github.com/junit-team/junit/commit/8f3a45c231a43b0b30adc4a315256a5f7e2e63af How do you think about that solution? Should I create a pull request?

marcphilipp:
+1 for @stefanbirkner's version

stefanbirkner:
Created pull request #1114.

marcphilipp:
Closed in favor of #1114.

@j--baker Thanks for submitting this pull request nevertheless!

PR #1091, opened 2015-03-06
keep @Ignore status independent from quantity and exception in @BeforeClass

OpenHelios:
* Separate between fireIgnoreChild() and runChild() in ParentRunner. The
method fireIgnoreChild() is only called for test methods annotated with
@Ignore and the method runChild() is only called for test methods, which
are not annotated with @Ignore.
* Add test to verify that the @Ignore status do not change, even if the
@BeforeClass method throws an exception.

### DIFF
--- a/src/main/java/org/junit/internal/runners/statements/RunBefores.java
+++ b/src/main/java/org/junit/internal/runners/statements/RunBefores.java
@@ -6,13 +6,24 @@ import org.junit.runners.model.FrameworkMethod;
 import org.junit.runners.model.Statement;
 
 public class RunBefores extends Statement {
+
+    private final Statement previous;
+
     private final Statement next;
 
     private final Object target;
 
     private final List<FrameworkMethod> befores;
 
-    public RunBefores(Statement next, List<FrameworkMethod> befores, Object target) {
+    /**
+     * @param previous The previous statement, or {@code null}.
+     * @param next The next statement after before methods.
+     * @param befores The list of before methods.
+     * @param target
+     */
+    public RunBefores(final Statement previous, final Statement next,
+            List<FrameworkMethod> befores, Object target) {
+        this.previous = previous;
         this.next = next;
         this.befores = befores;
         this.target = target;
@@ -20,6 +31,9 @@ public class RunBefores extends Statement {
 
     @Override
     public void evaluate() throws Throwable {
+        if (null != previous) {
+            previous.evaluate();
+        }
         for (FrameworkMethod before : befores) {
             before.invokeExplosively(target);
         }
--- a/src/main/java/org/junit/internal/runners/statements/RunBefores.java
+++ b/src/main/java/org/junit/internal/runners/statements/RunBefores.java
@@ -6,24 +6,13 @@ import org.junit.runners.model.FrameworkMethod;
 import org.junit.runners.model.Statement;
 
 public class RunBefores extends Statement {
-
-    private final Statement previous;
-
     private final Statement next;
 
     private final Object target;
 
     private final List<FrameworkMethod> befores;
 
-    /**
-     * @param previous The previous statement, or {@code null}.
-     * @param next The next statement after before methods.
-     * @param befores The list of before methods.
-     * @param target
-     */
-    public RunBefores(final Statement previous, final Statement next,
-            List<FrameworkMethod> befores, Object target) {
-        this.previous = previous;
+    public RunBefores(Statement next, List<FrameworkMethod> befores, Object target) {
         this.next = next;
         this.befores = befores;
         this.target = target;
@@ -31,9 +20,6 @@ public class RunBefores extends Statement {
 
     @Override
     public void evaluate() throws Throwable {
-        if (null != previous) {
-            previous.evaluate();
-        }
         for (FrameworkMethod before : befores) {
             before.invokeExplosively(target);
         }
--- a/src/main/java/org/junit/runners/ParentRunner.java
+++ b/src/main/java/org/junit/runners/ParentRunner.java
@@ -104,6 +104,13 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
     protected abstract Description describeChild(T child);
 
     /**
+     * Fires test ignore through {@code notifier} for child corresponding to
+     * {@code child}, which can be assumed to be an element of the list returned
+     * by {@link ParentRunner#getChildren()}.
+     */
+    protected abstract void fireIgnoreChild(T child, RunNotifier notifier);
+
+    /**
      * Runs the test corresponding to {@code child}, which can be assumed to be
      * an element of the list returned by {@link ParentRunner#getChildren()}.
      * Subclasses are responsible for making sure that relevant test events are
@@ -188,11 +195,13 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
      */
     protected Statement classBlock(final RunNotifier notifier) {
         Statement statement = childrenInvoker(notifier);
-        if (!areAllChildrenIgnored()) {
-            statement = withBeforeClasses(statement);
-            statement = withAfterClasses(statement);
-            statement = withClassRules(statement);
+        final Statement ignores = withIgnores(notifier);
+        if (areAllChildrenIgnored()) {
+            return ignores;
         }
+        statement = withBeforeClasses(ignores, statement);
+        statement = withAfterClasses(statement);
+        statement = withClassRules(statement);
         return statement;
     }
 
@@ -206,15 +215,33 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
     }
 
     /**
-     * Returns a {@link Statement}: run all non-overridden {@code @BeforeClass} methods on this class
+     * Returns a {@link Statement}: run all non-overridden {@code @Ignore} methods on this class
+     * and superclasses before executing {@code next}.
+     */
+    protected Statement withIgnores(final RunNotifier notifier) {
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                for (final T each : getFilteredChildren()) {
+                    if (isIgnored(each)) {
+                        ParentRunner.this.fireIgnoreChild(each, notifier);
+                    }
+                }
+            }
+        };
+    }
+
+    /**
+     * @param previous The previous statement, or {@code null}.
+     * @param next The next statement after methods annotated with {@link BeforeClass}.
+     * @return A {@link Statement}: run all non-overridden {@code @BeforeClass} methods on this class
      * and superclasses before executing {@code statement}; if any throws an
      * Exception, stop execution and pass the exception on.
      */
-    protected Statement withBeforeClasses(Statement statement) {
+    protected Statement withBeforeClasses(final Statement previous, final Statement next) {
         List<FrameworkMethod> befores = testClass
                 .getAnnotatedMethods(BeforeClass.class);
-        return befores.isEmpty() ? statement :
-                new RunBefores(statement, befores, null);
+        return new RunBefores(previous, next, befores, null);
     }
 
     /**
@@ -285,11 +312,13 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
         final RunnerScheduler currentScheduler = scheduler;
         try {
             for (final T each : getFilteredChildren()) {
-                currentScheduler.schedule(new Runnable() {
-                    public void run() {
-                        ParentRunner.this.runChild(each, notifier);
-                    }
-                });
+                if (!isIgnored(each)) {
+                    currentScheduler.schedule(new Runnable() {
+                        public void run() {
+                            ParentRunner.this.runChild(each, notifier);
+                        }
+                    });
+                }
             }
         } finally {
             currentScheduler.finished();
--- a/src/main/java/org/junit/runners/ParentRunner.java
+++ b/src/main/java/org/junit/runners/ParentRunner.java
@@ -64,9 +64,7 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
     private final TestClass testClass;
 
     // Guarded by childrenLock
-    private volatile Collection<T> filteredChildren;
-    private volatile Collection<T> ignoredChildren;
-    private volatile Collection<T> nonIgnoredChildren;
+    private volatile Collection<T> filteredChildren = null;
 
     private volatile RunnerScheduler scheduler = new RunnerScheduler() {
         public void schedule(Runnable childStatement) {
@@ -189,7 +187,7 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
      * @return {@code Statement}
      */
     protected Statement classBlock(final RunNotifier notifier) {
-        if (areAllFilteredChildrenIgnored()) {
+        if (areAllChildrenIgnored()) {
             return withNotificationOfIgnoredChildren(null, notifier);
         }
         Statement statement = childrenInvoker(notifier);
@@ -200,18 +198,13 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
         return statement;
     }
 
-    private boolean areAllFilteredChildrenIgnored() {
-        return getFilteredChildren().size() == getIgnoredFilteredChildren().size();
-    }
-
-    private Collection<T> getIgnoredFilteredChildren() {
-        getFilteredChildren();
-        return ignoredChildren;
-    }
-
-    private Collection<T> getNonIgnoredFilteredChildren() {
-        getFilteredChildren();
-        return nonIgnoredChildren;
+    private boolean areAllChildrenIgnored() {
+        for (T child : getFilteredChildren()) {
+            if (!isIgnored(child)) {
+                return false;
+            }
+        }
+        return true;
     }
 
     /**
@@ -225,9 +218,13 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
         return new Statement() {
             @Override
             public void evaluate() throws Throwable {
-                runChildren(getIgnoredFilteredChildren(), notifier,
-                        areAllFilteredChildrenIgnored());
-               if (null != next) {
+                for (T each : getFilteredChildren()) {
+                    // only notify methods annotated with @Ignore
+                    if (isIgnored(each)) {
+                        ParentRunner.this.runChild(each, notifier);
+                    }
+                }
+                if (null != next) {
                     next.evaluate();
                 }
             }
@@ -294,7 +291,7 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
         return new Statement() {
             @Override
             public void evaluate() {
-                runChildren(getNonIgnoredFilteredChildren(), notifier, true);
+                runChildren(notifier);
             }
         };
     }
@@ -310,20 +307,21 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
         return false;
     }
 
-    private void runChildren(Collection<T> children, final RunNotifier notifier, boolean isFinished) {
+    private void runChildren(final RunNotifier notifier) {
         final RunnerScheduler currentScheduler = scheduler;
         try {
-            for (final T each : children) {
-                currentScheduler.schedule(new Runnable() {
-                    public void run() {
-                        ParentRunner.this.runChild(each, notifier);
-                    }
-                });
+            for (final T each : getFilteredChildren()) {
+                // execute only, if the method is not annotated with @Ignore
+                if (!isIgnored(each)) {
+                    currentScheduler.schedule(new Runnable() {
+                        public void run() {
+                            ParentRunner.this.runChild(each, notifier);
+                        }
+                    });
+                }
             }
         } finally {
-            if (isFinished) {
-                currentScheduler.finished();
-            }
+            currentScheduler.finished();
         }
     }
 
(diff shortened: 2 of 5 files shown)


src/main/java/org/junit/internal/runners/statements/RunBefores.java
```diff
@@ -6,13 +6,24 @@
...
     private final List<FrameworkMethod> befores;
 
-    public RunBefores(Statement next, List<FrameworkMethod> befores, Object target) {
+    /**
+     * @param previous The previous statement, or {@code null}.
+     * @param next The next statement after before methods.
+     * @param befores The list of before methods.
+     * @param target
+     */
+    public RunBefores(final Statement previous, final Statement next,
```
kcooney: Unfortunately, `RunBefores` is a public class. Even though it's in an internal package, we should keep the previous constructor around (though we could deprecate it)

src/main/java/org/junit/internal/runners/statements/RunBefores.java
```diff
@@ -20,6 +31,9 @@ public RunBefores(Statement next, List<FrameworkMethod> befores, Object target)
 
     @Override
     public void evaluate() throws Throwable {
+        if (null != previous) {
```
kcooney: Is it reasonable for the caller to pass in a null value for previous into the constructor? If not, then the constructor should throw a `NullPointerException` if it is null.

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -104,6 +104,13 @@ protected TestClass createTestClass(Class<?> testClass) {
     protected abstract Description describeChild(T child);
 
     /**
+     * Fires test ignore through {@code notifier} for child corresponding to
+     * {@code child}, which can be assumed to be an element of the list returned
+     * by {@link ParentRunner#getChildren()}.
+     */
+    protected abstract void fireIgnoreChild(T child, RunNotifier notifier);
```
kcooney: We can't add new abstract methods to `ParentRunner`, because there's code outside of JUnit that extends `ParentRunner` directly. Would it be reasonable for this method to do nothing by default?

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -206,15 +215,33 @@ private boolean areAllChildrenIgnored() {
...
     /**
-     * Returns a {@link Statement}: run all non-overridden {@code @BeforeClass} methods on this class
+     * Returns a {@link Statement}: run all non-overridden {@code @Ignore} methods on this class
+     * and superclasses before executing {@code next}.
+     */
+    protected Statement withIgnores(final RunNotifier notifier) {
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                for (final T each : getFilteredChildren()) {
```
kcooney: style-nit: `each` doesn't need to be final

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -206,15 +215,33 @@ private boolean areAllChildrenIgnored() {
     }
 
     /**
-     * Returns a {@link Statement}: run all non-overridden {@code @BeforeClass} methods on this class
+     * Returns a {@link Statement}: run all non-overridden {@code @Ignore} methods on this class
+     * and superclasses before executing {@code next}.
```
kcooney: what does `next` mean here?

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -206,15 +215,33 @@ private boolean areAllChildrenIgnored() {
...
+
+    /**
+     * @param previous The previous statement, or {@code null}.
+     * @param next The next statement after methods annotated with {@link BeforeClass}.
+     * @return A {@link Statement}: run all non-overridden {@code @BeforeClass} methods on this class
      * and superclasses before executing {@code statement}; if any throws an
      * Exception, stop execution and pass the exception on.
      */
-    protected Statement withBeforeClasses(Statement statement) {
+    protected Statement withBeforeClasses(final Statement previous, final Statement next) {
```
kcooney: Again, we can't remove the existing `withBeforeClasses` method, because code outside of JUnit might call or override it.

also, style-nit: the params don't need to be final. We only make parameters/locals final if they need to be final in order to be used by an anonymous class

src/main/java/org/junit/internal/runners/statements/RunBefores.java
```diff
@@ -20,6 +31,9 @@ public RunBefores(Statement next, List<FrameworkMethod> befores, Object target)
...
+            previous.evaluate();
```
kcooney: This seem to be unrelated to the original intent of the `RunBefores` class. If you needed a previous statement to run, couldn't you wrap the `RunBefores` with another Statement that ran `previous` and then called the next `Statement` on the chain`? Perhaps create an abstract `DelegatingStatement` method if that makes sense.

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -285,11 +312,13 @@ private void runChildren(final RunNotifier notifier) {
         final RunnerScheduler currentScheduler = scheduler;
         try {
             for (final T each : getFilteredChildren()) {
-                currentScheduler.schedule(new Runnable() {
-                    public void run() {
-                        ParentRunner.this.runChild(each, notifier);
-                    }
-                });
+                if (!isIgnored(each)) {
```
kcooney: Please add a comment explaining why you are ignoring methods here (like "if the statement is ignored, a notification would have been (or would be) done via the code in someMethodOverThere")

src/main/java/org/junit/internal/runners/statements/RunBefores.java
OpenHelios: Yes, it's a good idea.

src/main/java/org/junit/internal/runners/statements/RunBefores.java
OpenHelios: With my patch the class BlockJUnit4ClassRunner uses null as value for previous. By providing the old constructor, previous will be null, too.

src/main/java/org/junit/internal/runners/statements/RunBefores.java
OpenHelios: I agree, the RunBefores class suggests a different behavior. My idea is to create a class like RunIgnores, which should execute the RunBefores as next statement.

src/main/java/org/junit/runners/ParentRunner.java
OpenHelios: It is possible to be compatible with a default implementation of fireIgnoreChild(), which calls runChild():
* The class BlockJUnit4ClassRunner is able to override both methods as intended. This means, new implementations have not to check, if the given child is ignored or not.
* Old implementations will have the new improved behavior, because they test already itself, if the given child is ignored or not.

src/main/java/org/junit/runners/ParentRunner.java
OpenHelios: Oops, this was a copy and past error. The method withIgnores do not have a parameter named state.

src/main/java/org/junit/runners/ParentRunner.java
OpenHelios: Yes, each doesn't need to be final. This is a safety programming approach, which prevents assigning each with a different content. If this is against your coding style, I will remove it.

src/main/java/org/junit/runners/ParentRunner.java
OpenHelios: No problem, it can be provided as an overloaded method with one more parameter to leave the original one untouched. Final can be removed, too.

src/main/java/org/junit/runners/ParentRunner.java
OpenHelios: Yes, adding a comment is a good idea.

src/main/java/org/junit/runners/ParentRunner.java
kcooney: You miss my point. Code outside of JUnit can currently create concrete subclasses ParentRunner. With your changes, those classes will have an abstract method, and won't compile

src/main/java/org/junit/runners/ParentRunner.java
kcooney: It is against our coding style. We generally only make locals final if they need to be in order to be accessed in an anonymous class. The extra "noise" of making the field final isn't worth the benefit

src/main/java/org/junit/internal/runners/statements/RunBefores.java
kcooney: A `RunIgnores` class sounds reasonable. Does that mean the changes to `RunBefores` can be reverted?

src/main/java/org/junit/internal/runners/statements/RunBefores.java
kcooney: If that's the case, the constructor should specify that the value can be `null`.

kcooney:
BTW, don't worry about squashing commits at this time; it actually makes it harder to review the changes in github. I can squash them when we merge this change.

src/main/java/org/junit/internal/runners/statements/RunBefores.java
OpenHelios: Yes, RunBefores can be reverted.

src/main/java/org/junit/runners/ParentRunner.java
OpenHelios: Yes, I see, you are right. My first suggestion was to move the abstract method to parent class Runner, but this could be a bad design, because fireIgnoreChild() and runChild() will then be declared in different classes.
I am now thinking of removing fireIgnoreChild(), because runChild() is enough for my purpose.

src/main/java/org/junit/runners/ParentRunner.java
OpenHelios: I understand.

OpenHelios:
This is my first pull requests. How can I add a second commit for this pull request?

UrsMetz:
@OpenHelios you just need to push your additional commits to the branch this pull request refers to (i.e. OpenHelios:ignorepatch), cf. https://help.github.com/articles/using-pull-requests/#initiating-the-pull-request.

kcooney:
@OpenHelios in your local repository, switch to the `ignorepatch` branch with `git checkout ignorepatch`. Make your changes locally, the rerun the test to make sure they still pass with `mvn test. Commit the changes locally, then push them to your github fork with `git push`

OpenHelios:
Thank you @UrsMetz and @kcooney. I pushed my second commit.

OpenHelios:
In my second commit, I decided to add a parameter named next to the withIgnores() method. Furthermore I have removed the fireIgnoreChild() method, because it is technically not needed. In this way my pull request changes only the ParentRunner class. The added test has not been changed.

I can add the fireIgnoreChild() method again without the key word abstract, if you prefer it for a clearer design. Nevertheless without the abstract key word the method loses it's interface character. What do you suggest?

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -206,6 +208,29 @@ private boolean areAllChildrenIgnored() {
...
+     * Returns a {@link Statement}: notify all non-overridden {@code @Ignore} methods on this class
+     * and superclasses before executing {@code next}.
+     * @param next The next statement, or {@code null}, if there is no next statement.
+     * @param notifier The runner notifier.
+     */
+    protected Statement withIgnores(final Statement next, final RunNotifier notifier) {
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                for (final T each : getFilteredChildren()) {
```
kcooney: I can imagine cases where users of JUnit expect the notifier to be called from the scheduler (say, if the scheduler sets a thread local variable).

Could we just change `runChildren` to take in an `Iterable<T>` and use it here?

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -206,6 +208,29 @@ private boolean areAllChildrenIgnored() {
     }
 
     /**
+     * Returns a {@link Statement}: notify all non-overridden {@code @Ignore} methods on this class
+     * and superclasses before executing {@code next}.
+     * @param next The next statement, or {@code null}, if there is no next statement.
+     * @param notifier The runner notifier.
+     */
+    protected Statement withIgnores(final Statement next, final RunNotifier notifier) {
```
kcooney: This name is confusing. Maybe `withNotificationOfIgnoredChildren`?

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -187,12 +187,14 @@ private void validateClassRules(List<Throwable> errors) {
...
-        if (!areAllChildrenIgnored()) {
-            statement = withBeforeClasses(statement);
-            statement = withAfterClasses(statement);
-            statement = withClassRules(statement);
+        if (areAllChildrenIgnored()) {
+            return withIgnores(null, notifier);
         }
+        Statement statement = childrenInvoker(notifier);
+        statement = withBeforeClasses(statement);
+        statement = withIgnores(statement, notifier);
```
kcooney: shouldn't this be the last thing you do before returning the statement chain? If the class rules or `@AfterClass` methods throw, don't you want the ignored methods to be notified?

src/test/java/org/junit/tests/running/methods/TestMethodTest.java
```diff
@@ -251,5 +251,29 @@ public void onlyIgnoredMethodsIsStillFineTestClass() {
         Result result = JUnitCore.runClasses(OnlyTestIsIgnored.class);
         assertEquals(0, result.getFailureCount());
         assertEquals(1, result.getIgnoreCount());
+        assertEquals(0, result.getRunCount());
     }
+
+    public static class OneTestIsIgnoredAndOneIsIncomplete {
+        @BeforeClass
+        public static void throwException() throws Exception {
```
kcooney: Could you add another test where the exception is thrown by an `@AfterClass` method?

src/test/java/org/junit/tests/running/methods/TestMethodTest.java
```diff
@@ -251,5 +251,29 @@ public void onlyIgnoredMethodsIsStillFineTestClass() {
...
+            throw new Exception();
+        }
+        @Ignore
+        @Test
+        public void ignored() {
+        }
+        @Test
```
kcooney: style-nit: blank line between methods (i.e. before this line)

src/main/java/org/junit/runners/ParentRunner.java
OpenHelios: Yes, good idea.

src/main/java/org/junit/runners/ParentRunner.java
OpenHelios: Yes, this should reduce code duplication.

src/main/java/org/junit/runners/ParentRunner.java
OpenHelios: * Yes, I want the ignored methods to be notified. At the first look, the original implementation on master is confusing, unless you look into implementation of the classes RunBefores and RunAfters. RunBefores execute the given statement *after* the @BeforeClass methods in contrast to RunAfters, which execute the given statement *before* the @AfterClass methods. I will add the suggested test to show that it works.
* No, I am not worried about sending ignore notifications before running the non ignored methods. The order of the running methods, which are not ignored, is the same.

src/main/java/org/junit/runners/ParentRunner.java
kcooney: Forget the Predicate idea. A `ignored` parameter would be simpler

OpenHelios: This sounds like a good use case for a collection library, i.e. a method like http://google-collections.googlecode.com/svn/trunk/javadoc/com/google/common/collect/Iterables.html#filter%28java.lang.Iterable%2C%20com.google.common.base.Predicate%29

Nevertheless the private implementation of runChildren() uses a private scheduler implementation, which do nothing else then creating own threads for each test method. These threads make sense for test methods, which are not ignored. The notification of ignored methods should run fast. Would you say, that the notified listeners can consume a lot of time, which perform better with threads?

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -206,6 +208,29 @@ private boolean areAllChildrenIgnored() {
...
+     * and superclasses before executing {@code next}.
+     * @param next The next statement, or {@code null}, if there is no next statement.
+     * @param notifier The runner notifier.
+     */
+    protected Statement withIgnores(final Statement next, final RunNotifier notifier) {
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                for (final T each : getFilteredChildren()) {
+                    // only notify methods annotated with @Ignore
```
kcooney: comment is out of date

src/main/java/org/junit/runners/ParentRunner.java
kcooney: @OpenHelios while I love Guava, we don't want to require a new third-party library for all users of JUnit.

As I said before, for backwards compatibility reasons, I think we should call the notifier via the scheduler. Unless there are a lot of ignored methods, the performance overhead of using the scheduler for notification of ignored tests shouldn't be noticeable. If you could show data that indicates otherwise, we can reconsider that.

kcooney:
@OpenHelios could you look into the test failures? Thanks!

OpenHelios:
All tests run in my committed 6. I do not fully understand currently, why the test fails, if the ignored methods are notified from scheduler before executing the BeforeClass methods - without the scheduler it is no problem to notify ignored methods before executing the BeforeClass methods.

kcooney:
@OpenHelios I'll take a look at the test failures in the previous version of this change. If we can't fix this unless the ignored methods are not called from the scheduler, then I'm personally tempted to move this to the JUnit 4.0 branch, where we are allowing changes that might not be compatible with 4.12 (there may not be a 4.13, so the release of the change might happen the same time regardless).

Otherwise, this change looks good to me. @junit-team/junit-committers can one of you take a second look?

kcooney:
@OpenHelios I think the reason why the tests broke was that the code you had to memoize the ignored filtered tests broke the filtering code. I propose an alternative fix in #1093 (see the last commit there). If you like it as-is, I can commit it along with your changes. If you want to make minor changes from my changes, you can merge in my commit into your branch.

OpenHelios:
Your enhanced change shows me how the scheduler needs the execution sequence.

kcooney:
@OpenHelios fantastic! I think the simplest path forward would be to get *this* pull merged (I'll squash the commits) and then I'll merge mine. I just would like to get a second opinion from one of the other owners.

@junit-team/junit-committers you can look at #1093 to see what the final result will be. I would like to get a second set of eyes on this when one of you has the chance.

dsaff:
I can see the changed test, but can I get a one-line description for why the behavior change is important and desirable?  Thanks!

OpenHelios:
Here is my one liner:
Test methods annotated with @Ignore should be reported with status ignored, independently of the result of all other test methods.

kcooney:
@OpenHelios it should be noted that if you have a suite with a `@BeforeClass` method, and that suite contains a test class with ignored methods, I think you wouldn't get notified of the ignore status event with this change.

OpenHelios:
@kcooney for solving your mentioned case, I have created an other branch, merged your changes into it and added a new commit. A further test is included in the commit to show that an ignored method is notified, if there is an exception in the BeforeClass method of the test suite class.
I do not know, if this is the correct way for adding an other commit on your branch, but a diff can be seen here https://github.com/kcooney/junit/compare/pull1091...OpenHelios:pull1093

kcooney:
@OpenHelios doing an `instanceof` check to see if the child runner is a `ParentRunner` is a bit ugly and looks fragile, so I wouldn't support those changes.

OpenHelios:
Yes, using `instanceof` is my last choice. A better design should use an interface, but this can not be added without braking the backward compatiblity. For now, I am totally happy with your last version of #1093, i.e. without my last commit.
Maybe for the next interface braking version, suites should catch exceptions in `BeforeClass` methods to notify all ignored methods, too.

dsaff:
It might help me to open a bug/feature request that this can relate to.  If I'm understanding, given a class with two test methods, one ignored, if BeforeClass throws an exception:

Behavior at HEAD:
- one failure reported at the class node, no feedback about methods

Behavior after patch:
- one failure reported at the class node, one ignored status posted about the ignored method

To my knowledge, there isn't a clear principle in the documentation that would prefer one of these over the other.  Do you have an experience that would help me understand why the second is desirable?  Thanks!

OpenHelios:
@dsaff Your described behavior is totally correct.

The behavior after the patch is desirable, because in my use case I want to compare two test runs of the same test suite in different environments, where the number of ignored test methods should obviously be the same. But depending on the environment, there are different test results - with and without exceptions in the `BeforeClass` methods of some test classes. With the behavior at HEAD it is not possible to validate the number of ignored methods correctly, because not all methods annotated with `Ignore` can be found in the test results (both exported with the JUnit view in Eclipse).

For my described use case, I wrote a compare tool using two test result XML files, which showed me this issue. The number and the list of ignored methods must be the same in both test results. In this way, it is a kind of sanity check. Otherwise there must be a mistake caused by different versions of the same test suite.

Nevertheless in the JUnit view of Eclipse, it would be useful to see ignored methods independent from having an exception in the `BeforeClass` methods of a test class, too. The patch makes it possible.

dsaff:
I'm sorry to say that I'm still not convinced that the benefits outweigh the costs.  

On the benefit side, the number of succeeding and failing tests changes when there's a failure in BeforeClass, as well, right?  Isn't that at least as much of a concern as the number of ignored tests?

On the cost side, since ParentRunner is a base class, it has documented behavior, including that "Run all non-overridden {@code @BeforeClass} methods on the test-class and superclasses; if any throws an Exception, stop execution and pass the exception on."

This patch must at least change that documentation, since execution is _not_ stopping for the ignored methods.  Also, subclasses of ParentRunner may be (explicitly or implicitly) depending on the fact that a failure in a @BeforeClass means that no more calls to runChild will be coming, and this change would break them.

Note that I would consider the costs of this behavior change at most medium, but since the benefit seems even less than that, I'm disinclined to support it.

Of course, I may very well have missed something causing me to underestimate the benefit, and I'm open to learning more!

kcooney:

Closing based on David's concerns.

PR #1093, opened 2015-03-08
Minor changes on time of pull #1091

kcooney:
Don't merge this. I'm just creating it to propose one change to #1091 (creating it as a pull ensures that the Travis build is run)

### DIFF
--- a/src/main/java/org/junit/runners/ParentRunner.java
+++ b/src/main/java/org/junit/runners/ParentRunner.java
@@ -104,6 +104,13 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
     protected abstract Description describeChild(T child);
 
     /**
+     * Fires test ignore through {@code notifier} for child corresponding to
+     * {@code child}, which can be assumed to be an element of the list returned
+     * by {@link ParentRunner#getChildren()}.
+     */
+    protected abstract void fireIgnoreChild(T child, RunNotifier notifier);
+
+    /**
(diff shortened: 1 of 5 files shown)


OpenHelios:
Great, this enhanced change looks good for me. I am looking forward to see it in the master branch.

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -206,6 +209,29 @@ private boolean areAllChildrenIgnored() {
...
+            public void evaluate() throws Throwable {
+                try {
+                    statement.evaluate();
+                } catch (Throwable e) {
+                    runChildren(notifier, new ChildSelector<T>() {
+                        public boolean include(T child) {
+                            return isIgnored(child);
+                        }
+                    });
+                    throw e;
```
OpenHelios: In the catch part, calling runChildren() could raise a second exception, which should be catched, too. Otherwise the information of the previous catched exception is lost.

kcooney: @OpenHelios I believe that exceptions thrown while running the children should be reported via the notifier, and not result in another exception. If we're paranoid, we could update runChildren() to do it's own try...catch and do it's own notification.

OpenHelios: I agree, this should only happen, if the runChildren(), runChild() or an overridden one contains implementation errors.

kcooney:

Closing based on David's concerns on #1091 1091

PR #1096, opened 2015-03-10
Fixes some SonarQube issues

hansjoachim:
- Moved array designators from the variable to the type
- Prefer interfaces over concrete classes. Note, there were a couple more cases where interfaces could be used instead, but some of them were part of public APIs so I haven't touched them for compatibility reasons.
- And removed an unnecessary semicolon

### DIFF
--- a/src/main/java/junit/textui/TestRunner.java
+++ b/src/main/java/junit/textui/TestRunner.java
@@ -131,7 +131,7 @@ public class TestRunner extends BaseTestRunner {
         }
     }
 
-    public static void main(String args[]) {
+    public static void main(String[] args) {
         TestRunner aTestRunner = new TestRunner();
         try {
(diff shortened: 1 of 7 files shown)


stefanbirkner:
I like the first two commits, but I don't think it is necessary to enforce interfaces within a single method. @junit-team/junit-committers what's your opinion?

kcooney:
I don't feel strongly about whether we should enforce interfaces within a single method, but to make this change less controversial, let's remove the last commit.

dsaff:
I do find that having interfaces in the variable declaration makes code easier to scan by removing unnecessary noise, so I'd be OK taking the whole thing (but wouldn't hold up the rest of the commit in order to win on this point.)

stefanbirkner:
Merged all three commits. @dsaff made a good case for using interfaces.

PR #1097, opened 2015-03-12
Add overload of Request.sortWith() which takes a Sorter

BartOutOfTheBox:
Updated SortingRequest so it can be instantiated with a given Sorter. Some kind of custom made sorter that takes more than just a comparator as argument can now be used.

### DIFF
--- a/src/main/java/org/junit/runner/Request.java
+++ b/src/main/java/org/junit/runner/Request.java
@@ -8,6 +8,7 @@ import org.junit.internal.requests.FilterRequest;
 import org.junit.internal.requests.SortingRequest;
 import org.junit.internal.runners.ErrorReportingRunner;
 import org.junit.runner.manipulation.Filter;
+import org.junit.runner.manipulation.Sorter;
 import org.junit.runners.model.InitializationError;
 
 /**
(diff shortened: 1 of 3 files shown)


kcooney:
I have no objections to this change, but `SortingRequest` is in an internal package, so it may be better not to depend on it

kcooney:
...though if we merge this a test would be nice

BartOutOfTheBox:
@kcooney any concerns I missed?

kcooney:
This looks good to me (we can squash the commits when we merge). @junit-team/junit-committers any objections?

@BartOutOfTheBox I am slightly curious why you want this change, since a `Sorter` can't sort by anything other than a `Compator`

BartOutOfTheBox:
I was thinking about making a subclass of Sorter which would take other Sorters as input, and sort tests using these other sorters in a fair way. While this can be achieved by making a more complex Comparator, I thought it may be more neat to do it with Sorters directly. What would be the the goal of having a Sorter class if they're only just a wrapper for comparators anyways.

kcooney:
@BartOutOfTheBox would something like https://github.com/junit-team/junit/compare/master...kcooney:ordering be better for you?

src/main/java/org/junit/runner/Request.java
```diff
@@ -166,4 +167,15 @@ public Request filterWith(final Description desiredDescription) {
...
     }
+    
+    /**
+     * Returns a Request whose Tests can be run in a certain order, defined by
+     * <code>sorter</code>
+     *
+     * @param sorter the sorter determining the order of the tests in this Request
+     * @return a Request with ordered Tests
+     */
+    public Request sortWith(Sorter sorter) {
```
kcooney: missing `@since` Javadoc

kcooney:

This seems reasonable even if my pull was ever merged, so let's try to get it in.

src/main/java/org/junit/runner/Request.java
kcooney: I just realized that there is no need for this method, since Sorter implements `Comparator<Description>`

PR #1099, opened 2015-03-13
refactory applying clean code

AlisonSouza:


kcooney:
Why are you changing the code to set `boolean` fields to `Boolean` values (forcing am implicit auto-unbox)?

AlisonSouza:
Both are unboxed, I believe Boolean.TRUE is more elegant than true,
it's as Bigdecimal, Example:

BigDecimal b1 = new BigDecimal(10);
BigDecimal b2 = BigDecimal.TEN;

Both are correct, but I think BigDecimal.TEN is more legible

What do you think about?


Sorry my english, I am from Brazil

Tibor17:
`boolean b = Boolean.TRUE` b*l*s*t  
Setting unboxed object to primitive is 100 years before monkeys.

kcooney:
`Boolean.TRUE` is a Boolean, not a boolean. I think I prefer the original code.

PR #1105, opened 2015-03-30
JavaDoc Clean-up Rules Implement TestRule

EarthCitizen:
All of the rules removed now implement TestRule instead of MethodRule.

### DIFF
--- a/src/main/java/org/junit/rules/MethodRule.java
+++ b/src/main/java/org/junit/rules/MethodRule.java
@@ -14,14 +14,7 @@ import org.junit.runners.model.Statement;
  * or write your own:
  *
  * <ul>
- *   <li>{@link ErrorCollector}: collect multiple errors in one test method</li>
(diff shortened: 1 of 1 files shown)


stefanbirkner:
LGTM

src/main/java/org/junit/rules/MethodRule.java
```diff
@@ -14,14 +14,7 @@
  * or write your own:
```
marcphilipp: Instead of using an unnumbered list with a single item, IMHO we should re-formulate the last sentence of this paragraph and include a link to `TestWatchman` there. What do you think?

marcphilipp:
@EarthCitizen Do you have time to take care of my line note or do you want me to do it myseld?

EarthCitizen:
@marcphilipp I seem to have deleted my fork for this patch. Is there a way I can edit this patch after doing that? Otherwise I may have to delete this and start over.

EarthCitizen:
Replaced this with #1143

PR #1107, opened 2015-03-31
Add Ordering and Orderable.

kcooney:
These APIs allow arbitrary ordernig of tests, including randomization.


kcooney:
I need to close this; will resubmit hopefully next week

PR #1111, opened 2015-04-04
[Issue 1110] StopWatch does not need to be Abstract

Tibor17:
+ removed "abstract" class modifier
+ referenced surefire version via property
+ POM schema split in 3 lines


kcooney:
Could you make this two commits?

Tibor17:
Rework takes more time.
Is it so bad that it has to be two commits or two PRs?

kcooney:
Two commits please. IMHO, a commit should have one purpose. One PR is fine.

PR #1115, opened 2015-04-07
Test methods can be parameterized

krueger:
I extended the parameterized classes such that the test method can be parameterized:

@RunWith(Parameterized.class)
public class JunitTest {
	
	@Parameters
    public static Collection<Object[]> data() {
        return Arrays.asList(new Object[][] {     
                 { 0, 0 }, { 1, 1 }, { 2, 1 }, { 3, 2 }, { 4, 3 }, { 5, 5 },{ 6, 8 }  
           });
    }
    
    @Test
    public void test(int i, int j) {
    	System.out.println("i=" + i + " j=" + j);
    }

}

### DIFF
--- a/src/main/java/org/junit/runners/Parameterized.java
+++ b/src/main/java/org/junit/runners/Parameterized.java
@@ -24,24 +24,25 @@ import org.junit.runners.parameterized.TestWithParameters;
  * cross-product of the test methods and the test data elements.
  * <p>
  * For example, to test a Fibonacci function, write:
+ * 
  * <pre>
  * &#064;RunWith(Parameterized.class)
  * public class FibonacciTest {
- *     &#064;Parameters(name= &quot;{index}: fib[{0}]={1}&quot;)
+ *     &#064;Parameters(name = &quot;{index}: fib[{0}]={1}&quot;)
  *     public static Iterable&lt;Object[]&gt; data() {
  *         return Arrays.asList(new Object[][] { { 0, 0 }, { 1, 1 }, { 2, 1 },
  *                 { 3, 2 }, { 4, 3 }, { 5, 5 }, { 6, 8 } });
  *     }
- *
+ * 
  *     private int fInput;
- *
+ * 
  *     private int fExpected;
- *
+ * 
  *     public FibonacciTest(int input, int expected) {
- *         fInput= input;
- *         fExpected= expected;
+ *         fInput = input;
+ *         fExpected = expected;
  *     }
- *
+ * 
  *     &#064;Test
  *     public void test() {
  *         assertEquals(fExpected, Fibonacci.compute(fInput));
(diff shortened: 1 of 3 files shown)


marcphilipp:
Can you please take a look and see why the build has failed?

stefanbirkner:
@krueger The same question again ;-) What problem the feature is expected to solve? Sorry for being petty, but this information helps a lot in evaluating the pull request.

krueger:
@stefanbirkner I don't want to add this code in the test class

```
@Parameter
private int i;
    
@Parameter(1)
private int j;
```

krueger:
@marcphilipp I have a look

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -253,9 +258,11 @@ private RunnersFactory(Class<?> klass) {
...
             Parameters parameters = getParametersMethod().getAnnotation(
                     Parameters.class);
-            return Collections.unmodifiableList(createRunnersForParameters(
-                    allParameters(), parameters.name(),
-                    getParametersRunnerFactory()));
+            ParametersRunnerFactory runnerFactory = getParametersRunnerFactory();
+            Iterable<Object> allParameters = allParameters();
+            List<Runner> runners = createRunnersForParameters(allParameters,
+                    parameters.name(), runnerFactory);
+            return Collections.unmodifiableList(runners);
```
stefanbirkner: Please separate refactoring commits from commits that change the classes behaviour (aka feature commits). This makes it a lot easier to understand the commit, because the reader doesn't have to find out whether your change is modifying the behaviour or not.

stefanbirkner:
Your pull request breaks the current behaviour of the Parameterized runner. Please fix this.

stefanbirkner:
Do you know the [JUnitParamsRunner](http://pragmatists.github.io/JUnitParams/)? Maybe it is already solving your problem.

src/main/java/org/junit/runners/parameterized/BlockJUnit4ClassRunnerWithParameters.java
```diff
@@ -73,6 +87,30 @@ private Object createTestUsingFieldInjection() throws Exception {
...
+        }
+    }
+    
+    @Override
+    protected void validateTestMethods(List<Throwable> errors) {
+        List<FrameworkMethod> methods = getTestClass().getAnnotatedMethods(Test.class);
+
+        for (FrameworkMethod eachTestMethod : methods) {
+            eachTestMethod.validatePublicVoid(false, errors);
+        }
```
stefanbirkner: This should delegate to super by default, but additionally check whether the test methods have the correct parameters in case `useTestInjection` is `true`.

krueger: Good remark. I fixed it.

kcooney:
Although I appreciate the contribution, I I don't think it makes sense to pass the parameters via the test method if the parameters are specified via a static method. If we add support for passing parameters via the test method, I would prefer something similar to https://github.com/Pragmatists/junitparams

I also personally don't see removing the fields as  being a big win here.

krueger:
???

Sorry, but I really get strange or no arguments from you. This makes it hard for me to argue.
It's your project. Do what ever you want with the code.

Greetings,
Martin

kcooney:
Adding fields is quite trivial. Passing the values via the test methods requires duplication in each test method, which makes the test class code harder to maintain (if you change the number of parameters, every method would need to have an updated signature). Since the number and type of the parameters need to be the same (specified via a static method) I also think it is confusing to pass the parameters via the test method.

I could be convinced otherwise, but I would need a bit more justification than was given here.

marcphilipp:
I would rather not have this merged. 

@junit-team/junit-committers What are your opinions on this?

ashleyfrieze:
I would like this sort of capability, but it doesn't make sense for me to have it in Parameterized. The whole test class is parameterized with this runner. A parameterized test implies something like kcooney's link - https://github.com/Pragmatists/junitparams. Even that's imperfect. NUnit's TestCase feature is much better - http://www.nunit.org/index.php?p=testCase&r=2.5, but not possible in Java owing to restrictions in what can be in an annotation (only compile-time constants, not objects) and the fact that you cannot have multiple annotations of the same type (in most versions of Java).

I'd like to see this sort of syntax enter the BlockJUnit4Runner, so it can be shared more widely. I had created that change, but it's a biggy. If people are interested, I could remake it in a separate pull request.

kcooney:
I also would rather not have this merged

dsaff:
My personal preference is to mark Parameterized as no longer open to extension: there are several third-party parameterized runners (JUnitParams being my favorite the last time I looked) with much better interfaces and functionality, which are not hampered by needing to be backwards compatible with Parameterized.

I would even consider making a wiki page with pointers to some of these runners, and linking to it from the javadoc of Parameterized.

PR #1116, opened 2015-04-07
Added TestCaseBuilder

krueger:


test_case_builder
```diff
@@ -0,0 +1 @@
+Builder class for parameterized tests
```
marcphilipp: What's the purpose of this file?

krueger: It was created when I created the fork. I will delete it.

marcphilipp:
I've written similar builders in a few projects, but I've found a general purpose implementation that takes `Object...` as parameters to be of limited use. It has the same pitfall as declaring nested arrays: no compile-time validation. 

If we decide to include such a builder we should IMHO experiment with some kind of validation, e.g. passing `Class<?>...` to the builder constructor and use it to validate the `testCase` arguments. We could also add support for method-chaining.

stefanbirkner:
@krueger What problem the `TestCaseBuilder` is expected to solve?

krueger:
@stefanbirkner creating the test data like this

```
@Parameters
public static Collection<Object[]> data() {
    return Arrays.asList(new Object[][] {     
             { 0, 0 }, { 1, 1 }, { 2, 1 }, { 3, 2 }, { 4, 3 }, { 5, 5 },{ 6, 8 }  
    });
}
```
isn't very readable. So I thought about creating a testcase builder. I think this looks a little bit better

```
@Parameters
public static Collection<Object[]> data() {
	TestCaseBuilder builder = new TestCaseBuilder();
	builder.testCase(0, 0);
	builder.testCase(1, 1);
	builder.testCase(2, 1);
	builder.testCase(3, 2);
	builder.testCase(4, 3);
	builder.testCase(5, 5);
	builder.testCase(6, 8);
	
	return builder.testData();
}
```

but as marcphillip said before, this approach does not handle the parameter types. I also think the method data() should not return Collection<Object[]> but an interface which can be handled by the framework.

```
@Parameters
public static TestCaseFactory data() {
	TestCaseBuilder builder = new TestCaseBuilder();
	builder.testCase(0, 0);
	builder.testCase(1, 1);
	
	return builder;
}
```
where TestCaseBuilder implements an interface TestCaseFactory.

Another approach could be

```
@Parameters
public static Collection<Object[]> data() {
	TestCaseBuilder<SimpleTestCase> builder = new TestCaseBuilder<SimpleTestCase>();
		builder.add(new SimpleTestCase(0, 0));
		builder.add(new SimpleTestCase(0, 1));
		
		return builder.testData();
}

public class SimpleTestCase extends ParameterizedTestCase{
	public SimpleTestCase(int i, int j) {
		super(i, j);
	}
}

public class ParameterizedTestCase {	
	private Object[] params;

	public ParameterizedTestCase(Object... params) {
		this.params = params;
	}
	
	public Object[] getParams() {
		return params;
	}
}
```

stefanbirkner:
IMHO the most readable way for creating test data is currently

    @Parameters
    public static Object[][] data() {
      return new Object[][] {     
        { 0, 0 },
        { 1, 1 },
        { 2, 1 },
        { 3, 2 },
        { 4, 3 },
        { 5, 5 },
        { 6, 8 }
      });
    }

It is probably more readable than the `TestDataBuilder` because it needs less boilerplate code. How do you think about this?

(I leave the discussion about interfaces and verification for later. First I would like to make up my mind whether the current `TestDataBuilder` is helpful or not.)

krueger:
I like using a TestCaseBuilder rather then using an object array but perhaps this is only my personal point of view. If the biolerplate code annoys you, this proposal could be better:

```
@Parameters
public static void data(TestBuilder b) {
	b.test(0, 0, 0);
	b.test(0, 1, 1);
	b.test(2, 2, 4);
}
```

stefanbirkner:
I close this pull request without merging it. IMHO it doesn't provide enough additional value that justifies extending the JUnit API. If you disagree, please reopen the pull request.

PR #980, opened 2014-09-02
Better timeout thread names

cfontes:
Pull request related to #799, this is a try to have more meaningful names on the timeout test threads.

Issue #799, opened 2014-01-10
Thread created with FailOnTimeout should be named sometime more useful than "Thread-#"


mkw:
When debugging or profiling highly threaded tests, leaving the JUnit timeout threads as "Thread-#" is not very useful.

The obvious name for threads would probably be the full name for the test.  This could be passed in from both BlockJUnit4ClassRunner#withPotentialTimeout, which has access to the Method object, and Timeout#apply, which has access to the Description object.

The fix for #657 names these threads "Time-limited test," which somewhat more useful, but still ambiguous when a large number of tests are run.

modocache:
An attempt to address this issue is being made in #1117 :point_up:

adrianosimoes:
The pull request originated from this issue was closed. Should this issue also be closed?

kcooney:
@adrianosimoes the pull was closed because of unresolved comments, not because we wouldn't want thsi feature.

junit-builds:
JUnit 4 is now in maintenance mode.

At this point, only critical bugs and security issues will be fixed.

The team has therefore decided to close this issue.

PR #1117, opened 2015-04-08
Better timeout thread names

modocache:
Builds on work done in #980, and addresses comments made on that pull request.

See individual commit messages for details.


src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
```diff
@@ -328,14 +328,14 @@ protected Statement possiblyExpectingExceptions(FrameworkMethod method,
      * takes more than the specified number of milliseconds.
      */
     @Deprecated
-    protected Statement withPotentialTimeout(FrameworkMethod method,
-            Object test, Statement next) {
+    protected Statement withPotentialTimeout(FrameworkMethod method, Statement next) {
```
stefanbirkner: Still not fixed: Changing the signature of withPotentialTimeout() is likely to break someone.

src/main/java/org/junit/rules/Timeout.java
```diff
@@ -143,16 +143,17 @@ protected final boolean getLookingForStuckThread() {
      * @since 4.12
      */
     protected Statement createFailOnTimeoutStatement(
-            Statement statement) throws Exception {
+            Statement statement, Description description) throws Exception {
```
stefanbirkner: Unfortunately this method is protected. Changing the signature may break someone.

modocache: > I'm in favour of modifying the methods signature and making it private. Users should create a new rule instead of extending Timeout. They could use the builder for creating the statement.

Agree on all points here!

> We could not add the new feature without breaking the extension point.

We can mark the old method as `@Deprecated`, and add a new, private method with a signature that takes a `Description`. The old method would work as before, not providing a `Description` to the `FailOnTimeout`--thus threads created via the old method would be given generic names like `"Time-limited test"`.

src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
modocache: Sorry, this change is likely due to poor merge conflict resolution on my part. I'll change this back. :wave:

src/main/java/org/junit/rules/Timeout.java
modocache: > We can mark the old method as `@Deprecated`, and add a new, private method with a signature that takes a `Description`. The old method would work as before, not providing a `Description` to the `FailOnTimeout`--thus threads created via the old method would be given generic names like `"Time-limited test"`.

Oops, no we can't--`Timeout#apply` calls `Timeout#createFailOnTimeoutStatement`. To keep both signatures of that method around we'd have to provide a hook for users to configure the behavior of `Timeout#apply`.

I'm in favor of making the method private, but how does this jive with the [deprecation policy](https://github.com/junit-team/junit/wiki/Deprecation-Policy)?

kcooney: We can't change the behavior or signature of the existing method. We could deprecate it (but not change or break it) and remove it later in accordance with our depreciation policy.

Often we get around this problem by creating a new protected or package scope method, and having the new method call the previous one. Sunglasses can override the new method. Note the Builder could create insrltances of a subclass of `Timeout`

modocache:
> Often we get around this problem by creating a new protected or package scope method, and having the new method call the previous one.

@kcooney @stefanbirkner I'm hoping 1ff5f23 does something similar to what you're describing here. :open_hands:

stefanbirkner:
IMHO commit 1ff5f23e53154e47c7e35a15043c91b990cfce4c is more worse than changing the methods signature (and maybe its visibility), because it breaks things silently.

A developer who created a modified version of Timeout by implementing `createFailOnTimeoutStatement(Statement)` does not get any warning if he upgrades to JUnit 4.13, although its own implementation of `createFailOnTimeoutStatement(Statement)` is not used anymore. IMHO this is the opposite of fail-fast and the developer may have wrong assumptions about his tests.

We have three options:
* Use 1ff5f23e53154e47c7e35a15043c91b990cfce4c with the disadvantages mentioned above.
* Make the method private and change its signature. Breaks API compatibility at compile-time.
* Not implementing the new feature.

Maybe there are other options that I'm not aware of.

modocache:
> Not implementing the new feature.

I'm actually fine with this, since it seems like the reaction to #799 was lukewarm in the first place. I just picked up the task because it was marked `lowhanging`.

> Often we get around this problem by creating a new protected or package scope method, and having the new method call the previous one. Sunglasses can override the new method. Note the Builder could create insrltances of a subclass of `Timeout`

Does 1ff5f23e53154e47c7e35a15043c91b990cfce4c not implement this correctly? Or are you against this approach in general, @stefanbirkner?

stefanbirkner:
The implementation is correct, but it would silently change the behaviour of classes that extend Timeout and override `createFailOnTimeoutStatement(Statement)`. Yet it disables exactly the intentionally modified behaviour.

kcooney:
I will try to find time to look at this next week (after I return from vacation) to see if I can make any suggestions.

modocache:
Awesome, thanks, @kcooney and @stefanbirkner! :+1:

src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
```diff
@@ -103,6 +107,20 @@ public Builder withLookingForStuckThread(boolean enable) {
...
+         * Specifies the description of this test. If this is provided, the
+         * thread generated when evaluating this statement will be given a
+         * name corresponding with the description.
+         *
+         * @param description a description of the test, to be used when
+         *                    naming the generated thread
+         * @return {@code this} for method chaining.
+         */
+        public Builder withDescription(Description description) {
+            this.description = description;
```
kcooney: I think it would be simpler to call `description.getDisplayName()` here, and replace the "description" field with a "threadName" field of type `String` with a default value of "Time-limited test"

src/main/java/org/junit/rules/Timeout.java
```diff
@@ -163,6 +162,19 @@ public Statement apply(Statement base, Description description) {
...
+     * Creates a {@link Statement} that will run the given
+     * {@code statement} in a thread with a name based on
+     * the given {@code description}, and timeout the operation
+     * based on the values configured in this rule.
+     */
+    private Statement createFailOnTimeoutStatement(
+            Statement statement, Description description) throws Exception {
+        return FailOnTimeout.builder()
+            .withDescription(description)
+            .build(createFailOnTimeoutStatement(statement));
```
kcooney: I'm confused... isn't this wrapping a `Statement` that applies the timeout inside of another `Statement` that applies a timeout?

src/main/java/org/junit/rules/Timeout.java
```diff
@@ -152,7 +151,7 @@ protected Statement createFailOnTimeoutStatement(
 
     public Statement apply(Statement base, Description description) {
         try {
-            return createFailOnTimeoutStatement(base);
+            return createFailOnTimeoutStatement(base, description);
```
kcooney: Couldn't you just add a `Description` field and set it here, and then pick up the value of that field in `createFailOnTimeoutStatement()`? We wouldn't expect a `Timeout` instance to be shared between two test methods, so that should be safe. Ugly, but safe :-)

modocache:
Sorry for the noise here! I'd love to contribute to JUnit one day, but this looks like it was a false start. Thanks for the review, and sorry to take up your time!

Issue #444, opened 2012-05-31
JUnit should have an exhaustive listener framework


brolund:
JUnit should have a listener model that provides all the events from test execution (run starting, suite starting, test case starting, test starting, failed, error, success, timeout, theories data, assumption failed etc etc), including a reference to the test/suite class and a Method object for the method at hand.

It should also allow for hooking in the listeners in a single place for an entire set of tests in an unobtrusive way, like a properties file (or possibly as an annotation on a suite, or a specific class in the execution tree). 

Rules are sort of listeners, but they require more detailed manipulations of tests. RunListener provides only a fragment of the available information. If the listener framework can also change the test execution (AOP-style), the Rule implementation, as well as the RunListener, could be build on top of this framework.

PR #1118, opened 2015-04-08
Add suite start/finish events to listener

baev:
Hi guys! 

This pull request adds ability to handle suite start/finish events via `RunListener`.

Maybe better to use `testStarted` and `testFinished` events instead new ones but I am afraid to break backwards compatibility. Tests will be added later. 

What are you guys think?

### DIFF
--- a/src/main/java/org/junit/internal/runners/model/EachTestNotifier.java
+++ b/src/main/java/org/junit/internal/runners/model/EachTestNotifier.java
@@ -45,4 +45,27 @@ public class EachTestNotifier {
     public void fireTestIgnored() {
         notifier.fireTestIgnored(description);
     }
+
+    /**
+     * Calls {@link RunNotifier#fireTestSuiteStarted(Description)}, passing the
+     * {@link Description} that was passed to the {@code EachTestNotifier} constructor.
+     * This should be called when a test suite is about to be started.
+     * @see RunNotifier#fireTestSuiteStarted(Description)
+     * @since 4.13
+     */
+    public void fireTestSuiteStarted() {
+        notifier.fireTestSuiteStarted(description);
+    }
+
+    /**
+     * Calls {@link RunNotifier#fireTestSuiteFinished(Description)}, passing the
+     * {@link Description} that was passed to the {@code EachTestNotifier} constructor.
+     * This should be called when a test suite has finished, whether the test suite succeeds
+     * or fails.
+     * @see RunNotifier#fireTestSuiteFinished(Description)
+     * @since 4.13
+     */
+    public void fireTestSuiteFinished() {
+        notifier.fireTestSuiteFinished(description);
+    }
 }
--- a/src/main/java/org/junit/runner/notification/RunListener.java
+++ b/src/main/java/org/junit/runner/notification/RunListener.java
@@ -70,6 +70,25 @@ public class RunListener {
     }
 
     /**
+     * Called when a test suite is about to be started.
+     *
+     * @param description the description of the test suite that is about to be run
+     *                    (generally a class name)
+     * @since 4.13
+     */
+    public void testSuiteStarted(Description description) throws Exception {
+    }
+
+    /**
+     * Called when a test suite has finished, whether the test suite succeeds or fails.
+     *
+     * @param description the description of the test suite that just ran
+     * @since 4.13
+     */
+    public void testSuiteFinished(Description description) throws Exception {
+    }
+
+    /**
      * Called when an atomic test is about to be started.
      *
      * @param description the description of the test that is about to be run
--- a/src/main/java/org/junit/runner/notification/SynchronizedRunListener.java
+++ b/src/main/java/org/junit/runner/notification/SynchronizedRunListener.java
@@ -43,6 +43,37 @@ final class SynchronizedRunListener extends RunListener {
         }
     }
 
+    /**
+     * {@inheritDoc}
+     * <p/>
+     * Synchronized decorator for {@link RunListener#testSuiteStarted(Description)}.
+     * @param description the description of the test suite that is about to be run
+     *                    (generally a class name).
+     * @throws Exception if any occurs.
+     * @since 4.13
+     */
+    @Override
+    public void testSuiteStarted(Description description) throws Exception {
+        synchronized (monitor) {
+            listener.testSuiteStarted(description);
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     * <p/>
+     * Synchronized decorator for {@link RunListener#testSuiteFinished(Description)}.
+     * @param description the description of the test suite that just ran.
+     * @throws Exception
+     * @since 4.13
+     */
+    @Override
+    public void testSuiteFinished(Description description) throws Exception {
+        synchronized (monitor) {
+            listener.testSuiteFinished(description);
+        }
+    }
+
     @Override
     public void testStarted(Description description) throws Exception {
         synchronized (monitor) {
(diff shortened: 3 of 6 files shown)


src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -358,6 +358,7 @@ public Description getDescription() {
     public void run(final RunNotifier notifier) {
         EachTestNotifier testNotifier = new EachTestNotifier(notifier,
                 getDescription());
+        testNotifier.fireTestSuiteStarted();
```
kcooney: Note that `ParentRunner` is the base class of `Parameterized`. I'm not sure if people would find the events sent from `Parameterized` confusing

baev: Ok, I see
```
suite started my.company.MyParameterizedTest
suite started [0]
test started
test
test finished
suite finished
suite started [1]
test started
test
test finished
suite finished
suite finished
```

Any ideas how to fix this?

baev: I guess it should looks like 
```
suite started my.company.MyParameterizedTest
test started
test
test finished
test started
test
test finished
suite finished
```

baev: if junit creates suite for each parameter maybe it's ok?

> Ok, I see
> 
> suite started my.company.MyParameterizedTest
> suite started [0]
> test started
> test
> test finished
> suite finished
> suite started [1]
> test started
> test
> test finished
> suite finished
> suite finished

baev:
link #444

src/main/java/org/junit/runners/ParentRunner.java
kcooney: It might be easier to determine if this is a problem (and if so, the best way to address it) if you could describe the problem you are trying to solve with this change and/or how you plan to use the API

baev: I am a developer of [Allure Framework](http://allure.qatools.ru/). We are using `RunListener` to collect information about tests. This change can help me to get more information about test suites - as example get the right start/stop times for it.  

BTW the problem with parameterized runner above -is not a problem for me, but I would like to choose the right way

kcooney: Well, since we can in theory report errors for the nested suites in `Parameterized` this is okay with me. I want to wait for the other maintainer to see if they have concerns.

I am curious to find the other maintainers plan on using `RunListener` for JUnit Lambda, and if so whether this change would work.

kcooney:
@junit-team/junit-committers any objections to this pull?

src/main/java/org/junit/internal/runners/model/EachTestNotifier.java
```diff
@@ -45,4 +45,12 @@ public void fireTestStarted() {
     public void fireTestIgnored() {
         notifier.fireTestIgnored(description);
     }
+
+    public void fireTestSuiteStarted() {
```
kcooney: Could you please add Javadoc that includes `@since 4.13` to all new methods?

baev: > Could you please add Javadoc that includes @since 4.13 to all new methods?

@kcooney done

src/main/java/org/junit/internal/runners/model/EachTestNotifier.java
```diff
@@ -45,4 +45,22 @@ public void fireTestStarted() {
     public void fireTestIgnored() {
         notifier.fireTestIgnored(description);
     }
+
+    /**
+     * Notify {@link #notifier} that a test suite is about to start.
```
kcooney: It seems odd to use `@link` to link to a private field. How about this;

"Calls {@link RunNotifier#fireTestSuiteStarted(Description)}, passing the {@link Description} that was passed to the {@code EachTestNotifier} constructor. This
should be called when a test suite is about to be started."

src/main/java/org/junit/runner/notification/RunListener.java
```diff
@@ -70,6 +70,25 @@ public void testRunFinished(Result result) throws Exception {
     }
 
     /**
+     * Called when test suite is about to be started.
```
kcooney: typo: "Called when _a_ test suite is about to be started."

src/main/java/org/junit/runner/notification/RunListener.java
```diff
@@ -70,6 +70,25 @@ public void testRunFinished(Result result) throws Exception {
...
+     *
+     * @param description the description of the test suite that is about to be run
+     *                    (generally a class name)
+     * @since 4.13
+     */
+    public void testSuiteStarted(Description description) throws Exception {
+    }
+
+    /**
+     * Called when test suite has finished, whether the test suite succeeds or fails.
```
kcooney: "Called when _a_ test suite has finished"

src/main/java/org/junit/runner/notification/SynchronizedRunListener.java
```diff
@@ -43,6 +43,37 @@ public void testRunFinished(Result result) throws Exception {
...
     }
 
+    /**
+     * Synchronized decorator for {@link RunListener#testSuiteStarted(Description)}.
+     * @param description the description of the test suite that is about to be run
+     *                    (generally a class name)
+     * @throws Exception if any occurs.
+     * @since 4.13
+     *
+     * @see RunListener#testSuiteStarted(Description)
```
kcooney: Instead of using `@see` use `{@inheritDoc}`. See http://docs.oracle.com/javase/1.5.0/docs/tooldocs/windows/javadoc.html#{@inheritDoc} and http://tornorbye.blogspot.com/2005/02/inheriting-javadoc-comments.html

kcooney:
LGTM other than very minor issues with the Javadoc (which I can fix when I merge if you prefer). Sorry for the delay.

Just need one of the other maintainers to add their LGTM and I'll merge this.

baev:
@kcooney done

dsaff:
This is a good thing in principle, and the implementation looks sound.

I'd like to add some very clear documentation, however.  Not every custom runner that reports a parent node and leaf nodes from getDescription() inherits from ParentRunner, so RunListeners must gracefully handle the situation where a suite node shows up in getDescription, but these new methods don't end up getting called.

Likewise, we should probably add to the documentation of RunNotifier that the new methods are strongly encouraged and helpful, but not required; otherwise we risk essentially "breaking" existing Runners.

kcooney:
@dsaff if the only concerns are the Javadoc, it might be easier to use the `hub` tool to get this branch into your local repository, update the Javadoc, squash the caches and push

dsaff:
@kcooney, I'll try to get some time to do that.  That said, @baev, if you have time, you may well finish before I do.

kcooney:
@dsaff It's been a month now. I would like to merge this (after applying the minor fixes). How about you improve the Javadoc after this is merged?

dsaff:
Fair enough.

marcphilipp:
@kcooney @dsaff Does one of you guys have enough time to work on this?

kcooney:
@marcphilipp I can take care of it this weekend

kcooney:
I created #1311 to try to resolve David's concerns about the Javadoc.

kcooney:
Merged (finallty!)

@baev would you please update the release notes at https://github.com/junit-team/junit4/wiki/4.13-Release-Notes ?

baev:
> would you please update the release notes at https://github.com/junit-team/junit4/wiki/4.13-Release-Notes ?

Done https://github.com/junit-team/junit4/wiki/4.13-Release-Notes#run-listener

marcphilipp:
Thanks!

PR #1120, opened 2015-04-08

baev:
ref #923


marcphilipp:
Please target the junit5 branch which is already on JDK 6.

baev:
@marcphilipp done

PR #1122, opened 2015-04-08

baev:


src/main/java/org/junit/runner/notification/RunListener.java
```diff
@@ -122,6 +123,15 @@ public void testAssumptionFailure(Failure failure) {
...
 
+    /**
+     * Called when some services loaded using {@link org.junit.internal.ServiceLoaderWrapper}
+     * @param serviceType a type of the loaded services
+     * @param services loaded services
+     *
+     * @since 4.12
+     */
+    public <T> void servicesLoaded(Class<T> serviceType, List<T> services) {
+    }
```
marcphilipp: @baev Why do we need this new method? It's certainly useful for testing but do we really need to add it to `RunListener`?

marcphilipp:
@baev Thanks for your pull request! Sorry it took me so long to respond. I have one concern: https://github.com/junit-team/junit/pull/1122#discussion_r30732906

src/main/java/org/junit/runner/notification/RunListener.java
kcooney: @marcphilipp I believe David or I asked for this method, so that a RunListener could print a message indicating that a service has loaded. Otherwise, you get "magic" new behavior with no notification.

marcphilipp: It feels a bit strange to me to have RunListeners notified of new RunListeners. What do you think?

kcooney: @marcphilipp the RunListeners are being notified of new services being loaded. In the future, we could allow services to do things besides add RunListeners.

src/main/java/org/junit/internal/ServiceLoaderWrapper.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ServiceConfigurationError;
+import java.util.ServiceLoader;
+
+/**
+ * Load services by specified service type using Java SPI.
+ *
+ * @since 4.12
```
marcphilipp: 5.0

src/main/java/org/junit/runner/notification/RunListener.java
```diff
@@ -122,6 +123,15 @@ public void testAssumptionFailure(Failure failure) {
     public void testIgnored(Description description) throws Exception {
     }
 
+    /**
+     * Called when some services loaded using {@link org.junit.internal.ServiceLoaderWrapper}
+     * @param serviceType a type of the loaded services
+     * @param services loaded services
+     *
+     * @since 4.12
```
marcphilipp: 5.0

src/main/java/org/junit/runner/notification/RunNotifier.java
```diff
@@ -43,6 +70,22 @@ public void removeListener(RunListener listener) {
     }
 
     /**
+     * Internal use only.
+     * <p/>
+     * Invoke to load all {@link RunListener}} from classpath using Java SPI.
+     * @see org.junit.internal.ServiceLoaderWrapper
+     *
+     * @since 4.12
```
marcphilipp: 5.0

src/main/java/org/junit/runner/notification/SynchronizedRunListener.java
```diff
@@ -15,9 +17,8 @@
  *
  * @author Tibor Digana (tibor17)
  * @author Kevin Cooney (kcooney)
- * @since 4.12
- *
  * @see RunNotifier
+ * @since 4.12
```
marcphilipp: 5.0

src/main/java/org/junit/runner/notification/RunNotifier.java
```diff
@@ -193,6 +237,21 @@ protected void notifyListener(RunListener each) throws Exception {
     }
 
     /**
+     * Invoke to tell listeners that a some services loaded using {@link org.junit.internal.ServiceLoaderWrapper}
+     * @param serviceType a type of loaded services
+     * @param services loaded services
+     * @since 4.12
```
marcphilipp: 5.0

src/main/java/org/junit/runner/notification/RunListener.java
marcphilipp: Do you think a separate listener `ServiceLoaderListener` would be feasible?

src/main/java/org/junit/internal/ServiceLoaderWrapper.java
kcooney: Actually, the main branch is 4.13. JUnit 5 changes are on the java5 branch. We may decide not to release 4.13 and jump to 5.0, but for the time being we're having people say `@since 4.13` on main

src/main/java/org/junit/runner/notification/RunListener.java
kcooney: @marcphilipp how would one register a listener of type `ServiceLoaderListener`?

Is there a specific concern about adding methods to `RunListener`?

marcphilipp: Through JUnitCore?

No general concern from my side. However, looking at the existing methods in `RunListener` adding a method called `servicesLoaded` seems in cohesive to me.

PR #1124, opened 2015-04-13
Provide all annotations except RunWith. Fixes #751

### LINKED ISSUE #751: Parameterized runner hides Category annotation
I have noticed that when we changed a few tests of ours to use the Parameterized runner, the @Category annotation stops working; ie our performance test suite skips tests like the one attached below.

We are using gradle and junit 4.11 in the project.

I tried to rectify the problem by removing the overridden method getRunnerAnnotations() in Parameterized.TestClassRunnerForParameters and it solves my problem. I'm not sure what other problems this would introduce, though...

Best regards 
/ Per

```
import org.junit.Test;
import org.junit.experimental.categories.Category;
import org.junit.runner.RunWith;
import org.junit.runners.Parameterized;

import static java.util.Collections.singletonList;

@Category(PerformanceTest.class)
@RunWith(Parameterized.class)
public class TheMissingPerformanceTest
{
    @Parameterized.Parameters
    public static Iterable<Object[]> data()
    {
        return singletonList(new Object[0]);
    }

    @Test
    public void testSomething() throws Exception
    {
    }
}
```

stefanbirkner:
Couldn't see any reason why we should remove annotations like
`@Category` from the child runners of parameterized tests. Now the
`Categories` runner can handle parameterized tests.

### DIFF
--- a/src/main/java/org/junit/experimental/categories/Categories.java
+++ b/src/main/java/org/junit/experimental/categories/Categories.java
@@ -315,7 +315,6 @@ public class Categories extends Suite {
         } catch (NoTestsRemainException e) {
             throw new InitializationError(e);
         }
-        assertNoCategorizedDescendentsOfUncategorizeableParents(getDescription());
     }
 
     private static Set<Class<?>> getIncludedCategory(Class<?> klass) {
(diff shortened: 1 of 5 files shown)


marcphilipp:
@dsaff Can you please take a look?

dsaff:
Looks alright.  I'm still a bit bugged about why we didn't do it this way in the first place, but just squinting my eyes harder isn't producing any more memories, so let's go ahead.

marcphilipp:
I think this pull is good to be merged, isn't it?

dsaff:
Done.

PR #1128, opened 2015-04-19
Update Test.java

Jonatino:
Fixed a grammatically incorrect comment.


marcphilipp:
It's not a sentence, so I don't think we need a period here, do we?

kcooney:
Having a period isn't required for parameter docs, but is generally used for method and class descriptions:

http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#format

marcphilipp:
Thanks for looking that up!

PR #1129, opened 2015-04-21
Update TestFailure.java

SnailKnows:


baev:
I guess this PR has no effect. `PrintWriter#close()` only close his writer (In this case it is `StringWriter`). From `StringWriter#close()` docs: 

*Closing a <tt>StringWriter</tt> has no effect. The methods in this class can be called after the stream has been closed without generating an <tt>IOException</tt>.*

kcooney:
Agreed. In fact, flushing a `StreamWriter` has no effect, either

@SnailKnows did your IDE show a warning here?

SnailKnows:
Thanks.I understand after reading the source of StringWriter.

SnailKnows:
@kcooney  Not warning.

PR #1130, opened 2015-04-25
Add Ordering, Orderable and @OrderWith

kcooney:
These APIs allow arbitrary ordering of tests, including randomization.


dsaff:
Woah, nice!

Beginning to dig in.  Thanks for the work here.

src/main/java/org/junit/runner/manipulation/GenericOrdering.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.List;
+
+import org.junit.runner.Description;
+
+/**
+ * An {@link Ordering} that is not a {@link Sorter}.
+ *
+ * @since 4.13
+ */
+public final class GenericOrdering extends Ordering {
```
dsaff: Name and comment here is a bit confusing.  I might call it DelegatingOrdering?

src/main/java/org/junit/internal/runners/JUnit38ClassRunner.java
```diff
@@ -170,6 +174,13 @@ public void sort(Sorter sorter) {
         }
     }
 
+    public void order(GenericOrdering ordering) throws InvalidOrderingException {
```
dsaff: Why can't this take regular Ordering?

src/main/java/org/junit/runner/OrderWith.java
```diff
@@ -0,0 +1,26 @@
...
+ * When a test class is annotated with <code>&#064;OrderWith</code> or extends a class annotated
+ * with <code>&#064;OrderWith</code>, JUnit will order the tests in the test class (and child
+ * test classes, if any) using the ordering defined by the {@link Ordering} class.
+ *
+ * @since 4.13
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Inherited
+public @interface OrderWith {
```
dsaff: A couple musings here (which I may not finish all at once)

1) I'd be strongly tempted to make this an annotated static field instead/in-addition.  Why?  Well, consider the case (which I've written several times in the past) where you try to run the tests from the fastest to the slowest, based on a record of previous runtimes kept in a file.  It probably makes sense to allow the location of the file to be specified by the test writer.  I think that with the current pull request, this would be 100% impossible.  One COULD imagine a different spec of this annotation where what's returned is not an Ordering, but a functional interface that takes a Description and returns an Ordering, so that one could write a Fastest Ordering that looks for a @TestTimesFile annotation on the Description to know what file to use, BUT...

It might be so much easier to instead just say:

@OrderWith public static Ordering ordering = new FastestOrdering("/tmp/test-times.txt");

or some such.

src/main/java/org/junit/runner/manipulation/GenericOrdering.java
kcooney: @dsaff I would rather not call it `DelegatingOrdering`. It is specifically an `Ordering` that is known not to be a `Sorter`, and some of the core code depends on that. I'm open to other names and improved doc, of course.

src/main/java/org/junit/internal/runners/JUnit38ClassRunner.java
kcooney: Because `Sorter` extends `Ordering`, but we want to make sure that anyone attempting to do a sort of a runner goes through `Ordering.apply(Object)`. Originally I didn't have  `Sorter` extend `Ordering`, and there the code was more complicated and had more duplication.

src/main/java/org/junit/runner/OrderWith.java
kcooney: @dsaff I don't see how what you are proposing would be 100% impossible. You could write your own ordering that reads the file the first time `apply()` is called. If the location of the file should be specified by the test writer, then your `Ordering` could use a custom annotation.

I'd hate to add more broiler-plate because of a possibly rare use case. Although I've personally never been bothered by adding a method with an annotation that returns something, other people have objected loudly to this (for example, with Rules).

The problem with passing a `Description` when you create an ordering is that an ordering applies to all children. Luckily, you can get a `Description` from a `Runner` so a custom `Ordering` do a different algorithm based on the `Description` inside of `Ordering.apply()`. I could be convinced that `Ordering.order()` should also take in the parent `Description`.

dsaff: 2) Changes like this (which affect how _any_ class processed by JUnit is run, regardless of the chosen runner) have been rare, on purpose.  There's a good argument to be made in favor of this one, but as long as we've got the hood open, I wonder if something more general would make sense.

For example, there's not currently a FilterWith annotation, but once we accept this change, that seems to be a natural next square in the matrix to fill in, which would be _another_ top-level change.  Also related would be the common request to attach a custom runlistener to a class, ListenWith

Just thinking out loud, what if we made a top-level annotation (RunRule?  Modify?) which would handle any of these changes?  Essentially:

public interface RunRule {
  public Runner apply(Runner runner, Description classBeingRun);
}

and then:

public class MyTest {
  @UseRunRule public static RunRule ordering = new UseOrdering(Ordering.random());
  @UseRunRule public static RunRule listening = new AddListener(new MyListener());
  // etc...
}

Note that: 

a) This builds on the static field suggestion in (1), but could be orthogonal to it.
b) I'm not thrilled with the top-of-my-head naming here; especially, RunRule makes this sound tied to Rule, for better or for worse.

src/main/java/org/junit/internal/runners/JUnit38ClassRunner.java
dsaff: Why the duplication?  I can see that one would want to be able to do both @OrderWith(MyOrdering.class) and @OrderWith(MySorter.class), and that's difficult unless one of the classes extends the other.  But it looks as though Ordering is essentially a marker interface at this point: most of the code that actually does anything is assuming that all Orderings are either Sorters or GenericOrderings.  If you _like_ the RunRule suggestion below, this becomes more of an asset than a liability, but if not, maybe there's some further shuffling that could happen here.

src/main/java/org/junit/runner/OrderWith.java
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * @return a class that extends {@link Ordering} (must have a public no-argument constructor)
```
dsaff: [Sorry, deleted and re-instated the comment on a different line once I realized what a nightmare having two discussions on the same line of code would be.  Alternatively, we could pop the whole conversation out to the general comment thread]

2) Changes like this (which affect how any class processed by JUnit is run, regardless of the chosen runner) have been rare, on purpose. There's a good argument to be made in favor of this one, but as long as we've got the hood open, I wonder if something more general would make sense.

For example, there's not currently a FilterWith annotation, but once we accept this change, that seems to be a natural next square in the matrix to fill in, which would be another top-level change. Also related would be the common request to attach a custom runlistener to a class, ListenWith

Just thinking out loud, what if we made a top-level annotation (RunRule? Modify?) which would handle any of these changes? Essentially:

public interface RunRule {
public Runner apply(Runner runner, Description classBeingRun);
}

and then:

public class MyTest {
@UseRunRule public static RunRule ordering = new UseOrdering(Ordering.random());
@UseRunRule public static RunRule listening = new AddListener(new MyListener());
// etc...
}

Note that:

a) This builds on the static field suggestion in (1), but could be orthogonal to it.
b) I'm not thrilled with the top-of-my-head naming here; especially, RunRule makes this sound tied to Rule, for better or for worse.

src/main/java/org/junit/runner/OrderWith.java
dsaff: So my apply() method would cast the Object to a Runner, call getDescription, and then look for the custom annotation in the Description?  On balance, I'm not a fan of asking end-extenders to cast things, but could see doing it if there are benefits gained.

I can't personally call this use case "possibly rare", since it's occurred in the majority of test-case sorting scenarios I've written (and since test re-ordering was an essential part of my thesis, I may be in the sorry position of having personally seen a healthy percentage of all test-ordering schemes our species have attempted...)

src/main/java/org/junit/internal/runners/JUnit38ClassRunner.java
dsaff: (Not editing earlier comment in case you're currently responding, but "Why the duplication?" is too strong.  More to the point: "I see the case of the target type of OrderWith.  Are there other places?")

dsaff:
Another thing potentially worth considering sooner-rather-than-later: I imagine that the request will come down the road for a --sort_with annotation to complement the existing --filter_with.  If a class has OrderWith, and the command line has --order_with, who wins?  (I assume we just say "command line" and move on, but if you have another idea?)

dsaff:
Holding off on digging for further feedback while top-level issues settle down.  Thanks again!

src/main/java/org/junit/internal/runners/JUnit38ClassRunner.java
kcooney: Perhaps we should chat over VC, but I'll try to answer

 If `Sorter` doesn't extend `Ordering` then we have to write extra code to allow creation of an `Ordering` from a `Sorter` or `Comparator` (since `Ordering` is a very generic concept, it seems reasonable that one could have an `Ordering` that performs a sort; to do otherwise would require people to choose an API based on the current implementation). 

Also, as I said earlier, if we allow `order()` to take in an `Ordering` and that ordering does a sort, then we'll get worse performance than we would if the caller was smart enough to "know" that they should instead call `sort()` (really they shouldn't call either, and instead apply the `Ordering` or `Sorter` to the `Runner`).

Note that I also made `Orderable` extend `Sortable`. We could change that if we think it's an extra burden to people that write `Orderable` runners to have to also make them `Sortable`. If you want to get a hint as to the changes I would need to make if  `Orderable` did not extend `Sortable`, look at the second commit in this pull (the "unreachable code" would be reachable).

src/main/java/org/junit/runner/OrderWith.java
kcooney: @dsaff would you be fine with allowing a static method annotated with `OrderWith` as an additional way of specifying an ordering, and document whether the static method overrides the annotation?

src/main/java/org/junit/runner/OrderWith.java
kcooney: I'm not sure why we would want end users to have to understand that filtering and sorting are applied via this new rule mechanism. They are orthogonal, and I think we should provide the simplest possible API to enable them.

I don't see why one couldn't implement `@FilterWith` today, without any changes outside of `ParentRunner`. Anyone extending `ParentRunner` would get that "for free", and filtering would work for children in a suite as long as the children implement `Filterable`. It sounds like what you are proposing would require a new general purpose mix-in interface, and users would have the confusing problem of having a `Runner` that's `Filterable` but can't be filtered via a `RunRule`

What you are proposing sounds like a larger redesign of runners, and one we should consider as part of the runner refactoring being proposed by `JUnitLambda`.

kcooney:
Do we thing that we have many users that use the JUnit command line flags to run tests where it makes sense to add `--sort_with`? I suspect most users run tests via an IDE, Ant, Maven or Gradle. Each of those would have their own way of adding a sort, and can do so via `Request.orderWith()`. If someone wanted us to support this in the command line, it wouldn't be much work, but I'd rather not do it in the same pull as the pull will get too large.

src/main/java/org/junit/runner/OrderWith.java
kcooney: @dsaff alternatively, as I suggested, we could add the parent `Description` to the parameters passed to `Ordering.order()`, which should handle your use case.

kcooney:
I do appreciate the feedback, and I'm open to looking for a more general concept there. I'm just pushing back because `@OrderWith(MyOrdering.class)` is extremely simple and self-documenting, while some of the APIs you've proposed are less simple. Many people have asked us for a way to specify rules via an annotation, because having a field was "too much code" or because "fields imply state", and while I don't always agree with them, it's happened enough times that I'm trying to think of the simplest API to express the concept of "order the tests here based on this conceptual ordering".

dsaff:
A couple non-exhaustive thoughts:

1) I'm throwing out a bunch of alternatives not necessarily because I consider them _obviously_ better, but because I think we'll be better off to have explicitly considered and rejected (or accepted) them; hopefully that comes across.

2) I had missed that this is plugged in through ParentRunner, not through the top-level RunnerBuilders.
a) On the one hand, that reduces the scope of the change, and also makes it more likely that systems that use custom RunnerBuilders (android test system, for example) will also work with this change.
b) On the other, it means that whether putting @OrderWith on a class will actually apply the Ordering depends on whether the class is a subclass of ParentRunner; at the very least, we should probably add documentation explicitly to BlockJUnit4Runner, JUnit3ClassRunner, and Suite to say that they "observe @OrderWith", with a link explaining what that means, and add documentation to OrderWith saying that you can only depend on its documented behavior if it is attached to a class that promises to "observe "@OrderWith".

3) One thing I need to let simmer for a while is the effect of this change on people just learning xTest-style testing.  JUnit has always been opinionated about what it makes easy and what it makes possible.

Put another way, I'd hate to give credence to the narrative "JUnit used to irrationally make it hard to order your tests so that they worked, but with 4.13, they added OrderWith, so now you can go ahead and write order-dependent tests, no more problems."

(There's no action item on #3, just a warning that I may need to let this process a day or two before I'm ready to pull the trigger.)

kcooney:
For the third point, although I agree that writing tests that depend on a specific ordering is a bad idea, we've been asked about this a number of times (as well as supporting shuffling tests) that we should consider some way to allow this, without baking into JUnit the particular strategy.

If you want to explore the `RunRule` route, could you create a pull request with an outline of what you had in mind?

When I have some time, I'll explore splitting `Ordering` and `Sorter` so you can see how that changes things.

I wonder if we could attach `@OrderWith` at the `RunnerBuilder` level.

One other concern: the way `@OrderWith` is implemented bothers me a bit, because the ordering is done the first time someone asks for the children, and that can happen as a result of a number of different actions. In particular, if a suite that uses `@OrderWith` includes a test class annotated with `@OrderWith`, it's unclear which will be applied first. I don't know if there's another hook that can be used to do ordering at a deterministic time. Perhaps doing the work in `RunnerBuilder` would solve that

kcooney:
@dsaff I added two commits.

The first one moves the handling of `@RunWith` to `RunnerBuilder`.

The second changes `Sorter` to no longer extend `Ordering`. I added this so you get an idea of the complexities that I had before I decided that `Sorter` should extend `Ordering`. Note that with these changes, you can't say `@OrderWith(MySort.class)`

dsaff:
(I think you meant "moves the handling of @OrderWith to RunnerBuilder").

Now that I see it, I think that RunnerBuilder is one step too far up to move this: now it's possible to write a suite that doesn't respect @RunWith on its children (by replacing AllDefaultPossibilitiesBuilder), but _not_ possible to not respect @OrderWith.  I'm still not sure about making @OrderWith be on the same level of the call tree as @RunWith, but I'm pretty sure it doesn't make sense to put it above.  What do you think?

Re: Sorter doesn't extend Ordering: Note that now we have Sorter with a void apply(Object) method, and Ordering with a void apply(Object) method.  One could create a [choosing random word] FooBox interface that has just this apply method, and cause OrderWith to accept any subclass of FooBox, thereby (I think) avoiding the need to wrap in many cases.

However, at that point, we're awfully close (again) to RunRule, since FooBox basically can do anything it wants to the Runner it's passed, not just ordering: it could certainly filter, and we might even be able to finagle it to inject listeners or rules.  So we'd either have to decide to arbitrarily hamstring OrderWith, or accept that very clever people will use OrderWith to do non-ordering things, or give it a different name at that point.

I'm again tempted to go the fully generic route, since (a) we're almost there already, and (b) given the rather big (in percentage terms) addition here to top-level API, I'd feel much better about it if the features unlocked including more than ordering, which (as I've mentioned above) I'm worried may prove a mixed blessing.

Don't feel a need to go change the code just yet; I'd love to hear your thoughts on the above (unless changing the code is the easiest way to do that).  Thanks again!

And I can work on a fork that shows what RunRule might look like, although today isn't necessarily looking conducive.

kcooney:
@dsaff I actually think that `RunnerBuilder` is the right way to handle `@OrderWith`. A runner indicates that it can be sorted via `Sortable`. `@OrderWith` is a simple way to order any runner that is `Sortable` or `Orderable` without going trough `Request`.

I don't see any problems with third-party subclasses of `AllDefaultPossibilitiesBuilder` not being able to turn off the `@OrderWith` behavior. Could you give an example of a problem? And is that problem worse than the complexity of the Javadoc we would need on `@OrderWith` to describe when it would be ignored?

I'm sorry, but I still don't quite follow how your `RunRule` proposal would work. Can you whip up a quick implementation and point us to a branch on your github repo that has the changes? It's much easier to talk about actual code.

src/main/java/org/junit/runner/manipulation/SortBasedOrdering.java
```diff
@@ -0,0 +1,38 @@
...
+class SortBasedOrdering extends Ordering {
+    private final Sorter sorter;
+
+    public SortBasedOrdering(Sorter sorter) {
+        this.sorter = sorter;
+    }
+
+    @Override
+    public void apply(Object runner)  {
+        if (runner instanceof Sortable) {
```
kcooney: This was some of the ugly code that caused me to want to make `Sorter` an `Ordering`.

dsaff:
We currently allow Suites to ignore RunWith on their children, for example:

http://www.patterntesting.com/release/apidocs/src-html/patterntesting/runtime/junit/internal/SmokeBuilder.html

This facility may be under-documented, but it already exists, and is used.

As one extension that I think we'd want to allow, imagine a BackAndForthSuite runner that first runs its children in their native order, and then runs them again in the precise opposite direction, in order to rout out any order-dependent tests.  I'll need to play with the code to see whether such a thing would be possible with the current implementation.

kcooney:
@dsaff if you wanted to disable `@OrderWith` you could do something similar. You would wrap the runner with another runner that didn't implement `Sortable` or `Orderable`

kcooney:
The current implementatiion makes sure the reordered children includes all original children exactly once. I could allow running tests multiple times, but I am not sure that we should.

marcphilipp:
Given our plans for JUnit Lambda and JUnit 5, I don't think we should introduce big new features like this one. Have you considered that?

kcooney:
@marcphilipp  I thought lambda-based tests were not going to use the Runner classes? Wasn't the plan to create a different extension mechanism that gave us more freedom?

It would be good to be on the same page as to the kinds of changes we would postpone due to JUnit Lambda, and how long we were willing to put those features on hold if JUnit Lambda is delayed.

kcooney:
I added two more commits.

The first changes `Sorter` to extend `Ordering`, because I think that provides the cleanest API for people using `Ordering` and `@OrderWith`.

The second changes the signature of `Ordering` to take in the parent, so the `Ordering` could look at the parent annotations to help drive the decision. I think this should address David's concern about an ordering that wants to use some kind of external data source (to run fast and non-flaky tests first, for example). An ordering that needs more data can allow/require a custom annotation.

dsaff:
OK, made a commit demonstrating what I had called RunRule above.  In this commit, I've called it ModifyWith:

https://github.com/dsaff/junit/commit/280a07658353cc96af7edb5708e00bbda95bc1b2

Just to be clear, this commit is not intended to be a strong suggestion of an endpoint, but a demonstration of one idea with as few changed lines as possible.  The tests show how this allows either orderings or filters to be applied.  I had it in mind to also use this to make a general Injectable strategy, to demonstrate the flexibility but (1) I didn't want to spend too much time on it and (2) I was worried about derailing conversation onto issues about injection.

Orthogonal issues NOT intended to be addressed in this commit:

1) Naming (Modifier is dangerously bland and uninformative)
2) Class-level annotations vs. annotation static fields
3) Error handling (the duplicate code around ErrorReportingRunner is probably a smell that could be eliminated)

Does that help show where I was considering going?

dsaff:
(Also about the above commit, I'm not sure whether allowing modify to return a different Runner is useful;  we might remove that quirk.)

kcooney:
@dsaff interesting I am curious if the tests pass (especially with the filter changes).

It seems inconsistent to have this functionality available via an annotation bit not via `Request`. Was that an intentional decision, or just not something you had time to add to your proof-of-concept.

I'll try to find time to play around with this.

dsaff:
@kcooney, yes the tests pass.

Good point; I assume we'd add this to Request as well, just hadn't thought about it.

kcooney:
I few more thoughts

1. I actually like the name `Modifier` :-)
2. I would want `Ordering.modify()` to throw an exception if the runner was not `Orderable` (rather than silently doing nothing). Not sure if I feel the same way about `Filter.modify()`
3. I'll look again to see if, with this change, it still makes sense for `Sorter` to extend `Ordering`
4. I think returning a `Runner` would encourage people to reuse runners in ways that can be fragile and would therefore limit us in the future. Long-term, I would rather create extension points that remove the need to write custom runners (for example, an extension point that acted like a class rule and a method rule) and returning a `Runner` moves us in the wrong direction.

dsaff:
1) OK.  :-)
2) Hmm.  I could certainly imagine a use case where I'd want to say "try to run these tests in alphabetical order, but if for some of the subtests I can't (because they're run by a custom non-Orderable Runner), I'd rather run them in native order than not at all".  Would that be possible in your scheme?  Actually, it seems possible to me that best-effort ordering might even be the common case.  Do you not think so?
3) OK, looking forward to results.
4) The flip side of _not_ returning a Runner is that it means that all modification has to involve mutation, which is sometimes problematic.  But as I mentioned above, I think I can be convinced.

kcooney:
2) yes, if the suite was Orderable but it contains child Runners that are not Orderable, that should be fine. I just don't want to apply the annotation and have it do nothing at all
4) can you think of a specific example of where not returning a Runner would be problematic?

dsaff:
2) Ah, got it.
4) The biggest one that comes to mind (which I should have just coded up, and still may) is a Modifier that wraps the RunNotifier of the modified Runner.  This requires no knowledge of the underlying Runner, but is a request that has come up many times.  Actually, this example makes me think that maybe we _do_ want to return a (potentially new) Runner.  Let me know if code would help explain that point.

Can you give an example of how people might reuse Runners in fragile ways, to elaborate on your comment above?

kcooney:
@dsaff one problem with returning a different runner is the caller might be trying to wrap the runner but ends up doing it incorrectly. If this API was added before `Orderable` existed, for example, the Modifier would make the test no longer orderable. We could solve this by having our own RunnerWrapper class that implements `Orderable`, `Fitlerable` and `Sortable` but people might not think to use it.

Another example is my proposal for an `Initializable` mix-in (where `RunnerBuilder` calls `Initializable.initialize()`). Again, the Modifier might not know about `Initializable`, so it takes in a runner that was initialized and returns a different runner that was not initialized.

I also have some kind of vague feeling that we would have fragile tests if people used this runner wrapping mechanism to have one runner delegate some calls to another runner. Runners, for better or for worse, share code via inheritance (ex: `BlockJUnit4ClassRunner`) or via the composite pattern (ex: `ParentRunner`), not delegation.

Can you give me a concrete example of a need to wrap the `RunNotifier`? I wonder if that's a different use case.

kcooney:
New concern: `@ModifyWith` is a class annotation, so either we can only allow one annotation for a class (preventing ordering and filtering) or we can't limit how many instances we get nor can we control the ordering (causing nonsensical behavior like ordering and sorting).

What if we make this a meta-annotation? So users would still say `@OrderWith(MyOrdering.class)` and `OrderWith` itself was annotated with `@ModifierAnnotation`? Then runner builder looks for annotations that are annotated with `@ModifierAnnotation`

dsaff:
Here are two instances in which people have wanted to inject RunListeners, although the first is likely more relevant than the second:

https://groups.yahoo.com/neo/groups/junit/conversations/topics/24645;_ylc=X3oDMTM2aml1NmF2BF9TAzk3MzU5NzE0BGdycElkAzI0MjMzMjgEZ3Jwc3BJZAMxNzA1MDA2OTA1BG1zZ0lkAzI0NjQ1BHNlYwNmdHIEc2xrA3Z0cGMEc3RpbWUDMTQyNTg2NDI5OAR0cGNJZAMyNDY0NQ--

https://github.com/junit-team/junit/pull/923

That said, your point about losing mixinability is a good one, so trying to squish this into that may be a step too far.

dsaff:
Thoughts:

1) We can't prevent all nonsensical combinations if this is general-purpose, so I file that under "you asked for this much rope, your choice what to do with it"

2) I'm pretty staunchly against meta-annotations here: it can exponentially increase the number of files one must look at in order to answer the simple question "is this class doing what it looks like it's doing"?  This has impact both on humans, and machines (annotation reflection was at one point the biggest culprit of JUnit runtime overhead, until some clever and subtle fixes).  

So what about @ModifyWith({MyFilter.class, MySorter.class})?  At least it's ordered.

I'm debating whether we might want to pass a Class to the Modifier as well:

+) It might be a good idea because it would _allow_ Modifiers that want to use runtime data to do so.  For example, imagine a sorting strategy that can be tuned in many ways (how much weight to give to runtime, flakiness, historical probability of failure, etc.)  While one _could_ create an extensive Annotation-based "language" to express all of these settings, it might end up much more natural and extensible to have the configuration expressed in a static field or method.

-) On the other hand, I've tried to avoid wherever possible the assumption that every Runner is strongly attached to a particular Java class.

Musing...

kcooney:
@dsaff responses to your responses

1) many people might not know that the ordering of annotations doesn't matter.

2) I don't understand your concerns about meta annotations.

The actual annotations 99% of users of JUnit would use (like `@OrderWith`) can have clear Javadoc that explains how they can be used. To understand how things work under the covers, I think an "extender's guide to JUnit" is long overdue.

The syntax benefits of allowing use of meta annotations is huge. To go back to an example you have earlier, if I wanted the ordering to depend on the history of previous runs (say, failing tests first) and you needed a way to specify where the config file lives, `@RunFailingTestsFirst(dataFile="/tmp/runHistory")` is much easier to understand than `@ModifyWith(RunFailingTestsFirst.class) @RunFailingTestsConfig("/tmp/runHistory").

To give a second example, look at pull 1138. I'd love to allow third-parties to provide APIs like is being proposed. It would be great if the author of 1138 could create a third-party library defining an annotation named `@Paramertized` so you could do this:

    @Test
    @Paramertized(testCases={
        @TestCase({"1","2","3"}),
        @TestCase({"2","3","4"})
        })
    public void check(int a, int b, int c) {
        assertEquals(c, a+b);
    }

(note: I'm not a fan of this particular syntax for parameterized methods, but who knows, it could take off)

If we didn't allow meta-annotations, the best the author could do is this:

    @Test
    @ParamertizeWith(FancyParameters.class)
    @FancyParametersParams(testCases={
        @TestCase({"1","2","3"}),
        @TestCase({"2","3","4"})
        })
    public void check(int a, int b, int c) {
        assertEquals(c, a+b);
    }

As for the performance concerns, I don't think we should preemptively disregard possible features because we think they might not be performant. We can implement, measure, optimize and iterate.

kcooney:
@dsaff finally had time to look at the examples of when users what to wrap the Runner. I think the issue from that groups.yahoo.com thread can be solved by providing a way to add a `RunListener` via an SPI. That would be cleaner than wrapping the runner.

dsaff:
I'm worried I'm going to miss some points, but here goes:

1) I'd like to come back to the ordering question in a minute, since I think the problem is different depending on what the base language for modifiers is.

2) I agree that configuring one annotation with another is ugly and hard to figure out.  That's one reason why we went with annotated fields for Rules.  

Here's my pitch for Modifier fields, as at least one of the ways they can be specified.  (a) The drawbacks don't seem that bad to me (b) There are clear advantages.  (c) I think there are some clear disadvantages to meta-annotations.

a) The drawbacks don't seem that bad to me.

I think that annotated fields have worked for Rules.  I know that _some_ people would like a shorthand (and there's nothing preventing us from adding one, for when the shorthand works), but in general, I would say that I've been pretty happy with that decision, and I still like that as a solution here.  So then your RunFailingTests example becomes:

@ModifyWith public static Modifier failuresFirst = new RunFailuresFirst("/tmp/runHistory");

Yes, there's "public static Modifier [name]", none of which add much to the meaning, on one hand.  But Java is a boilerplate language; the more we try to find tricks to elide boilerplate, the less the result is going with the grain of Java, IMHO.

b) There are clear advantages of annotated fields:

Fields can be reused:

@ModifyWith public static Modifier OUR_TEAM_MODIFIERS = TeamConfig.OUR_TEAM_MODIFIERS.

And they can include code:

@ModifyWith public static Modifier sortByNumberOfCharactersInName = new Sorter(new Comparator() {
  public int compare(Description a, Description b) {
    return a.getName().length() - b.getName().length();
  }
}

They can provide obvious callback points for related code:

@ModifyWith public static RunFailuresFirst failuresFirst = new RunFailuresFirst("/tmp/runHistory")

... later ...

// This test is super flaky; I don't want it always popping to the top of the test queue
// TODO(saff, 1/4/2002: fix this flaky test)
failuresFirst.dontPrioritizeThisTestIfItFails();

c) There are clear disadvantages of meta-annotations, _at this level_

I have sometimes been in the situation in which someone used a test framework in an extremely clever way that I couldn't find without serious digging.  Because of this cleverness, tests failed in ways that looked like they were finding bugs in the production code, or (much worse!) tests passed that should have failed, because it wasn't clear that there was a convention that needed to be followed for the tests to have any meaning at all.

In most cases, this test-case cleverness was implemented by a software engineer that had long since left the team (sometimes for reasons not entirely unconnected with their tendency to produce over-clever solutions (yep, you guessed it, they were promoted)).

Because of this, I really really really don't want to be in a position where to truly understand what a test is doing, I need to start _also_ looking at every annotation on every annotation of every superclass of my test case, as well as any @RunWiths on those classes; adding this as a vector for invisible cleverness is, in my opinion, very much a regression.

Now, note that my primary concern here is that this is a _top level_ API.  Deep in the Theories API, for example, there are meta-annotations for selecting parameters, which look very similar indeed to the meta-annotations you're suggesting for Parameterized.  I think that's a different situation, because it is part of how one particular optional runner works, not all runners, and I'd be much more open to such a change there.

NEW 3) The point of passing the Class was for finding things through reflection that weren't in the original description.  But, similar to (1), I think I'd rather dig in on the question of simple class-level annotations vs. class-level meta-annotations vs annotated fields before going to far on this question.

kcooney:
@dsaff while I can see some of the advantages, I think the `@OrderWith(MyTeamDefaultOrdering.class)` is much more clear.

I'll go through your points

2a)

yes, Java is a broilerplate language, but we've had multiple people request for a simpler way to add rules to classes. Some of our users feel really strongly that adding fields to add favor is fundamentally broken (remember the discussions in past threads about "fields implying state")

2b)

annotations can be reused as well:

    @MyTeamModifier(behavior=Behaviors.FOR_INTEGRATION_TESTING)

yes, fields can contain code, but with my proposal, the annotation's behavior is specified via the meta-annotation, and the annotation can require a class to be specified that allows the use to augment the behavior:

    @OrderWith(MyCustomOrdering.class);
    public class MyTest {
      public static MyCustomOrdering { ...}
    }

Yes, that's a lot of code for the case where you don't need sharing, but as you said, Java is a broilerplate language. In the less common cases, I'm fine with broilerplate.

2c)

It's possible for anyone to write unmaintainable code. Even in the situation you gave, if our annotations are `@Documented` then all someone needs to do is look at the javadoc of their annotation and see that it's annotated with `@Modifier`, then go to the `@Modifier` javadoc to see what's going on.

@marcphilipp can we have a third opinion? I think these are the open issues

1. Do we generalize this from `@OrderWith` (with an Ordering) to a Modifer
2. If the answer to 1 is 'yes' do still allow an ordering to be specified with `@OrderWith`, or do we require a modifier API?
3. If the answer to 1 is 'yes', do we allow people to specify new annotations to apply that behavior via meta annotations

kcooney:
One compromise would be to update this pull to rename `Orderable` to `Modifiable` but otherwise keep it a-is. So the internal API would be generic, and you could specify a modifier via the`Request` API, but there would be no way to specify a modifier in your test class.

kcooney:
Ping

ffbit:
Once the arbitrary method ordering is introduced to JUnit, we say "Good bye!" to the original Kent Beck's idea of isolated test cases and say "Wellcome!" to the test cases coupling.
It might be that you need another unit test framework with that support, like [TestNG](http://testng.org/doc/index.html).

kcooney:
@ffbit while that is true, many _many_ people have asked for a way to specify ordering, for numerous reasons:

"I would like the simple test cases to run before the complex ones so people can see which tests to fix first"

"I want to shuffle the test order to ensure my tests are actually isolated"

"I want to write non-isolated tests"

I don't think it is unreasonable to create APIs that would support the first two use cases.

cardil:
:+1: for this

pivotaldish:
+1, we **REALLY** need a way to specify ordering (random, esp)

cardil:
I'm wondering is there a way to have a cookie and eat it at same time. If we can figure out a way to validate provided Ordering class to produce random execution order, test cases coupling could not be possible. :thinking:

kcooney:
@cardil as I said previously, many of our users want test case coupling.

"I would like the simple test cases to run before the complex ones so people can see which tests to fix first"


"I want to write non-isolated tests"

kcooney:
@@marcphilipp could you please comment on this?

marcphilipp:
> could you please comment on this?

All I can say is that we have an issue in the junit5 backlog that deals with ordering but didn't get to it, yet: junit-team/junit5#48.

kcooney:
Thanks, @marcphilipp 

I didn't see anyone on that thread saying we shouldn't provide a way for tests to be ordered, which I think is a good sign.

Most of the discussion was related to how the test authors would specify ordering, not on what the extension API would look like (though at least two people seemed to state that there should be a way for an extension to do this). Since my pull is about adding the API, hopefully we can bypass most of the debate on that read.

What's our path forward for making largish changes to JUnit 4? How do I find reviewers, and how do we decide whether to add the features at all?

src/main/java/org/junit/runner/OrderWith.java
kcooney: Sorry for the long delay, but I hope that we can release JUnit 4.13 in 2017 and I'd like to include this pull.

I updated the code to pass a context object that allows you to access the top-level description
of the class annotated with `@OrderWith`. This will handle David's use-case of " run the tests from the fastest to the slowest, based on a record of previous runtimes kept in a file". The filename could
either be specified as an annotation on the class annotated with `@OrderWith` or it could be
derived from the class name of the class annotated with `@OrderWith`

kcooney:
@marcphilipp I rebased this branch and I believe I resolved David's concern. Could you please give this a full review? Users keep asking for sorting/ordering, so I really want this in JUnit 4.13

marcphilipp:
@dsaff Can you take another look?

kcooney:
@marcphilipp I don't think David is keeping up with JUnit these days.

Could you take a look? Being able to shuffle tests has been requested multiple times.

marcphilipp:
Sure, I'll give it a thorough review in the next few days.

src/main/java/org/junit/runner/manipulation/Ordering.java
```diff
@@ -0,0 +1,158 @@
...
+ * order or even shuffle the order.
+ * 
+ * <p>In general you will not need to use a <code>Ordering</code> directly.
+ * Instead, use {@link org.junit.runner.Request#orderWith(Ordering)}.
+ *
+ * @since 4.13
+ */
+public abstract class Ordering {
+    private static final String CONSTRUCTOR_ERROR_FORMAT
+            = "Ordering class %s should have a public constructor with signature %s(Ordering.Context context)";
```
marcphilipp: Can't we pass in the context to the `order` method and remove this requirement?

kcooney: I prefer not to for a few reasons:

1. It would make the API confusing, since context contains data about the top-level class that has the annotation, while the other parameter is about the current set of tests that need to be ordered.
2. The Ordering might need to do expensive initialization based on the content (David's example was loading the result of the previous run so failing tests run first). If we pass context via the method then either the implementation has to do that initialization each time or it has to do lazy initialization

Note we have similar requirements on Runner implementations

kcooney:
@marcphilipp I don't see how we can do this without a new abstraction. The sorting abstraction don't work, because shuffling isn't a sort.

kcooney:
@marcphilipp I provided a sample implementation of a `Sorter` that uses a supplied `Random` instance in this commit: https://github.com/kcooney/junit/commit/ef1fde96adba60b8b4d18c8f66a9820811c07852

As you can see, it's not very pretty and not all that efficient either.

src/main/java/org/junit/runner/manipulation/Ordering.java
kcooney: One option would be to define a `Factory` interface inside of `Ordering`:

    public interface Factory {
        Ordering create(Context context);
    }

Then `OrderWith` would take in a `Class<? extends Ordering.Factory>` and the only requirement on `Factory` would be that it would need to have a public no-arg constructor.

marcphilipp: I wasn't aware that an `Ordering` instance was reused multiple times. When does that happen?

kcooney: This is similar to how Sorter works today. If you apply a Sorter to a ParentRunner it applies it to each or the child runners: https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/runners/ParentRunner.java#L418

I didn't want each runner to pass on the context, because they might do the wrong thing (create a context for the current class instead of passing on the context from the parent, which references the class annotated with `@OrderWith`)

marcphilipp: Ah, I missed that part. Thanks for the explanation! Let's leave it like it is then.

marcphilipp:
Do you want to deprecate `Sortable`etc. then? Why does `Orderable` extend `Sortable`?

kcooney:
@marcphilipp I do not want to deprecate `Sortable`. If you want to sort tests, `Sorter` is much more efficient than `Ordering` because when we use `Ordering` we need to verify that the ordering implementation didn't add or remove tests.

`Orderable` extends `Sortable` because existing runners that are `Sortable` are not necessarily `Orderable`

kcooney:
FYI: I added a factory for ordering, so now you pass an `Ordering.Factory` to `OrderWith`. The factory must have a public-no arg constructor, and has a `create` method that is passed the context. This is much more clear and is more flexible.

kcooney:
@panchenko 

Hey, Alex. Since you have been so active in JUnit recently, I wondered if you wanted to give some thoughts about this approach. Totally optional, of course.

marcphilipp:
@kcooney Thanks for the ping! I think this PR looks promising, but there are still few things to be sorted out.

Currently `Orderable` extends `Sortable`. This forces every runner that wants to implement `Orderable` to implement `Sortable` as well. Did you do it this way so `Sorter` could extend `Ordering`?

Why do we need `GeneralOrdering` instead of? I locally replaced it with `Ordering` and deleted the class and all the tests were green.

src/test/java/org/junit/tests/manipulation/ComparsionBasedOrdering.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Comparator;
+import java.util.List;
+
+import org.junit.runner.Description;
+import org.junit.runner.manipulation.Ordering;
+
+/**
+ * An ordering that internally uses a {@link Comparator}.
+ */
+class ComparsionBasedOrdering extends Ordering {
```
marcphilipp: There's a typo in the class name. How about `ComparatorBasedOrdering`?

src/main/java/org/junit/runner/manipulation/Ordering.java
```diff
@@ -0,0 +1,198 @@
...
+            throw new InvalidOrderingException(
+                    "Could not create ordering for " + annotatedTestClass, e);
+        }
+        return definedBy(factory, annotatedTestClass);
+    }
+
+    /**
+     * Creates an {@link Ordering} from the given factory.
+     *
+     * @param factoryClass class to use to create the ordering
```
marcphilipp: There's no `factoryClass` parameter.

src/main/java/org/junit/runner/manipulation/Sorter.java
```diff
@@ -33,16 +38,50 @@ public Sorter(Comparator<Description> comparator) {
...
+     * @since 4.13
+     */
+    @Override
+    protected final List<Description> orderItems(Collection<Description> descriptions) {
+        /*
+         * In practice, we will never get here--Sorters do their work in the
+         * compare() method--but the Liskov substitution principle demands that
+         * we obey the general contract of Orderable. Luckily, it's trivial to
+         * implement.
+         */
```
marcphilipp: Since this method is protected, I'm wondering whether we shouldn't just throw an `UnsupportedOperationException`. Thoughts?

src/main/java/org/junit/runner/manipulation/Ordering.java
```diff
@@ -0,0 +1,198 @@
...
+         * GeneralOrdering overrides apply() to avoid having a GenericOrdering
+         * wrap another GenericOrdering.
+         */
+        if (target instanceof Orderable) {
+            Orderable orderable = (Orderable) target;
+            orderable.order(new GeneralOrdering(this));
+        }
+    }
+
+    boolean validateOrderingIsCorrect() {
```
marcphilipp: Can you please add a comment to document what this method is used for?

src/main/java/org/junit/runner/manipulation/Sorter.java
kcooney: Since it's trivial to implement, and Ordering defines the method, I think we should implement it. A future subclass or Sorter could decide to call this method.

kcooney:
> Currently Orderable extends Sortable. This forces every runner that wants to implement
> Orderable to implement Sortable as well. Did you do it this way so Sorter could extend Ordering?

Yes, that's why it was done. I'm not too worked about forcing `Sortable` runners to implement `Orderable` for the following resons

1. If you implement `Orderable`, you need to be able to arbitrarily order children. If you could do that, you can sort the children. In fact, an `Orderable` runner _could_ implement `Sortable` by using the `Sorter` to sort the `Description` objects and then use their `Orderable` implementation to reorder the children.
1. Many runners don't implement `Sortable` directly; they implement it by extending `ParentRunner`. Those runners will automatically implement `Orderable` with this change

> Why do we need `GeneralOrdering`

If you need to do a sort, calling `Sortable.sort(sorter)` is much more efficient than calling `Orderable.order(sorter)`. `GeneralOrdering` exists so that developers call the more-efficient method.

kcooney:
Oh, and `Sorter` extends `Ordering` so `OrderWith` can work with both. I didn't want two annotations; one is simpler and it doesn't make sense to both order and sort your tests.

kcooney:
> Why do we need GeneralOrdering

I renamed GeneralOrdering to Orderer and it no longer implements Ordering. Now Orderer has the logic for doing the ordering, calling the Ordering to order the items and then validating that there are no duplicates or removed items. I think it's much simpler than what I had before, though I'm not excited about the naming.

kcooney:
@marcphilipp do my changes and comments address your concerns?

Issue #1125, opened 2015-04-16
Stop request is not properly handled if @AfterClass throws exception


ghost:
Consider the following code example:

```
@RunWith(Suite.class)
@Suite.SuiteClasses({Example1.class, Example2.class})
public class SuiteExample {
}

@FixMethodOrder(MethodSorters.NAME_ASCENDING)
public class Example1 {
    @Test
    public void test1() throws InterruptedException {
        Thread.sleep(5000);
    }

    @Test
    public void test2() {}

    @AfterClass
    public static void afterClass() {
        throw new RuntimeException("Exception");
    }
}


public class Example2 {
    @BeforeClass
    public static void beforeClass() {
        System.out.println("This is not expected to be invoked");
    }

    @Test
    public void test() {}

    @AfterClass
    public static void afterClass() throws IOException {
        System.out.println("This is not expected to be invoked");
    }
}
```
1. Execute `SuiteExample` class in the Eclipse IDE 
2. While execution of `Example1.test1()` is in the progress, hit the "Stop request" red button in the JUnit view (not in the console view, which will kill the process)
3. Execution process will be interrupted after `Example1.test1()` has been completed
4. Method `Example1.afterClass()` will be invoked, which will throw exception

Result: Method `Example2.test()` will not be invoked as expected, but methods `Example2.beforeClass()` and `Example2.afterClass()` will be invoked, while they should not be.

After successful triaging of this issue, if you don't mind I wish to create pull request for this issue. It shouldn't be too hard to fix that.

Thanks.

kcooney:
@NikitaMerinov if you have a simple fix for this, please send us a pull request. Thanks!

junit-builds:
JUnit 4 is now in maintenance mode.

At this point, only critical bugs and security issues will be fixed.

The team has therefore decided to close this issue.

PR #1131, opened 2015-04-25
Handling of StoppedByUserException will correctly stop execution in case of any additional errors coming from @AfterClass methods or class rules.

NikitaMerinov:
Due to #1125.

Several points:
1. Original code did not notify listeners about StoppedByUserException. I took the same approach for handling this exception with other execution errors - StoppedByUserException will not be delivered to listeners while all other errors will. Though I think listeners should get notification about reason of execution interruption.
2. Method for unpacking all exceptions looks a little bit helpful and at the moment the same logic is used in EachTestNotifier. I was thinking about moving that method to MultipleFailureException class to make it common and allow usage of one in different places (e.g. writers of custom runners can get benefit from one), but in that case method will become a part of the public API, and I'm not sure that JUnit needs that.
3. I don't like methods' names, but was unable to find something more appropriate.

### DIFF
--- a/src/main/java/org/junit/runners/ParentRunner.java
+++ b/src/main/java/org/junit/runners/ParentRunner.java
@@ -35,6 +35,7 @@ import org.junit.runner.notification.RunNotifier;
 import org.junit.runner.notification.StoppedByUserException;
 import org.junit.runners.model.FrameworkMethod;
 import org.junit.runners.model.InitializationError;
+import org.junit.runners.model.MultipleFailureException;
 import org.junit.runners.model.RunnerScheduler;
 import org.junit.runners.model.Statement;
 import org.junit.runners.model.TestClass;
@@ -363,11 +364,58 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
             statement.evaluate();
         } catch (AssumptionViolatedException e) {
             testNotifier.addFailedAssumption(e);
-        } catch (StoppedByUserException e) {
-            throw e;
         } catch (Throwable e) {
-            testNotifier.addFailure(e);
+            processExecutionFailures(testNotifier, gatherAllFailures(e));
         }
+    }
+
+    /**
+     * Goes through all execution failures trying to find and re-throw 
+     * {@link StoppedByUserException} while adding all other errors to
+     * test notifier.
+     * 
+     * @param testNotifier notifier to add failures
+     * @param failures list of execution failures
+     */
+    private void processExecutionFailures(EachTestNotifier testNotifier,
+            List<Throwable> failures) {
+        StoppedByUserException stoppedByUserException = null;
+        
+        for (Throwable failure : failures) {
+            if (failure instanceof StoppedByUserException) {
+                stoppedByUserException = (StoppedByUserException) failure;
+            } else {
+                testNotifier.addFailure(failure);
+            }
+        }
+        
+        if (stoppedByUserException != null) {
+            throw stoppedByUserException;
+        }
+    }
+    
+    /**
+     * Returns all failures from the specified exception going through each
+     * encountered {@link MultipleFailureException} recursively.
+     * 
+     * @param throwable input exception
+     * @return list of all failures
+     */
+    private List<Throwable> gatherAllFailures(Throwable throwable) {
+        List<Throwable> result = new ArrayList<Throwable>();
+        
+        if (throwable instanceof MultipleFailureException) {
+            MultipleFailureException multipleFailureException =
+                    (MultipleFailureException) throwable;
+            
+            for (Throwable failure : multipleFailureException.getFailures()) {
+                result.addAll(gatherAllFailures(failure));
+            }
+        } else {
+            result.add(throwable);
+        }
+        
+        return result;
     }
 
     //
--- a/src/test/java/org/junit/tests/listening/UserStopTest.java
+++ b/src/test/java/org/junit/tests/listening/UserStopTest.java
@@ -1,23 +1,41 @@
 package org.junit.tests.listening;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.AfterClass;
 import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
 import org.junit.Test;
 import org.junit.runner.Request;
+import org.junit.runner.RunWith;
+import org.junit.runner.notification.Failure;
+import org.junit.runner.notification.RunListener;
 import org.junit.runner.notification.RunNotifier;
 import org.junit.runner.notification.StoppedByUserException;
+import org.junit.runners.Suite;
+import org.junit.runners.Suite.SuiteClasses;
 
 public class UserStopTest {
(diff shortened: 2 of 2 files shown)


src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -363,13 +364,60 @@ public void run(final RunNotifier notifier) {
             statement.evaluate();
         } catch (AssumptionViolatedException e) {
             testNotifier.addFailedAssumption(e);
-        } catch (StoppedByUserException e) {
```
kcooney: Why was the catch block for `StoppedByUserException` removed?

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -363,13 +364,60 @@ public void run(final RunNotifier notifier) {
...
+     */
+    private void processExecutionFailures(EachTestNotifier testNotifier,
+            List<Throwable> failures) {
+        StoppedByUserException stoppedByUserException = null;
+        
+        for (Throwable failure : failures) {
+            if (failure instanceof StoppedByUserException) {
+                stoppedByUserException = (StoppedByUserException) failure;
+            } else {
+                testNotifier.addFailure(failure);
```
kcooney: This is a behavior change. You are sending to the notifier each of the failures that was contained in the `MultipleFailure` exception, when before it was notified once.

It sounds the problem is that a `StoppedByUserException` wrapped in a `MultipleFailureException` doesn't cause `run(RunNotifier)` to exit. Couldn't you check if any of the exceptions inside of the `MultipleFailureException` is a `StoppedByUserException`, and if so, throw that `StoppedByUserException` after sending the `MultipleFailureException` to the notifier?

Also, where is the `MultiFailureException` being created that contains the `StoppedByUserException`? It's possible that code is in error.

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -363,13 +364,60 @@ public void run(final RunNotifier notifier) {
             statement.evaluate();
         } catch (AssumptionViolatedException e) {
             testNotifier.addFailedAssumption(e);
-        } catch (StoppedByUserException e) {
-            throw e;
         } catch (Throwable e) {
-            testNotifier.addFailure(e);
```
kcooney: I have a simpler suggestion that still works. We could add a method to `RunNotifier` called `stopRequested()` that returns `true` if `pleaseStop()` was called. Right here, after calling `addFailure()`, we would call that method, and if it returns `true` throw `StoppedByUserException`

Note that in #1118 someone is proposing adding `fireTestSuiteStarted()` and `fireTestSuiteFinished()` to `RunNotifier`. Those methods would be other good places to throw `StoppedByUserException()`

src/main/java/org/junit/runners/ParentRunner.java
NikitaMerinov: Now `StoppedByUserException` is handled in the `processExecutionFailures()` method, despite single or multiple exceptions came from evaluating class block statement.

src/main/java/org/junit/runners/ParentRunner.java
NikitaMerinov: > This is a behavior change. You are sending to the notifier each of the failures ...

Behavior change from what point of view? Before and after my changes no clients are getting multiple exceptions at once. As I said in item 2 in my pull request message, class `EachTestNotifier` does the same as I do - it unpacks all exceptions and notifies clients about them one by one.

> It sounds the problem is that a StoppedByUserException wrapped ...

Sending `MultipleFailureException` to notifier before finding `StoppedByUserException` will lead to behavior changes. Clients will get notifications about `StoppedByUserException` what is never happened before. My first item in the pull request message was about that. If we are ok with these behavior changes, then it can simplify the code.

> Also, where is the MultiFailureException being created that contains ...

You can get full example from #1125 or check new test I've added. Basically, `StoppedByUserException` is combined with other exceptions while evaluating 'after' methods. I don't think that this is an error, since JUnit guarantees to invoke all 'after' methods in case of user stop request and if any of 'after' methods will throw exception I believe clients should get notification about them.

src/main/java/org/junit/runners/ParentRunner.java
NikitaMerinov: I agree, this was one of my first attempt to fix this issue, but returning to my points in the previous conversation - if we will simply pass `MultipleFailureException` to notifier we will get behavior changes, therefore I decided to unpack all failures manually and notify listeners about all errors except `StoppedByUserException`.
Other solution can be add filtering inside `EachTestNotifier` - modify `addFailure()` method to skip `StoppedByUserException` and pass all other errors. This can make code even simpler, but again this will lead to behavior changes - such alteration will make JUnit to completely ignore `StoppedByUserException` thrown from the test method manually, which I believe is not acceptable.

kcooney: @NikitaMerinov So is there a problem with my suggestion to have `ParentRumner` notice that a stop was requested and throw a new `StoppedByUserException` (or,  better yet, do this in a listener after we add suite start/stop calls?

NikitaMerinov: @kcooney Partially. Your suggestion will prevents other classes from execution, but we still need to do something with notifying clients about `StoppedByUserException`.

kcooney: @NikitaMerinov if we don't want to notify users about that `StoppedByUserException` when it is part of a `MultiFailureException`, you can update `EachTestNotifier` to do that

NikitaMerinov: As it was said, updating `EachTestNotifier` with these changes will make JUnit ignoring `StoppedByUserException` thrown manually from test method.

But, in this case, we can distinguish `StoppedByUserException` thrown from test method by checking `Description.isTest()` inside `EachTestNotifier`. Looks like it can be an appropriate solution.
@kcooney Are we ok with that approach?

NikitaMerinov:
Closing this pull request since original fix was not accepted. Discussion with kcooney helped to figure out better approach to fix the issue.

@kcooney I believe it will be more appropriate if JUnit team will fix this issue when #1118 will be merged successfully.

kcooney:
@NarendraPathai good point about tests throwing `StoppedByUserException` directly. I don't see a reason why someone would have a test do that, but if they did, failing the test is the most reasonable thing to do.

Perhaps it's best if we don't worry about listeners being notified too often (or not often enough) of the `StoppedByUserException`. The listeners that are tracking the pass/fail status of tests should "notice" that some tests weren't run, and presumably the user would know why.

I hope I didn't push you away with all of the comments. There is no one correct solution to this, and making changes to the core of JUnit without adversely affecting existing code is tricky.

Note that the majority of changes to JUnit aren't made by the JUnit team. The JUnit team reviews changes, do merges, update release notes and do releases, but most functional changes are from contributors like you. If you want to take another stab at this after #1118 is merged, please feel free.

NikitaMerinov:
Sure, I understand. From my point of view, I just think that person who has suggested better approach is the first candidate to implement that approach :-) And if you will not be able to do that, then someone else will be.

kcooney:
I think it depends on how difficult it is to implement the "better" approach, how much time everyone has, and how important the feature or fix is.

Issue #1133, opened 2015-04-29
Add lazy evaluation for assertion messages


ruckc:
It would be nice if `assertEquals(String message, Object expected, Object actual)` was overloaded with a lazy evaluated String.format style Message object.

Example:
   `assertEquals(Message message, Object expected, Object actual)`
Usage:
   `assertEquals(message("This is bad %s %s", expected, actual), expected, actual)`

the message() could be added to Assert for inclusion in the static includes, and it could return a lazy eval Message object.  This would enable faster unit tests while minimizing complexity in unit tests.  Reference https://twitter.com/NathanTippy/status/593435755453677568

nathantippy:
Also take a peak at how SLF4j does this for faster logging when the log level may not go that deep.

kcooney:
•assertEquals()` already includes the expected and actual values in the exceptuon message. Would you not want it to do that if you passed in whatever value is returned from `message()`? Could you pass other optional.values?

BTW, I am.not currently concerned about performance here. Assertion failures should be rare and fixed right away, and the methods throw exceptions on failure. Logging framework gave different performance needs.

nathantippy:
The situation happens when there are cases that the expected and actual values are not enough information to debug the failure.  I want to see more system state to help me find the problem.  Gathering this information takes time and should only be done upon failure.  I have a work around using a finally block however it is less than elegant.

When I have a large % of coverage in projects it takes a long time get a good build due to the time it takes the test to run. If every assertEquals is doing extra object creation that will never be used then I see a growing impact of this wasted work with each test I add.  

For very large projects performance of the passing case becomes very important.

ruckc:
I've submitted a potential Pull request for this in #1134

kcooney:
Can you give a specific example where the expected and actual values are not enough to debug a failure and this proposed API would help (including a code snippet of how you would use this API to solve the problem)? Thanks!

nathantippy:
I would like to redirect the conversation and continue with #1134

kcooney:
@nathantippy could you please answer my previous question? The pull is addressing a specific solution to a problem. I'd like to make sure we're all on the same page about what the problem is, and discuss what APIs would be best for solving that problem.

nathantippy:
There are times when there is a matching failure between expected and
actual however those two numbers are not enough for diagnosis.
I would like to present to the developer the last 10 successful values in
the stream.
This gives valuable context for the failure to help the developer determine
where the fault happened.


[...]

PR #1134, opened 2015-04-29
Use widing refactoring to allow passing a lazy evaluated CharSequence (#1133)

ruckc:
This allows for using the included `Message.message()` to generate lazy evaluated CharSequence based on String.format

### DIFF
--- a/src/main/java/org/junit/Message.java
+++ b/src/main/java/org/junit/Message.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2015 JUnit.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    JUnit - initial API and implementation and/or initial documentation
+ */
+package org.junit;
+
+/**
+ *
+ * @author ruckc
+ */
+public class Message implements CharSequence {
+    private String evaluated;
+    private final String format;
+    private final Object[] values;
+    
+    public Message(final String format, final Object ... values) {
+        this.format = format;
+        this.values = values;
+    }
+    
+    private void evaluate() {
+        if(evaluated==null) {
+            evaluated = String.format(format, values);
+        }
+    }
+
+    public int length() {
+        evaluate();
+        return evaluated.length();
+    }
+
+    public char charAt(int index) {
+        evaluate();
+        return evaluated.charAt(index);
+    }
+
+    public CharSequence subSequence(int start, int end) {
+        evaluate();
+        return evaluated.subSequence(start, end);
+    }
+    
+    @Override
+    public String toString() {
+        evaluate();
+        return evaluated;
+    }
+    
+    public static Message message(String format, Object ... arguments) {
+        return new Message(format, arguments);
+    }
+    
+    public static Message msg(String format, Object ... arguments) {
+        return new Message(format, arguments);
+    }
+}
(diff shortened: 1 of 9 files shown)


nathantippy:
This is expected to be a large change but one with little risk.  All Strings are instances of CharSequence.  The message parameter is of type String and only used when the test fails.  If message were a CharSequence the existing behavior would not change.   

The benefit of this change would be to enable some classes that extend CharSquence to permit lazy construction of the message.  In this way expensive detail gathering could be one only upon failure, eg use of the message.

src/main/java/org/junit/Message.java
```diff
@@ -0,0 +1,61 @@
...
+    public String toString() {
+        evaluate();
+        return evaluated;
+    }
+    
+    public static Message message(String format, Object ... arguments) {
+        return new Message(format, arguments);
+    }
+    
+    public static Message msg(String format, Object ... arguments) {
```
ashleyfrieze: Why have both a msg and a message function? Is this duplication a real benefit?

ashleyfrieze:
I like the idea of this, but I'm not sure about the reality. A couple of observations. One, any third party extensions to JUnit will continue to use Strings for messages, rather than CharSequence, which may lead to people having to blend between Message and String. Two: the painful effort in initialisation of strings in real life, is usually in retrieving what to put into the string, not making the string itself. That won't be removed by the Message pattern.

E.g. message("Something: %s", something.toString()); // the effort is in compiling the output of something.toString, not in compiling the message - something.toString() might require a lot of fields to be examined by reflection or something. That would be evaluated when creating the Message object, even though it doesn't run through String.format if it's not used.

If this could be used with lambdas, then it would make a difference.

e.g. message("Something: %s", something -> something.toString());

In this situation, the lambda would not be evaluated at the point of creating Message and would genuinely save time if the message were never needed.

kcooney:
@ashleyfrieze you could postpone producing the value to be put into the string by passing in an Object as a parameter, and producing the value in the object's `toString()`, but agreed that the lambda syntax is cleaner

ashleyfrieze:
@kcooney love your thinking. But practically it would be this...

From:

    new Message("Broken %s", myBigThing.getBigData());

to

    new Message("Broken %S", new Object() { @Override public void toString() { myBigThing.getBigData()); }});

That's quite a workaround :)

ashleyfrieze:
Perhaps there's an alternative to this by adding an overload to assertThat allowing for a just-in-time message creation.

    public interface Message {
        String getMessage();
    }


    public void assertThat(Message message, Object o, Matcher m) {
        ...
        // pseudo code
        fail(message.getMessage);
     }

called thusly:

    assertThat(new Message() { public String getMessage() { return myObject.getBigData(); }},
        someValue, is(expected));

The benefit of changing just assertThat is that it's a smaller change. The benefit of the Message interface is that it could easily be subclassed by the user for reusable messages. We could also provide a Java 8 message provider which uses lambda:

    assertThat(new LambdaMessage(() -> myObject.getBigData()), someValue, is(expected));

Just tossing the ideas out here.

marcphilipp:
Actually, the equivalent lambda expression would be:

```java
assertThat(() -> myObject.getBigData(), someValue, is(expected));
```

marcphilipp:
However, we want to get rid of the Hamcrest dependency in the future. Can you try to submit a similar pull request there for [`MatcherAssert`](https://github.com/hamcrest/JavaHamcrest/blob/master/hamcrest-core/src/main/java/org/hamcrest/MatcherAssert.java)?

ashleyfrieze:
@marcphilipp my suggestion was for a Java 8 adapter over a Java 5 interface. If everything were Java 8 this would all be moot!

marcphilipp:
I understand. However, for `assertThat` I think you should ask over at hamcrest/JavaHamcrest.

ashleyfrieze:
See https://github.com/hamcrest/JavaHamcrest/pull/102 - which is this turned into a potential hamcrest change.

kcooney:
I'm closing this since mark closed 1133

PR #1137, opened 2015-05-03
Suggested improvement - use the TestClass object to create instances of the test object

ashleyfrieze:
This allows for a runner to be created by something
which intends to modify the test object before the test starts. This can
lead to one runner executing a different sort of runner in a chain of responsibility pattern.

It could allow an outer class to have one runner and a static inner class to have another test runner, which can have values injected into its test object by the parent outer class.

Example - a MockitoJunitRunner could be used to inject values into
a test object running in a SpringJUnitRunner inner class. This would
require those runners to change too, of course.

Right now, the problem with runners is that they're monolithic and use
inheritance to share features. This change is intended to allow more
injection of behaviour/data across runners via the test object.


kcooney:
Looks like you changed all the line endings. Can you did that so we can see the diffs?

ashleyfrieze:
Rookie mistake. Will sort it out.

ashleyfrieze:
I've fixed the line endings. Feedback on this idea would be appreciated.

ashleyfrieze:
I'm going to make this one again when I'm properly synced with 4.13

kcooney:
Chaining runners this way sounds a bit fragile. If the intention is to be able to change or augment the way the test class is created I would personally prefer something like TestNG's `@Factory` annotation.

PR #1138, opened 2015-05-04
TestCases: common @Test methods called multiple times, with parameters, by BlockJUnit4Runner

ashleyfrieze:
Similar to the intent of pull request 1115 - https://github.com/junit-team/junit/pull/1115 - and as a way to provide the same behaviour of NUnit's [TestCase] mechanism - http://www.nunit.org/index.php?p=testCase&r=2.5 - this feature suggestion would allow for common test method implementation to be written once in a given test class, with the test annotation specifying the inputs to try it with.

I am aware that these changes may graze a few sacred cows, so I present this for review. I've done some refactoring within the tests of the project, to support my own tests, which I hope are useful in their own right too - the MockTestRunner being the centre of it.

There is a real-world problem to solve here. I commonly end up having to write tests like this:

```
@Test
public void example1() {
    check(1, 2, 3);
}

@Test
public void example2() {
    check(2, 3, 5);
}

private void check(int a, int b, int c) {
    assertEquals(c, a+b);
}
```

The check function is doing the actual work, but the examples each need their own @Test annotated method.

With this change, the check function becomes the test method

```
@Test(cases={
    @TestCase({"1","2","3"}),
    @TestCase({"2","3","4"})
    })
public void check(int a, int b, int c) {
    assertEquals(c, a+b);
}
```

You can do this with Parameterized, but only if you create a whole class around the parameters you're testing with, and it's quite clunky and not interoperable with other test runners.

You can do this with JUnitParams with a similar syntax but that means you can only use the JUnitParams runner. Putting this feature into BlockJUnit4TestRunner will make it a central part of the library, which means a lot of other test runners, which reuse Block, will inherit this capability, e.g. Mockito.


Stephan202:
Without otherwise voicing an opinion on the merit of this change, I think it's worth pointing out that the first piece of example test code (i.e., the one that represents a "real-world problem") forces one to name each test. This allows one (or forces one, depending on ones perspective) to come up with descriptive names for each test case. This can be overkill or extremely helpful, depending on context. So in that sense the "before" and "after" example code is not equivalent.

ashleyfrieze:
@Stephan202 really good point. This is a reason to allow for mixing between individual named tests and groups of tests where it is only the test data which really tell them apart. Example, I once wrote unit tests for a conversion from integer to natural language. The test data was self describing.

kcooney:
@ashleyfrieze Have you looked at https://github.com/Pragmatists/junitparams ? That seems to solve this problem, and has multiple ways of specifying parameters (including specifying a method to call)

ashleyfrieze:
@kcooney I've looked at JUnitParams and I don't like it as a solution. The reason for this is interoperability. There are lots of JUnit runners out there and many of them work by extending Block if you choose to write tests using SpringJUnitRunner, MockitoRunner, or Parameterized, then you can't use JUnitParams too. If we extend Block to allow for this syntax, then it becomes common to a whole family of runners.

To be honest, the real limitation here is that JUnit runners are monolithic and their functionality comes through inheritance, rather than composition.

If the goal were to have rich parameterisation, then a specialist runner like JUnitParams would outperform this implementation. If, like me, you want to mix some primitive parameterized test cases into your usual tests with your favourite runner, then this approach seems simple enough to work.

BTW - even if this whole pull can't be taken, please check out the first couple of commits where I extract a handy test utility for checking the behaviour of runners.

kcooney:
The test utility work looks interesting. I suggest a separate pull for that, and modify a few of our existing tests tests to use it. Some of our tests verify the ordering of the calls to the listener methods, not just the count, so you might want to look at those tests to see if we can make them simpler, too. I'd personally prefer a different name, since `MockTestRunner` isn't a mock object, and isn't a runner.

I have some thoughts on the parameterized changes, but I'm off to a ballgame, so those will have to wait :-)

kcooney:

Given that JUnit 5 is around the corner, I don't think it makes sense to make more ways to do parameterized tests in JUnit 4. I vote we close this

ashleyfrieze:
It pains me to say this, but I agree with you. Rather than close it myself, let's see if anyone else wants to adopt it as a java7 solution for this sort of thing... if not, I think it's time to declare it obsolete.

ashleyfrieze:
@kcooney would there still be merit in the test helpers I refactored getting into the junit tests? In other words, will junit 4 see a new version at all? If it's going to be maintained and slightly extended, maybe I should create a new pull request for some of the test tweaks I did... whaddya think?

kcooney:
@ashleyfrieze I still hope we can have a 4.13 release in the next ~6 months.

PR #1140, opened 2015-05-11
Removed usage of Double Checked Locking.

pwagland:
Given that Double Checked Locking is not thread safe, it should not be used.


kcooney:
I think your information is out of date. That field is volatile, so it is thread safe

pwagland:
My bad, I missed that the variable was volatile! Sorry about that. nine out of ten times when that pattern is seen it is incorrect. In this case it was OK. Thanks for checking.

PR #1141, opened 2015-05-12
Some sonar fixes

baev:


src/main/java/org/junit/experimental/theories/ParameterSignature.java
```diff
@@ -34,7 +34,7 @@
         map.put(b, a);
     }
     
-    public static ArrayList<ParameterSignature> signatures(Method method) {
+    public static List<ParameterSignature> signatures(Method method) {
```
kcooney: This change could break people. Ditto for the next one

kcooney:
Thanks for the fixes! The 'static public' bothers me too :-)

src/main/java/org/junit/experimental/theories/ParameterSignature.java
baev: Is there way to migrate to `List` without breaking people?

kcooney: Not in public API signatures. If we have a public method that returns an `ArrayList` we can't change it to return a `List`:

    ArrayList<String> s = ParameterSignature.signatures(method); // woops!

baev: I understand. I ask about some best practices like a "add deprecation with a comment about migration to `List` in next releases". Btw I put it back to the `ArrayList`

kcooney: I personally don't think it's worth it to deprecate the method and replace it with a similarly-named method that returns a `List`

src/main/java/org/junit/Assume.java
```diff
@@ -34,6 +34,15 @@
  * @since 4.4
  */
 public class Assume {
+
+    /**
+     * Do not use instance if this class.
```
kcooney: "if" -> "of"

kcooney:
LGTM. I can fix the minor problems that are left when I merge this.

@junit-team/junit-committers can one of you please take a look?

baev:
@kcooney done

src/main/java/org/junit/Assert.java
```diff
@@ -591,33 +592,21 @@ static public void assertEquals(String message, float expected,
...
-    static public void assertNotEquals(String message, float unexpected,
+    public static void assertNotEquals(String message, float unexpected,
             float actual, float delta) {
         if (!floatIsDifferent(unexpected, actual, delta)) {
-            failEquals(message, Float.valueOf(actual));
+            failEquals(message, actual);
         }
     }
 
-    static private boolean doubleIsDifferent(double d1, double d2, double delta) {
```
dsaff: These implementations (floatIsDifferent, doubleIsDifferent) were deliberate stylistic choices for readability.  I don't feel strongly about the style, but feel that muddying the git history for a different stylistic change doesn't seem worth it.

src/main/java/org/junit/Assume.java
```diff
@@ -34,6 +34,15 @@
  * @since 4.4
  */
 public class Assume {
+
+    /**
+     * Do not instantiate.
+     * @deprecated will be private soon.
```
dsaff: "soon" seems too vague to be useful.  I think I'd prefer "deprecated since 4.13"; if people want to know what that means, they can look up the deprecation policy.

src/main/java/org/junit/experimental/theories/internal/Assignments.java
```diff
@@ -47,7 +47,7 @@ public static Assignments allUnassigned(Method testMethod,
     }
 
     public boolean isComplete() {
-        return unassigned.size() == 0;
+        return unassigned.isEmpty();
```
dsaff: Are we requiring Java 1.6 now?  I honestly don't remember, sorry.

src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
```diff
@@ -379,12 +382,13 @@ private Statement withRules(FrameworkMethod method, Object target,
 
     private Statement withMethodRules(FrameworkMethod method, List<TestRule> testRules,
             Object target, Statement result) {
+        Statement withMethodRules = result;
         for (org.junit.rules.MethodRule each : getMethodRules(target)) {
-            if (!testRules.contains(each)) {
-                result = each.apply(result, method, target);
+            if (!(each instanceof TestRule && testRules.contains(each))) {
```
dsaff: What's the motivation here?

src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
```diff
@@ -379,12 +382,13 @@ private Statement withRules(FrameworkMethod method, Object target,
 
     private Statement withMethodRules(FrameworkMethod method, List<TestRule> testRules,
             Object target, Statement result) {
+        Statement withMethodRules = result;
```
dsaff: What's the motivation here?

dsaff:
Mostly great stuff.  I've commented the few places that give me pause: in places where there's at best a minor improvement in style, or just a fit to a different preference, I'd prefer not to make the change, a lesson learned since our "git blame" history became largely useless after we "fixed" all the indentation.

src/main/java/org/junit/experimental/theories/internal/Assignments.java
baev: https://docs.oracle.com/javase/1.5.0/docs/api/java/util/List.html#isEmpty()

src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
baev: More clear way to specify why we are looking for the `MethodRule` in list of  `TestRule ` isn't?

src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
baev: http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1226

src/main/java/org/junit/experimental/theories/internal/Assignments.java
dsaff: Got it, thanks.

src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
dsaff: Makes sense.

src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
dsaff: Got it.

src/main/java/org/junit/Assert.java
baev: fixed

baev:
@dsaff thx for the review. All the comments are fixed now.

src/main/java/org/junit/experimental/results/ResultMatchers.java
```diff
@@ -14,6 +14,15 @@
  * </pre>
  */
 public class ResultMatchers {
+
+    /**
+     * Do not instantiate.
+     * @deprecated will be private soon.
```
marcphilipp: since 4.13

src/main/java/org/junit/internal/Classes.java
```diff
@@ -6,6 +6,15 @@
  * Miscellaneous functions dealing with classes.
  */
 public class Classes {
+
+    /**
+     * Do not instantiate.
+     * @deprecated will be private soon.
```
marcphilipp: since 4.13

marcphilipp:
@baev Thanks, I've commented two minor things inline.

marcphilipp:
LGTM!

@dsaff Do you want to take another look?

dsaff:
LGTM2!

PR #1147, opened 2015-05-22
Depend on JDK 7

### LINKED ISSUE #1146: Update to Java 7
We've already decided to move on to Java 6 in JUnit 5. Java 6 has reached end-of-life over two years ago. I think we should skip JDK 6 and move on to JDK 7.

marcphilipp:
Fixes #1146


Tibor17:
Java 7 is out of date as well.
Changing junit5 mind again and no more Lambda?
AFAIK junit5 was supposed to go with java 8.

kcooney:
Why are we moving to JDK 7? JUnit is used in many environments (including Android and GWT), some of which cannot easily upgrade their JDK, so I think we should change the required JDK slowly.

PR #1148, opened 2015-05-22
Remove Hamcrest compile dependency

### LINKED ISSUE #1145: Decouple JUnit from Hamcrest
@npryce has nicely summed up the reasons to do this in his blog:
http://www.natpryce.com/articles/000806.html

marcphilipp:
Use hamcrest-junit 2.0.0.0 instead.

Fixes #1145. Depends on #1147.

This is in an early stage. I'm aware that we would need to deprecated the classes/methods I've deleted in 4.13. I'm looking for your feedback about the general effort.

@junit-team/junit-committers Opinions?


kcooney:
Why does this change require JDK 7?

As much as I would love to see Hamcrest go away, given our deprecation policy, if we remove these in JUnit 5, we wouldn't be able to release JUnit 5 until two years after JUnit 4.13 is released.

kcooney:
I see... The hamcrest-junit library requires JDK 7. I posted a message on https://github.com/hamcrest/JavaHamcrest/issues/92

marcphilipp:
Closing this pull request because JUnit 5 won't depend on JDK 7 and therefore we cannot use Hamcrest 2. I will submit another pull request that uses Hamcrest 1.3 along with hamcrest-junit 1.x.

kcooney:
Thanks, Marc. I suggest we not delete APIs in that pull, just migrate our tests to java-hamcrest

PR #1149, opened 2015-05-23
Fix incorrect javadoc of the Rule running order

Jimmy-Shi:
In the javadoc of Rule and ClassRule, it says field rules are applied before method rules. Actually, method rules are applied before field rules.

### DIFF
--- a/src/main/java/org/junit/ClassRule.java
+++ b/src/main/java/org/junit/ClassRule.java
@@ -28,7 +28,8 @@ import java.lang.annotation.Target;
  * annotated {@link ClassRule}s on a class, they will be applied in an order
  * that depends on your JVM's implementation of the reflection API, which is
  * undefined, in general. However, Rules defined by fields will always be applied
- * before Rules defined by methods.
+ * after Rules defined by methods, i.e. the Statements returned by the former will
+ * be executed around those returned by the latter.
  * <p>
  * For example, here is a test suite that connects to a server once before
  * all the test classes run, and disconnects after they are finished:
--- a/src/main/java/org/junit/Rule.java
+++ b/src/main/java/org/junit/Rule.java
@@ -16,12 +16,14 @@ import java.lang.annotation.Target;
  * to the {@link org.junit.rules.TestRule} will run any {@link Before} methods,
  * then the {@link Test} method, and finally any {@link After} methods,
  * throwing an exception if any of these fail.  If there are multiple
- * annotated {@link Rule}s on a class, they will be applied in order of fields first, then methods.
+ * annotated {@link Rule}s on a class, they will be applied in order of methods first, then fields.
  * However, if there are multiple fields (or methods) they will be applied in an order
  * that depends on your JVM's implementation of the reflection API, which is
  * undefined, in general. Rules defined by fields will always be applied
- * before Rules defined by methods. You can use a {@link org.junit.rules.RuleChain} if you want
- * to have control over the order in which the Rules are applied.
+ * after Rules defined by methods, i.e. the Statements returned by the former will
+ * be executed around those returned by the latter. You can use a {@link org.junit.rules.RuleChain} 
+ * if you want to have control over the order in which the Rules are applied.
+ *
  * <p>
  * For example, here is a test class that creates a temporary folder before
  * each test method, and deletes it after each:


Jimmy-Shi:
Why this pull request is not getting merged for 4 months?

marcphilipp:
Sorry for the delay. I just double-checked: Field rules *are* applied before method rules.

### Example

```java
public class Example {

	@Rule
	public TestWatcher methodWatcher() {
		return testWatcher("method");
	}

	@Rule
	public TestWatcher fieldWatcher = testWatcher("field");
	
	@Test
	public void test() {
		System.out.println("test");
	}

	private static TestWatcher testWatcher(final String prefix) {
		return new TestWatcher() {
			@Override
			protected void starting(Description description) {
				System.out.println(prefix + ": starting");
			}

			@Override
			protected void finished(Description description) {
				System.out.println(prefix + ": finished");
			}
		};
	}

}
```

This prints:

```
field: starting
method: starting
test
method: finished
field: finished
```

---

Do you have an example where methods are actually applied before fields?

kcooney:
Actually, technically field rules are applied after method rules, but when a rule is applied the new rule runs "around" the existing rule chain.

Jimmy-Shi:
@marcphilipp 
I run your example and check my example again. The conclusion is: 
* My Example shows that: Method rules are applied before field rules.
* You example shows that: Statements returned by field rules' apply methods are evaluated before statements returned by method rules' apply method.

My Example:
```java
public class MethodOrfieldFirst {
    @Rule
    public TestRule fieldRule = new MyRule("FieldRule");

    @Rule
    public TestRule methodRule() {
        return new MyRule("MethodRule");
    }

    @Test
    public void test() {
        System.out.println("Running Test");
    }

    static class MyRule implements TestRule {

        private String name;

        MyRule(String name) {
            this.name = name;
        }

        @Override
        public Statement apply(Statement base, Description description) {
            System.out.println("applying: "+ name);
            return base;
        }
    }
}
```

This prints:
```
applying: MethodRule
applying: FieldRule
Running Test
```

marcphilipp:
Thanks for your example. I now get what you meant. I think the original intention of the Javadoc was do describe the execution order of the `Statements` returned by rules. We should definitely clarify this.

How about:

> Rules defined by fields will always be applied after Rules defined by methods, i.e. the Statements returned by the former will be executed around those returned by the latter.

Is that clearer? I'm not really satisfied with it and open to your suggestions.

Jimmy-Shi:
Besides exchange the word `method` and `field`, I added below comment and update the pull request:
>
Since almost all of the in-box org.junit.rules.TestRule implementations are adding decorators to the passed in statement, the added decorators are evaluated in a reversed order that rules are applied. 

Is it better?

kcooney:
Which rules don't add decorators to the statement chain?

Jimmy-Shi:
Some Rules added in the feature.

src/main/java/org/junit/ClassRule.java
```diff
@@ -27,8 +27,11 @@
  * If there are multiple
  * annotated {@link ClassRule}s on a class, they will be applied in an order
  * that depends on your JVM's implementation of the reflection API, which is
- * undefined, in general. However, Rules defined by fields will always be applied
- * before Rules defined by methods.
+ * undefined, in general. However, Rules defined by methods will always be applied
+ * before Rules defined by fields.
+ * Since almost all of the in-box {@link org.junit.rules.TestRule} implementations are adding
```
kcooney: All in-box rule implementations are adding decorators (in fact, all rules are decorators) so this is confusing. Can we fix it?

Jimmy-Shi: Sorry for late response, I was on a long holiday without computer
How about let's remove the word **almost**

kcooney: How about "Since all {@link org.junit.rules.TestRule} implementations..." ?

Jimmy-Shi: Implementing with decorator pattern is not a requirement. We should not say "all" with out the limitation "in-box". What do you think.

kcooney: All rules are decorate the statement chain. Even if a given rule doesn't "look like" a decorator it is one

Jimmy-Shi: I could not get your point. I could not see any decorating in below implementation.
```java
    static class MyRule implements TestRule {

        private String name;

        MyRule(String name) {
            this.name = name;
        }

        @Override
        public Statement apply(Statement base, Description description) {
            System.out.println("applying: "+ name);
            return base;
        }
    }
```

kcooney: No one should write rules like that. The implementation of the rule should occur when the returned statement is run.

A test rule is a factory for decorators for statements. On very rare occasions, that factory might returned the passed-in statement but even then it is clearly the decorator pattern.

If you still disagree, let's drop this sentence completely.

Jimmy-Shi: Yes, I disagree. Let's drop this sentence and take Marc Philipp's 
> 
Rules defined by fields will always be applied after Rules defined by methods, i.e. the Statements returned by the former will be executed around those returned by the latter.

kcooney: SGTM

Jimmy-Shi: Updated ^_^

PR #1152, opened 2015-05-31
Make exception message more descriptive

shadowfacts:


marcphilipp:
I'm not a native speaker but I think "void method" is a set expression. 

@dsaff @kcooney What do you think?

kcooney:
I think "Method foo() should have a return type of void" or "Method foo() should have a void return type" would both be fine. I am personaly not a fan of appending "'s" to the method name.

dsaff:
Agreed that the space after the method name needs to go back in.  "Method foo() should return void" seems succinct and correct, as another option.

dsaff:
Thanks for doing this.

kcooney:
Squashed commits and merged. Thanks!

Issue #1106, opened 2015-03-30
ExpectedException Cannot Be Chained After MethodRules


EarthCitizen:
I recently had an issue with a 3rd-party Rule which extends MethodRule. This rule modified the runtime in a way in which the expectations set up in ExpectedException were erased by the time that the test actually ran. So my first thought was to just chain ExpectedException after the 3rd-party rule. As, it turns out this is not possible because the 3rd-party rule extends MethodRule.

I was able to wrap the MethodRule inside a TestRule, but you have to pass "this" to the constructor. And based on many conversations and other issues I have seen around this topic, there seems to be a general agreement that passing "this" to a TestRule constructor is ugly and undesirable and that being given access to the test class instance is helpful and reasonable. I immediately see this as useful for injecting values or processing other custom annotations.

After taking a look at the code, it seems that there needs to be a slight redesign of the rule system and that a simple patch would not suffice and discussion about possible redesigns would be needed to solve this issue.

My personal opinion is that there should be one interface for class rules, and once interface for test rules, and that the test rule should give you the instance of the test class. It does not seem to be best to continue on with two different interfaces that can be applied at the test level (MethodRule and TestRule).

Having access to the test-class instance for the test-level rules seems to provide the best options for what can be accomplished at that level. I am willing to help out in any way I can if there is a decision made on a course of action.

kcooney:
@EarthCitizen There are many issues being discussed here. I'll try to take the issues one by one, but I think we should close this and redirect the conversation to the other related bugs.

1. ExpectedException not supporting chaining and being affected by other rules

I suspect that we'll make a Lambda-based exception API and deprecate `ExpectedException`.

2. Third-party libraries that use `MethodRule` so cannot be chained

Unfortunately,  the `MethodRule` API isn't compatible with `TestRule` so I don't see a way to support chaining. The options I see are either wrap the rule with a `TestRule` (quite ugly) or convince the third-party owner that you want to use their rule in a rule chain and/or at the test level, and therefore need a version of the API that implements `TestRule`

3. Implement a new API to replace and/or augment `TestRule` and `MethodRule`

I agree we should consider new APIs. I suggest adding to the discussion on #793

Adding a new API doesn't solve the problems that the third-party libraries implement `MethodRule`. It would probably be easier to convince them to add a `TestRule` API because `TestRule` has been around a while.

4. The test instance isn't passed to `TestRule`

This has been discussed in #351

EarthCitizen:
[...]

PR #1153, opened 2015-06-03
Add assertThrows

rschmitt:
This is a rough cut (no Javadoc, public interfaces in Assert, questionable formatting, etc), but I want to first get consensus that these methods are worth adding. The goal here is to provide (in a backwards-compatible manner) a Lambda-based exception API as hinted at by #1106. (It's important to comprehend that lambdas are provided by the call site; there is no reason why a JDK5-based library cannot facilitate the use of lambda expressions, method references, and higher order functions in general.)

I have not implemented assertThrows in terms of Matchers, since the Hamcrest stuff appears to be on the way out (and rightly so).

### DIFF
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -6,6 +6,9 @@ import org.junit.internal.ArrayComparisonFailure;
 import org.junit.internal.ExactComparisonCriteria;
 import org.junit.internal.InexactComparisonCriteria;
 
+import java.util.concurrent.Callable;
+import java.util.concurrent.Executors;
+
 /**
  * A set of assertion methods useful for writing tests. Only failed assertions
  * are recorded. These methods can be used directly:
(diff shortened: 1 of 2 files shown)


kcooney:
Thanks for starting the conversation.

Some of us were leaning towards having these methods return the exception that was thrown, so the user can use their favorite assertion framework (Truth, Hamcrest, Fest, etc) to make further assertions on the message, cause or custom attributes of the test

rschmitt:
That's an interesting idea. I guess that would look something like:

```java
expectThrow(client::invoke); // throw anything
assertTrue(expectThrow(client::invoke) instanceof RuntimeException); // throw a specific type
assertEquals("message", expectThrow(client::invoke).getMessage()); // throw a specific message
```

This is clearly simpler. It also gets you out of having to copy the `Predicate<T>` interface (although you still want `ThrowingRunnable` for ergonomic reasons). I like it.

kcooney:
We could also allow the caller to pass in the type of the expected exception:

    IllegalArguementException e = expectThrows(
        IllegalArguementException.class, client::invoke);
    assertThat(e.getMessage()).contains("boom");

rschmitt:
Well, maybe. There's already a bit of a Cartesian explosion going on with all the convenience methods (message or no message, Callable/Runnable/ThrowingRunnable; adding an optional Class makes it twelve methods). Additionally, a separate assertion on the expected exception type seems to do a better job of separating the "act" part of the test from the "assert" part.

kcooney:
@rschmitt why do we need to support Callable, Runnable and ThrowingRunnable? Why can't we just support ThrowingRunnable? Are the extra overloaded there for people that can't use JDK 8?

If we don't need to support Runnable or Callable, then we have four overloads: with and without a message, and with or without the expected exception type.

We would only have two overloads if we require the expected exception, which I think we should. Asserting that *any* exception is thrown should be rare.

While I do like splitting "act" vs "assert", any lambda based approach is doing both, because the method calls the lambda, and if there is no exception thrown by the lambda, the method should throw an `AssertionError`

rschmitt:
@kcooney I'll need to review the details on this, but the basic idea is that Java does not reliably infer a void SAM type from a reference to a non-void method. Moving the type assertions out doesn't completely decouple execution from assertion, but it provides you the greatest degree of decoupling possible, given the fact that throwing exceptions is inextricably linked to execution flow.

rschmitt:
It turns out you're right about `ThrowingRunnable`. If that is the only type offered, we can support all four cases without wrapping:

* void methods that don't throw
* non-void methods that don't throw
* void methods that throw
* non-void methods that don't throw

The result otherwise is either type errors related to checked exceptions or an ambiguous method call.

rschmitt:
Oh, and the other possibility is `cannot convert void to java.lang.Object`. Java will discard a return value to make SAM type inference work out, but it won't make one up.

rschmitt:
New proposal is at #1154

Issue #706, opened 2013-07-20
New Method Proposal: ExpectedException#expect(Throwable, Callable)


Tibor17:
Many times I needed to assert a block of code in test-method to throw an exception, and check how the throwing block has changed values.
I had to use try-catch.

So I would like to implement new method with such of signature `ExpectedException#expect(Throwable, Callable)`.

Usage:

`Object obj = expect( Exception.class ,  new Callable() { ..do and return something.. });`

Usage of Lambda in Java 8:

`expect( Exception.class ,  () -> { ..do and return something.. }  )`

kcooney:
Fixed in #1154

PR #1154, opened 2015-06-04
Add Assert#expectThrows

rschmitt:
Continued from #1153. This is dramatically simpler than the first draft.

### DIFF
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -1042,7 +1042,8 @@ public class Assert {
             }
         }
         if (message == null) {
-            message = String.format("Expected %s to be thrown, but nothing was thrown", throwableClass.getSimpleName());
+            message = String.format("Expected %s to be thrown, but nothing was thrown",
+                    throwableClass.getSimpleName());
         }
         throw new AssertionError(message);
     }
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -957,19 +957,17 @@ public class Assert {
     }
 
     /**
-     * This interface facilitates the use of expectThrows from Java 8. It allows
-     * method references to void methods (that declare checked exceptions) to be
-     * passed directly into expectThrows without wrapping. It is not meant to be
-     * implemented directly.
+     * This interface facilitates the use of expectThrows from Java 8. It allows method references
+     * to void methods (that declare checked exceptions) to be passed directly into expectThrows
+     * without wrapping. It is not meant to be implemented directly.
      */
     public interface ThrowingRunnable {
         void run() throws Throwable;
     }
 
     /**
-     * Asserts that {@code runnable} throws an exception when executed. If it
-     * does, the exception object is returned. If it does not, an
-     * {@link AssertionError} is thrown.
+     * Asserts that {@code runnable} throws an exception when executed. If it does, the exception
+     * object is returned. If it does not, an {@link AssertionError} is thrown.
      *
      * @param runnable A function that is expected to throw an exception when executed
      * @return The exception thrown by {@code runnable}
@@ -979,11 +977,11 @@ public class Assert {
     }
 
     /**
-     * Asserts that {@code runnable} throws an exception when executed. If it
-     * does, the exception object is returned. If it does not, an
-     * {@link AssertionError} is thrown with the given {@code message}.
+     * Asserts that {@code runnable} throws an exception when executed. If it does, the exception
+     * object is returned. If it does not, an {@link AssertionError} is thrown with the given {@code
+     * message}.
      *
-     * @param message the identifying message for the {@link AssertionError}
+     * @param message  the identifying message for the {@link AssertionError}
      * @param runnable A function that is expected to throw an exception when executed
      * @return The exception thrown by {@code runnable}
      */
@@ -992,11 +990,10 @@ public class Assert {
     }
 
     /**
-     * Asserts that {@code runnable} throws an exception of type
-     * {@code throwableClass} when executed. If it does, the exception object
-     * is returned. If it does not throw an exception, an {@link AssertionError}
-     * is thrown. If it throws the wrong type of exception, an
-     * {@code AssertionError} is thrown describing the mismatch.
+     * Asserts that {@code runnable} throws an exception of type {@code throwableClass} when
+     * executed. If it does, the exception object is returned. If it does not throw an exception, an
+     * {@link AssertionError} is thrown. If it throws the wrong type of exception, an {@code
+     * AssertionError} is thrown describing the mismatch.
      *
      * @param throwableClass the expected type of the exception
      * @param runnable       A function that is expected to throw an exception when executed
@@ -1007,13 +1004,12 @@ public class Assert {
     }
 
     /**
-     * Asserts that {@code runnable} throws an exception of type
-     * {@code throwableClass} when executed. If it does, the exception object
-     * is returned. If it does not throw an exception, an {@link AssertionError}
-     * is thrown with the given {@code message}. If it throws the wrong type
+     * Asserts that {@code runnable} throws an exception of type {@code throwableClass} when
+     * executed. If it does, the exception object is returned. If it does not throw an exception, an
+     * {@link AssertionError} is thrown with the given {@code message}. If it throws the wrong type
      * of exception, an {@code AssertionError} is thrown describing the mismatch.
      *
-     * @param message the identifying message for the {@link AssertionError}
+     * @param message        the identifying message for the {@link AssertionError}
      * @param throwableClass the expected type of the exception
      * @param runnable       A function that is expected to throw an exception when executed
      * @return The exception thrown by {@code runnable}
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -298,7 +298,7 @@ public class Assert {
     public static void assertArrayEquals(Object[] expecteds, Object[] actuals) {
         assertArrayEquals(null, expecteds, actuals);
     }
-    
+
     /**
      * Asserts that two boolean arrays are equal. If they are not, an
      * {@link AssertionError} is thrown with the given message. If
@@ -313,8 +313,8 @@ public class Assert {
     public static void assertArrayEquals(String message, boolean[] expecteds,
             boolean[] actuals) throws ArrayComparisonFailure {
         internalArrayEquals(message, expecteds, actuals);
-    }    
-    
+    }
+
     /**
      * Asserts that two boolean arrays are equal. If they are not, an
      * {@link AssertionError} is thrown. If <code>expected</code> and
@@ -960,6 +960,8 @@ public class Assert {
      * This interface facilitates the use of expectThrows from Java 8. It allows method references
      * to void methods (that declare checked exceptions) to be passed directly into expectThrows
      * without wrapping. It is not meant to be implemented directly.
+     *
+     * @since 4.13
      */
     public interface ThrowingRunnable {
         void run() throws Throwable;
@@ -971,6 +973,7 @@ public class Assert {
      *
      * @param runnable A function that is expected to throw an exception when executed
      * @return The exception thrown by {@code runnable}
+     * @since 4.13
      */
     public static <T extends Throwable> T expectThrows(ThrowingRunnable runnable) {
         return (T) expectThrows(null, Throwable.class, runnable);
@@ -984,6 +987,7 @@ public class Assert {
      * @param message  the identifying message for the {@link AssertionError}
      * @param runnable A function that is expected to throw an exception when executed
      * @return The exception thrown by {@code runnable}
+     * @since 4.13
      */
     public static <T extends Throwable> T expectThrows(String message, ThrowingRunnable runnable) {
         return (T) expectThrows(message, Throwable.class, runnable);
@@ -998,6 +1002,7 @@ public class Assert {
      * @param throwableClass the expected type of the exception
      * @param runnable       A function that is expected to throw an exception when executed
      * @return The exception thrown by {@code runnable}
+     * @since 4.13
      */
     public static <T extends Throwable> T expectThrows(Class<T> throwableClass, ThrowingRunnable runnable) {
         return expectThrows(null, throwableClass, runnable);
@@ -1013,6 +1018,7 @@ public class Assert {
      * @param throwableClass the expected type of the exception
      * @param runnable       A function that is expected to throw an exception when executed
      * @return The exception thrown by {@code runnable}
+     * @since 4.13
      */
     public static <T extends Throwable> T expectThrows(String message, Class<T> throwableClass, ThrowingRunnable runnable) {
         try {
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -1041,6 +1041,9 @@ public class Assert {
                 throw assertionError;
             }
         }
+        if (message == null) {
+            message = String.format("Expected %s to be thrown, but nothing was thrown.", throwableClass.getSimpleName());
+        }
         fail(message);
         throw new AssertionError(); // This statement is unreachable.
     }
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -1042,7 +1042,7 @@ public class Assert {
             }
         }
         if (message == null) {
-            message = String.format("Expected %s to be thrown, but nothing was thrown.", throwableClass.getSimpleName());
+            message = String.format("Expected %s to be thrown, but nothing was thrown", throwableClass.getSimpleName());
         }
         fail(message);
         throw new AssertionError(); // This statement is unreachable.
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -975,6 +975,7 @@ public class Assert {
      * @return The exception thrown by {@code runnable}
      * @since 4.13
      */
+    @SuppressWarnings("unchecked")
     public static <T extends Throwable> T expectThrows(ThrowingRunnable runnable) {
         return (T) expectThrows(null, Throwable.class, runnable);
     }
@@ -989,6 +990,7 @@ public class Assert {
      * @return The exception thrown by {@code runnable}
      * @since 4.13
      */
+    @SuppressWarnings("unchecked")
     public static <T extends Throwable> T expectThrows(String message, ThrowingRunnable runnable) {
         return (T) expectThrows(message, Throwable.class, runnable);
     }
@@ -997,7 +999,8 @@ public class Assert {
      * Asserts that {@code runnable} throws an exception of type {@code throwableClass} when
      * executed. If it does, the exception object is returned. If it does not throw an exception, an
      * {@link AssertionError} is thrown. If it throws the wrong type of exception, an {@code
-     * AssertionError} is thrown describing the mismatch.
+     * AssertionError} is thrown describing the mismatch; the exception that was actually thrown can
+     * be obtained by calling {@link AssertionError#getCause}.
      *
      * @param throwableClass the expected type of the exception
      * @param runnable       A function that is expected to throw an exception when executed
@@ -1012,7 +1015,8 @@ public class Assert {
      * Asserts that {@code runnable} throws an exception of type {@code throwableClass} when
      * executed. If it does, the exception object is returned. If it does not throw an exception, an
      * {@link AssertionError} is thrown with the given {@code message}. If it throws the wrong type
-     * of exception, an {@code AssertionError} is thrown describing the mismatch.
+     * of exception, an {@code AssertionError} is thrown describing the mismatch; the exception that
+     * was actually thrown can be obtained by calling {@link AssertionError#getCause}.
      *
      * @param message        the identifying message for the {@link AssertionError}
      * @param throwableClass the expected type of the exception
@@ -1020,14 +1024,22 @@ public class Assert {
      * @return The exception thrown by {@code runnable}
      * @since 4.13
      */
+    @SuppressWarnings("unchecked")
     public static <T extends Throwable> T expectThrows(String message, Class<T> throwableClass, ThrowingRunnable runnable) {
         try {
             runnable.run();
         } catch (Throwable t) {
-            String mismatchMessage = String.format("Expected %s to be thrown, but got %s instead",
-                    throwableClass.getSimpleName(), t.getClass().getSimpleName());
-            assertTrue(mismatchMessage, throwableClass.isInstance(t));
-            return (T) t;
+            try {
+                return throwableClass.cast(t);
+            } catch (ClassCastException ex) {
+                String mismatchMessage = String.format("Expected %s to be thrown, but %s was thrown",
+                        throwableClass.getSimpleName(), t.getClass().getSimpleName());
+
+                // The AssertionError(String, Throwable) ctor is only available on JDK7.
+                AssertionError assertionError = new AssertionError(mismatchMessage);
+                assertionError.initCause(t);
+                throw assertionError;
+            }
         }
         fail(message);
         throw new AssertionError(); // This statement is unreachable.
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -968,50 +968,51 @@ public class Assert {
     }
 
     /**
-     * Asserts that {@code runnable} throws an exception of type {@code throwableClass} when
+     * Asserts that {@code runnable} throws an exception of type {@code expectedThrowable} when
      * executed. If it does not throw an exception, an {@link AssertionError} is thrown. If it
      * throws the wrong type of exception, an {@code AssertionError} is thrown describing the
      * mismatch; the exception that was actually thrown can be obtained by calling {@link
      * AssertionError#getCause}.
      *
-     * @param throwableClass the expected type of the exception
+     * @param expectedThrowable the expected type of the exception
      * @param runnable       a function that is expected to throw an exception when executed
      * @since 4.13
      */
-    public static <T extends Throwable> void assertThrows(Class<T> throwableClass, ThrowingRunnable runnable) {
-        expectThrows(throwableClass, runnable);
+    public static void assertThrows(Class<? extends Throwable> expectedThrowable, ThrowingRunnable runnable) {
+        expectThrows(expectedThrowable, runnable);
     }
 
     /**
-     * Asserts that {@code runnable} throws an exception of type {@code throwableClass} when
+     * Asserts that {@code runnable} throws an exception of type {@code expectedThrowable} when
      * executed. If it does, the exception object is returned. If it does not throw an exception, an
      * {@link AssertionError} is thrown. If it throws the wrong type of exception, an {@code
      * AssertionError} is thrown describing the mismatch; the exception that was actually thrown can
      * be obtained by calling {@link AssertionError#getCause}.
      *
-     * @param throwableClass the expected type of the exception
+     * @param expectedThrowable the expected type of the exception
      * @param runnable       a function that is expected to throw an exception when executed
      * @return the exception thrown by {@code runnable}
      * @since 4.13
      */
-    public static <T extends Throwable> T expectThrows(Class<T> throwableClass, ThrowingRunnable runnable) {
+    public static <T extends Throwable> T expectThrows(Class<T> expectedThrowable, ThrowingRunnable runnable) {
         try {
             runnable.run();
-        } catch (Throwable t) {
-            if (throwableClass.isInstance(t)) {
-                return throwableClass.cast(t);
+        } catch (Throwable actualThrown) {
+            if (expectedThrowable.isInstance(actualThrown)) {
+                @SuppressWarnings("unchecked") T retVal = (T) actualThrown;
+                return retVal;
             } else {
-                String mismatchMessage = String.format("Expected %s to be thrown, but %s was thrown",
-                        throwableClass.getSimpleName(), t.getClass().getSimpleName());
+                String mismatchMessage = format("unexpected exception type thrown;",
+                        expectedThrowable.getSimpleName(), actualThrown.getClass().getSimpleName());
 
                 // The AssertionError(String, Throwable) ctor is only available on JDK7.
                 AssertionError assertionError = new AssertionError(mismatchMessage);
-                assertionError.initCause(t);
+                assertionError.initCause(actualThrown);
                 throw assertionError;
             }
         }
-        String message = String.format("Expected %s to be thrown, but nothing was thrown",
-                    throwableClass.getSimpleName());
+        String message = String.format("expected %s to be thrown, but nothing was thrown",
+                expectedThrowable.getSimpleName());
         throw new AssertionError(message);
     }
 }
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -969,14 +969,29 @@ public class Assert {
 
     /**
      * Asserts that {@code runnable} throws an exception of type {@code throwableClass} when
+     * executed. If it does not throw an exception, an {@link AssertionError} is thrown. If it
+     * throws the wrong type of exception, an {@code AssertionError} is thrown describing the
+     * mismatch; the exception that was actually thrown can be obtained by calling {@link
+     * AssertionError#getCause}.
+     *
+     * @param throwableClass the expected type of the exception
+     * @param runnable       a function that is expected to throw an exception when executed
+     * @since 4.13
+     */
+    public static <T extends Throwable> void assertThrows(Class<T> throwableClass, ThrowingRunnable runnable) {
+        expectThrows(throwableClass, runnable);
+    }
+
+    /**
+     * Asserts that {@code runnable} throws an exception of type {@code throwableClass} when
      * executed. If it does, the exception object is returned. If it does not throw an exception, an
      * {@link AssertionError} is thrown. If it throws the wrong type of exception, an {@code
      * AssertionError} is thrown describing the mismatch; the exception that was actually thrown can
      * be obtained by calling {@link AssertionError#getCause}.
      *
      * @param throwableClass the expected type of the exception
-     * @param runnable       A function that is expected to throw an exception when executed
-     * @return The exception thrown by {@code runnable}
+     * @param runnable       a function that is expected to throw an exception when executed
+     * @return the exception thrown by {@code runnable}
      * @since 4.13
      */
     public static <T extends Throwable> T expectThrows(Class<T> throwableClass, ThrowingRunnable runnable) {
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -968,34 +968,6 @@ public class Assert {
     }
 
     /**
-     * Asserts that {@code runnable} throws an exception when executed. If it does, the exception
-     * object is returned. If it does not, an {@link AssertionError} is thrown.
-     *
-     * @param runnable A function that is expected to throw an exception when executed
-     * @return The exception thrown by {@code runnable}
-     * @since 4.13
-     */
-    @SuppressWarnings("unchecked")
-    public static <T extends Throwable> T expectThrows(ThrowingRunnable runnable) {
-        return (T) expectThrows(null, Throwable.class, runnable);
-    }
-
-    /**
-     * Asserts that {@code runnable} throws an exception when executed. If it does, the exception
-     * object is returned. If it does not, an {@link AssertionError} is thrown with the given {@code
-     * message}.
-     *
-     * @param message  the identifying message for the {@link AssertionError}
-     * @param runnable A function that is expected to throw an exception when executed
-     * @return The exception thrown by {@code runnable}
-     * @since 4.13
-     */
-    @SuppressWarnings("unchecked")
-    public static <T extends Throwable> T expectThrows(String message, ThrowingRunnable runnable) {
-        return (T) expectThrows(message, Throwable.class, runnable);
-    }
-
-    /**
      * Asserts that {@code runnable} throws an exception of type {@code throwableClass} when
      * executed. If it does, the exception object is returned. If it does not throw an exception, an
      * {@link AssertionError} is thrown. If it throws the wrong type of exception, an {@code
@@ -1008,24 +980,6 @@ public class Assert {
      * @since 4.13
      */
     public static <T extends Throwable> T expectThrows(Class<T> throwableClass, ThrowingRunnable runnable) {
-        return expectThrows(null, throwableClass, runnable);
-    }
-
-    /**
-     * Asserts that {@code runnable} throws an exception of type {@code throwableClass} when
-     * executed. If it does, the exception object is returned. If it does not throw an exception, an
-     * {@link AssertionError} is thrown with the given {@code message}. If it throws the wrong type
-     * of exception, an {@code AssertionError} is thrown describing the mismatch; the exception that
-     * was actually thrown can be obtained by calling {@link AssertionError#getCause}.
-     *
-     * @param message        the identifying message for the {@link AssertionError}
-     * @param throwableClass the expected type of the exception
-     * @param runnable       A function that is expected to throw an exception when executed
-     * @return The exception thrown by {@code runnable}
-     * @since 4.13
-     */
-    @SuppressWarnings("unchecked")
-    public static <T extends Throwable> T expectThrows(String message, Class<T> throwableClass, ThrowingRunnable runnable) {
         try {
             runnable.run();
         } catch (Throwable t) {
@@ -1041,10 +995,8 @@ public class Assert {
                 throw assertionError;
             }
         }
-        if (message == null) {
-            message = String.format("Expected %s to be thrown, but nothing was thrown",
+        String message = String.format("Expected %s to be thrown, but nothing was thrown",
                     throwableClass.getSimpleName());
-        }
         throw new AssertionError(message);
     }
 }
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -974,25 +974,58 @@ public class Assert {
      * @param runnable A function that is expected to throw an exception when executed
      * @return The exception thrown by {@code runnable}
      */
-    public static Throwable expectThrows(ThrowingRunnable runnable) {
-        return expectThrows(null, runnable);
+    public static <T extends Throwable> T expectThrows(ThrowingRunnable runnable) {
+        return (T) expectThrows(null, Throwable.class, runnable);
     }
-
 
     /**
      * Asserts that {@code runnable} throws an exception when executed. If it
      * does, the exception object is returned. If it does not, an
-     * {@link AssertionError} is thrown with the given message.
+     * {@link AssertionError} is thrown with the given {@code message}.
      *
      * @param message the identifying message for the {@link AssertionError}
      * @param runnable A function that is expected to throw an exception when executed
      * @return The exception thrown by {@code runnable}
      */
-    public static Throwable expectThrows(String message, ThrowingRunnable runnable) {
+    public static <T extends Throwable> T expectThrows(String message, ThrowingRunnable runnable) {
+        return (T) expectThrows(message, Throwable.class, runnable);
+    }
+
+    /**
+     * Asserts that {@code runnable} throws an exception of type
+     * {@code throwableClass} when executed. If it does, the exception object
+     * is returned. If it does not throw an exception, an {@link AssertionError}
+     * is thrown. If it throws the wrong type of exception, an
+     * {@code AssertionError} is thrown describing the mismatch.
+     *
+     * @param throwableClass the expected type of the exception
+     * @param runnable       A function that is expected to throw an exception when executed
+     * @return The exception thrown by {@code runnable}
+     */
+    public static <T extends Throwable> T expectThrows(Class<T> throwableClass, ThrowingRunnable runnable) {
+        return expectThrows(null, throwableClass, runnable);
+    }
+
+    /**
+     * Asserts that {@code runnable} throws an exception of type
+     * {@code throwableClass} when executed. If it does, the exception object
+     * is returned. If it does not throw an exception, an {@link AssertionError}
+     * is thrown with the given {@code message}. If it throws the wrong type
+     * of exception, an {@code AssertionError} is thrown describing the mismatch.
+     *
+     * @param message the identifying message for the {@link AssertionError}
+     * @param throwableClass the expected type of the exception
+     * @param runnable       A function that is expected to throw an exception when executed
+     * @return The exception thrown by {@code runnable}
+     */
+    public static <T extends Throwable> T expectThrows(String message, Class<T> throwableClass, ThrowingRunnable runnable) {
         try {
             runnable.run();
         } catch (Throwable t) {
-            return t;
+            String mismatchMessage = String.format("Expected %s to be thrown, but got %s instead",
+                    throwableClass.getSimpleName(), t.getClass().getSimpleName());
+            assertTrue(mismatchMessage, throwableClass.isInstance(t));
+            return (T) t;
         }
         fail(message);
         throw new AssertionError(); // This statement is unreachable.
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -1029,9 +1029,9 @@ public class Assert {
         try {
             runnable.run();
         } catch (Throwable t) {
-            try {
+            if (throwableClass.isInstance(t)) {
                 return throwableClass.cast(t);
-            } catch (ClassCastException ex) {
+            } else {
                 String mismatchMessage = String.format("Expected %s to be thrown, but %s was thrown",
                         throwableClass.getSimpleName(), t.getClass().getSimpleName());
 
@@ -1044,7 +1044,6 @@ public class Assert {
         if (message == null) {
             message = String.format("Expected %s to be thrown, but nothing was thrown", throwableClass.getSimpleName());
         }
-        fail(message);
-        throw new AssertionError(); // This statement is unreachable.
+        throw new AssertionError(message);
     }
 }
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -955,4 +955,46 @@ public class Assert {
             Matcher<? super T> matcher) {
         MatcherAssert.assertThat(reason, actual, matcher);
     }
+
+    /**
+     * This interface facilitates the use of expectThrows from Java 8. It allows
+     * method references to void methods (that declare checked exceptions) to be
+     * passed directly into expectThrows without wrapping. It is not meant to be
+     * implemented directly.
+     */
+    public interface ThrowingRunnable {
+        void run() throws Throwable;
+    }
+
+    /**
+     * Asserts that {@code runnable} throws an exception when executed. If it
+     * does, the exception object is returned. If it does not, an
+     * {@link AssertionError} is thrown.
+     *
+     * @param runnable A function that is expected to throw an exception when executed
+     * @return The exception thrown by {@code runnable}
+     */
+    public static Throwable expectThrows(ThrowingRunnable runnable) {
+        return expectThrows(null, runnable);
+    }
+
+
+    /**
+     * Asserts that {@code runnable} throws an exception when executed. If it
+     * does, the exception object is returned. If it does not, an
+     * {@link AssertionError} is thrown with the given message.
+     *
+     * @param message the identifying message for the {@link AssertionError}
+     * @param runnable A function that is expected to throw an exception when executed
+     * @return The exception thrown by {@code runnable}
+     */
+    public static Throwable expectThrows(String message, ThrowingRunnable runnable) {
+        try {
+            runnable.run();
+        } catch (Throwable t) {
+            return t;
+        }
+        fail(message);
+        throw new AssertionError(); // This statement is unreachable.
+    }
 }
--- a/src/test/java/org/junit/tests/assertion/AssertionTest.java
+++ b/src/test/java/org/junit/tests/assertion/AssertionTest.java
@@ -1,5 +1,14 @@
 package org.junit.tests.assertion;
 
+import org.junit.Assert;
+import org.junit.Assert.ThrowingRunnable;
+import org.junit.ComparisonFailure;
+import org.junit.Test;
+import org.junit.internal.ArrayComparisonFailure;
+
+import java.io.IOException;
+import java.math.BigDecimal;
+
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.hamcrest.CoreMatchers.is;
@@ -13,15 +22,6 @@ import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.expectThrows;
 import static org.junit.Assert.fail;
-
-import org.junit.Assert;
-import org.junit.Assert.ThrowingRunnable;
-import org.junit.ComparisonFailure;
-import org.junit.Test;
-import org.junit.internal.ArrayComparisonFailure;
-
-import java.io.IOException;
-import java.math.BigDecimal;
 
 /**
  * Tests for {@link org.junit.Assert}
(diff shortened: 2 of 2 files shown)


src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,46 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+
+    /**
+     * Asserts that {@code runnable} throws an exception when executed. If it
+     * does, the exception object is returned. If it does not, an
+     * {@link AssertionError} is thrown.
+     *
+     * @param runnable A function that is expected to throw an exception when executed
+     * @return The exception thrown by {@code runnable}
+     */
+    public static Throwable expectThrows(ThrowingRunnable runnable) {
```
kcooney: I strongly feel this should take a class object of the expected type, for all the reasons that I gave as #/1153

While Arrange-Act-Assert is a good framework to keep in mind when writing tests, it isn't the only lens we should use when we design APIs.

A common thing I think about when designing an API is "Make Interfaces Easy to Use Correctly and Hard to Use Incorrectly" (see http://programmer.97things.oreilly.com/wiki/index.php/Make_Interfaces_Easy_to_Use_Correctly_and_Hard_to_Use_Incorrectly).

The API here encourages code that provides less than useful error messages when the code under test doesn't meet the tester's expectations. For example:

    assertTrue(expectThrow(client::invoke) instanceof RuntimeException); // horrible message
    IllegalArgumentException e
        = (IllegalArgumentException) expectThrow(client::invoke); // not much better

Worse, the test could pass for the wrong reason

    expectThrow(client::invoke); // what the method starts throwing StackOverflowError ?

It also makes it difficult to do simple things:

    assertEquals(3, ((MyException) expectThrow(client::invoke)).getErrorCode());

Both of these are solved if we pass in the expected type:

    ClientException e = expectThrow(ClientException.class, client::invoke);
    assertThat(e.getMessage()).contains("connection is closed");

    assertEquals(3, expectThrow(MyException.class, client::invoke).getErrorCode());

rschmitt: There's a middle ground here that I actually like better. We can adjust the definition by genericizing the return type:

```java
    public static <T extends Throwable> T expectThrows(ThrowingRunnable runnable) {
        return expectThrows(null, runnable);
    }

    public static <T extends Throwable> T expectThrows(String message, ThrowingRunnable runnable) {
        try {
            runnable.run();
        } catch (Throwable t) {
            return (T) t;
        }
        Assert.fail(message);
        throw new AssertionError(); // This statement is unreachable.
    }
```

This permits the following:

```java
        // I don't care what gets thrown, and that's okay:
        expectThrows(() -> {throw new NullPointerException();});

        // No explicit cast takes place here:
        NullPointerException npe = expectThrows(() -> {throw new NullPointerException();});
        Assert.assertNull(npe.getMessage());

        // The following will throw a CCE at runtime:
        IOException ex = expectThrows(() -> {throw new NullPointerException();});
```

Your examples become:

```java
ClientException e = expectThrow(client::invoke);
assertThat(e.getMessage()).contains("connection is closed");

// This is still tricky without a cast--either way, an explicit type hint is required
assertEquals(3, ((ClientException) expectThrow(client::invoke)).getErrorCode());

ClientException e = expectThrow(client::invoke);
assertThat(e.getMessage()).contains("connection is closed");
assertEquals(3, e.getErrorCode());
```

kcooney: I don't think that addresses my concern about having a good error message if an unexpected exception is thrown.

I think being explicit provides a cleaner, more obvious and easier to use API.

rschmitt: See the latest diff.

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,75 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+    /**
+     * Asserts that {@code runnable} throws an exception when executed. If it does, the exception
+     * object is returned. If it does not, an {@link AssertionError} is thrown with the given {@code
+     * message}.
+     *
+     * @param message  the identifying message for the {@link AssertionError}
+     * @param runnable A function that is expected to throw an exception when executed
+     * @return The exception thrown by {@code runnable}
+     */
+    public static <T extends Throwable> T expectThrows(String message, ThrowingRunnable runnable) {
```
kcooney: Could you explain the use case for this? When would you know a call would throw but not know or care what exception is thrown?

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,75 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+
+    /**
+     * Asserts that {@code runnable} throws an exception of type {@code throwableClass} when
+     * executed. If it does, the exception object is returned. If it does not throw an exception, an
+     * {@link AssertionError} is thrown. If it throws the wrong type of exception, an {@code
+     * AssertionError} is thrown describing the mismatch.
+     *
+     * @param throwableClass the expected type of the exception
+     * @param runnable       A function that is expected to throw an exception when executed
+     * @return The exception thrown by {@code runnable}
```
kcooney: Please add `@since 4.13`

rschmitt: If 4.13 doesn't happen, will this be changed to `@since 5.0` by a regular expression or something? I think I saw this come up in another PR or source branch.

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,75 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+     * @param message        the identifying message for the {@link AssertionError}
+     * @param throwableClass the expected type of the exception
+     * @param runnable       A function that is expected to throw an exception when executed
+     * @return The exception thrown by {@code runnable}
+     */
+    public static <T extends Throwable> T expectThrows(String message, Class<T> throwableClass, ThrowingRunnable runnable) {
+        try {
+            runnable.run();
+        } catch (Throwable t) {
+            String mismatchMessage = String.format("Expected %s to be thrown, but got %s instead",
```
kcooney: You can avoid the formatting by calling `isInstance` true. It wouldn't be a large gain compared to the cost of the exception handling, but it's probably worth it.

Also, I suggest "expected %s to be thrown, but %s was thrown", and prepend "message" if it's not null, like we do in `Assert.format(String, Object, Object)`

src/test/java/org/junit/tests/assertion/AssertionTest.java
```diff
@@ -1,5 +1,14 @@
 package org.junit.tests.assertion;
 
+import org.junit.Assert;
```
kcooney: we put static imports before non-static ones. See
https://google-styleguide.googlecode.com/svn/trunk/javaguide.html

when in doubt, don't reorder imports.

src/test/java/org/junit/tests/assertion/AssertionTest.java
```diff
@@ -674,4 +677,80 @@ public void assertNotEqualsIgnoresDeltaOnNaN() {
...
+    @Test(expected = AssertionError.class)
+    public void expectThrowsRequiresAnExceptionToBeThrown() {
+        expectThrows(nonThrowingRunnable());
+    }
+
+    @Test
+    public void expectThrowsIncludesNoMessageByDefault() {
+        try {
+            expectThrows(nonThrowingRunnable());
+            fail();
```
kcooney: Note that `fail()` throws an `AssertionError`, so you need to move the `fail()` call outside of the `try` block and put a return after the `assertNull`

src/test/java/org/junit/tests/assertion/AssertionTest.java
```diff
@@ -674,4 +677,80 @@ public void assertNotEqualsIgnoresDeltaOnNaN() {
...
+    @Test
+    public void expectThrowsReturnsTheSameObjectThrown() {
+        NullPointerException npe = new NullPointerException();
+
+        Throwable throwable = expectThrows(throwingRunnable(npe));
+
+        assertSame(npe, throwable);
+    }
+
+    @Test(expected = ClassCastException.class)
```
kcooney: I personally think throwing `ClassCastException` isn't very useful. You have to dig through the code (and possibly the implementation of `expectThrows()`) to understand why this was thrown. If the caller instead had to pass the exception type, the caller would always get a clear exception message.

src/main/java/org/junit/Assert.java
kcooney: Yes, we will update all of the `@since 4.13` to `@since 5.0` if we decide to skip 4.13.

Note one of the other maintainers might request to just have this branch target the 5.0 branch.

kcooney:
Thanks for the Javadoc and the tests

src/main/java/org/junit/Assert.java
rschmitt: If I have a constructor that enforces preconditions on its arguments (e.g. no null references), I don't _really_ care what happens when bad arguments get passed into the constructor, as long as it doesn't return normally (thereby allowing a reference to an inconsistent object to escape). It doesn't matter if the constructor throws `IllegalArgumentException` or a `NullPointerException` or an `IndexOutOfBoundsException`. That's an irrelevant detail, and over-specifying it will lead to the kind of incidental complexity that Stuart Halloway warns about in the [Narcissistic Design](https://vimeo.com/77199361) talk (at about 15:10).

The caller doesn't need to specify `T`, it will be inferred by the compiler in the context of (for instance) an assignment statement. This is just what I showed in my earlier example:

```java
// No explicit cast takes place 
NullPointerException npe = expectThrows(() -> {throw new 
Assert.assertNull(npe.getMessage());
```

src/main/java/org/junit/Assert.java
kcooney: If someone feels strongly that they don't want to have to specify the expected exception type, they could do that outside of JUnit with a one-line method. Or we could add it in a future release. Let's start by requiring the caller explicitly, and see what people say when they start using the API.

src/test/java/org/junit/tests/assertion/AssertionTest.java
rschmitt: Now that I think about it, we should be more concerned about losing the exception that actually _was_ caught. That's the main piece of information you would need to debug a problem in the underlying SUT. (Note that fixing this is orthogonal to whether the expected exception type is specified/required.)

src/test/java/org/junit/tests/assertion/AssertionTest.java
rschmitt: Nuts, this is the second time I've screwed this up.

src/main/java/org/junit/Assert.java
rschmitt: I don't understand this remark. What about `isInstance`?

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,81 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+    /**
+     * Asserts that {@code runnable} throws an exception when executed. If it does, the exception
+     * object is returned. If it does not, an {@link AssertionError} is thrown.
+     *
+     * @param runnable A function that is expected to throw an exception when executed
+     * @return The exception thrown by {@code runnable}
+     * @since 4.13
+     */
+    public static <T extends Throwable> T expectThrows(ThrowingRunnable runnable) {
+        return (T) expectThrows(null, Throwable.class, runnable);
```
marcphilipp: This is an unchecked cast. If we keep this method and the one below IMHO it should rather return just `Throwable` without any generics.

src/main/java/org/junit/Assert.java
marcphilipp: Since this is only really usable in the presence of JDK 8 and lambda expressions, I believe this should end up in our planned Java 8 extensions project which is codenamed junit-lambda. We've already created the repo on GitHub but the project is not set up to accept pull requests yet.

src/main/java/org/junit/Assert.java
marcphilipp: Maybe this method and the one below should be called `expectThrowable`?

src/main/java/org/junit/Assert.java
kcooney: You don't care if the constructor throws a `StackOverflowError` or `ThreadDeath`?

If you don't care between `IllegalArgumentException` and `NullPointerException`, then expect a `RunTimeException`. If you find that to be too much typing, create your own version that doesn't require an exception.

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,81 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+    public static <T extends Throwable> T expectThrows(String message, Class<T> throwableClass, ThrowingRunnable runnable) {
+        try {
+            runnable.run();
+        } catch (Throwable t) {
+            String mismatchMessage = String.format("Expected %s to be thrown, but got %s instead",
+                    throwableClass.getSimpleName(), t.getClass().getSimpleName());
+            assertTrue(mismatchMessage, throwableClass.isInstance(t));
+            return (T) t;
+        }
+        fail(message);
```
marcphilipp: When I call `expectThrows(IllegalArgumentException.class, runnable)` but `runnable` does not throw any exception we will call `fail(null)` here, won't we? I think we need a better default message than that.

src/main/java/org/junit/Assert.java
kcooney: @marcphilipp I actually disagree we should target these changes to JUnitLambda, for two reasons. Firstly, we don't have a timeframe for the JDK 8 release. Secondly, we are deprecating Hamcrest APIs in 4.13/5.0, making `ExpectedException` much less useful. The API in this pull could be used by JUnit users that are already on JDK 8, without waiting for us to come up with other Lambda APIs.

src/test/java/org/junit/tests/assertion/AssertionTest.java
kcooney: Unfortunately, the exception that was caught could be a checked exception, so if an unexpected exception is caught, we would need to do wrapping at least for checked exceptions.

How about we if we catch an unexpected exception, we make that exception the cause of the `AssertionFailure` that we throw?

src/main/java/org/junit/Assert.java
rschmitt: Ok:

```java
        if (message == null) {
            message = String.format("Expected %s to be thrown, but nothing was thrown.", throwableClass.getSimpleName());
        }
```

src/main/java/org/junit/Assert.java
kcooney: Sigh. I really dislike how hard it is to follow comments on anything but the latest commit in github. :-(

My point was that `mismatchMessage` only needs to be constructed if there's a mismatch.

src/main/java/org/junit/Assert.java
rschmitt: The cast is checked at runtime, just like an explicit downcast from `Throwable`.

src/test/java/org/junit/tests/assertion/AssertionTest.java
rschmitt: Yes, I've already done that. You can throw an undeclared checked exception using a technique like Lombok uses to implement `sneakyThrow`, but wrapping in an `AssertionError` is the right thing to do if you want inline type assertions.

src/main/java/org/junit/Assert.java
rschmitt: First of all, you're not objecting to my API design or my testing practices, you're objecting to catching `Throwable`. And you can play similar games with `Exception` too: what are you, crazy? You can't catch `Exception`! What if that's an `InterruptedException` you just swallowed?! You didn't set the interrupted flag, you monster!

Catching `RuntimeException`, if it even works, is just a different kind of hack. In your case, it's not what you actually want to catch, it just happens to be the "most recent common ancestor" of the exceptions that you're actually interested in, but since there's no way to specify a type union in Java you specify an ancestor class instead. I could just as easily respond by saying: don't you care if your constructor throws `ArithmeticException`? You weren't actually expecting the constructor to divide by zero, were you? Now the test is passing for the wrong reason!

Furthermore: no, I don't care if my constructor throws `ThreadDeath` or `StackOverflowError`, because I know apodictically and with absolute metaphysical certitude that no constructor _I_ wrote is going to throw a `VirtualMachineError` during a unit test, unless the entire JVM is coincidentally destabilized at the exact same time for an unrelated reason.

Finally, even if you're very very precise in the types that you throw and catch, and you assert that (say) `NullPointerException` and _only_ `NullPointerException` is caught, you _still_ haven't necessarily demonstrated that you're making correct assertions. You might be catching an NPE thrown by a different (e.g. prior) statement, or one that was propagated from code the SUT called, and then your code is passing for the wrong reason. So at this point we have a few options:
* Associate a unique public exception type with each `throws` statement (which still doesn't work with recursion, maybe we need to add global locking or some sort of `Semaphore`)
* Require the user to specify an assertion containing both the exception type _and_ the message that that exception is to contain, and then I guess publish some sort of Checkstyle config that ensures at build time that all such messages are unique (I guess this brings us to those opaque Microsoftian "error codes")
* Be reasonable and concede that "this call does not return via normal control flow" is a coherent and intelligible assertion that can stand alone in a unit test

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,96 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+
+    /**
+     * Asserts that {@code runnable} throws an exception when executed. If it does, the exception
+     * object is returned. If it does not, an {@link AssertionError} is thrown.
+     *
+     * @param runnable A function that is expected to throw an exception when executed
+     * @return The exception thrown by {@code runnable}
+     * @since 4.13
+     */
+    @SuppressWarnings("unchecked")
```
marcphilipp: Sorry, I'm still convinced that we should rather return Throwable instead of using generics in this way.

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,96 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+                AssertionError assertionError = new AssertionError(mismatchMessage);
+                assertionError.initCause(t);
+                throw assertionError;
+            }
+        }
+        if (message == null) {
+            message = String.format("Expected %s to be thrown, but nothing was thrown", throwableClass.getSimpleName());
+        }
+        fail(message);
+        throw new AssertionError(); // This statement is unreachable.
```
marcphilipp: the above two lines can be replaced by 

```java
throw new AssertionError(message);
```

marcphilipp:
@rschmitt Thanks for your pull request, I like it!

I think this needs to go into junit-lambda, though. That would allow us to use JDK 8 interfaces like `Supplier` or `Predicate`. I would really like to use a `Supplier<String>` for the user-supplied message and maybe even a `Predicate<? super Throwable>` for the exception check so one could easily replace `instanceof` with exact-class semantic.

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,96 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+     * @return The exception thrown by {@code runnable}
+     * @since 4.13
+     */
+    @SuppressWarnings("unchecked")
+    public static <T extends Throwable> T expectThrows(String message, Class<T> throwableClass, ThrowingRunnable runnable) {
+        try {
+            runnable.run();
+        } catch (Throwable t) {
+            try {
+                return throwableClass.cast(t);
```
marcphilipp: Why have you replaced the call of `isInstance` with `cast`?

kcooney:
Although I understand your argument about not caring about the exception type, very few people share that view. Those that do can pass in Throwable.class or make their own overload.

I also share Marc's concerns about the casting, and the return type being  implicitlly referred (APIs that do often don't do what you want when you want).

Finally, it's extremely difficult to remove a method from JUnit once it's added. We can add it later. YAGNI.

I'm sorry, but I still strongly feel that the exception type should be explicitly specified, just like we do in ExpectedException and `@Test`, and just like you do in the tests you are adding.

That all said, I think we are at the point where we should ask the other maintainers for their view. @dsaff @marcphilipp and @stefanbirkner can you comment on these proposed APIs?

kcooney:
@marcphilipp I'm intrigued by your idea of using `Supplier` or `Predicate` in an exception API. Do you think you'd have time to put together a pull request against JUnitLambda with a sketch of an API and some example usages?

kcooney:
@rschmitt apologies if I sound argumentative; it isn't my intention. We're all very passionate about APIs and testing, so naturally we'll have disagreements. As you can see, the JUnit maintainers don't always agree with each other

src/main/java/org/junit/Assert.java
dsaff: I'm just getting up to speed with this pull request, but I also am a little unsure about this, since it appears to provide a way to write code that is functionally equivalent to one with an explicit cast, just with the cast hidden.

src/main/java/org/junit/Assert.java
rschmitt: Ah. Well, that's true now, since message can't be null.

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,96 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
     }
+
+    /**
+     * This interface facilitates the use of expectThrows from Java 8. It allows method references
+     * to void methods (that declare checked exceptions) to be passed directly into expectThrows
+     * without wrapping. It is not meant to be implemented directly.
+     *
+     * @since 4.13
+     */
+    public interface ThrowingRunnable {
```
Tibor17: why the junit has still the same tendency to use nested interfaces and nested annotations like Parameter.
It's really awful code like Assert.ThrowingRunnable and using static import for Assert.* and much easier to use and import ThrowingRunnable. Bad example is Parameterized.Parameter.
The only reason why nested classes is that they are not static and share one instance; otherwise it's useless.

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,96 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+     * @since 4.13
+     */
+    @SuppressWarnings("unchecked")
+    public static <T extends Throwable> T expectThrows(String message, Class<T> throwableClass, ThrowingRunnable runnable) {
+        try {
+            runnable.run();
+        } catch (Throwable t) {
+            try {
+                return throwableClass.cast(t);
+            } catch (ClassCastException ex) {
```
Tibor17: ex? Very usual is e.

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,96 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+                        throwableClass.getSimpleName(), t.getClass().getSimpleName());
+
+                // The AssertionError(String, Throwable) ctor is only available on JDK7.
+                AssertionError assertionError = new AssertionError(mismatchMessage);
+                assertionError.initCause(t);
+                throw assertionError;
+            }
+        }
+        if (message == null) {
+            message = String.format("Expected %s to be thrown, but nothing was thrown", throwableClass.getSimpleName());
```
Tibor17: too wide line

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,96 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+    /**
+     * Asserts that {@code runnable} throws an exception when executed. If it does, the exception
+     * object is returned. If it does not, an {@link AssertionError} is thrown.
+     *
+     * @param runnable A function that is expected to throw an exception when executed
+     * @return The exception thrown by {@code runnable}
+     * @since 4.13
+     */
+    @SuppressWarnings("unchecked")
+    public static <T extends Throwable> T expectThrows(ThrowingRunnable runnable) {
```
Tibor17: Choose 1 . with Generics and then ThrowingRunnable<T> and without cast, 
or without Generics .

src/main/java/org/junit/Assert.java
rschmitt: You don't need JDK8 to operate with functional interfaces like `Supplier<T>` and `Predicate<T>`. Any backport of any subset or superset of those interfaces is basically equivalent. My first draft of this change used a predicate instead of returning a throwable at all. The _only_ major feature that actually requires JDK8 (from the perspective of a library author) is `default` methods on interfaces. However, all of interfaces you've already defined ship with JUnit itself, and cannot be redefined in junit-lambda. So unless you're planning on building something substantial on a lesser-known JDK8 feature like JSR308 type annotations, I'm not sure what the point of junit-lambda is. The JDK8 backwards compatibility story is very, very solid. You're not _supposed_ to need to vend an extension library.

src/main/java/org/junit/Assert.java
rschmitt: There's no reason you need a star import. You can just static import `Assert.ThrowingRunnable`. Whether or not this interface should be extracted into a top-level file is up to the maintainers.

src/test/java/org/junit/tests/assertion/AssertionTest.java
```diff
@@ -674,4 +677,98 @@ public void assertNotEqualsIgnoresDeltaOnNaN() {
...
+    @Test
+    public void expectThrowsWrapsAndPropagatesUnexpectedExceptions() {
+        NullPointerException npe = new NullPointerException("inner-message");
+
+        try {
+            expectThrows(IOException.class, throwingRunnable(npe));
+        } catch (AssertionError ex) {
+            assertSame(npe, ex.getCause());
+            assertEquals("inner-message", ex.getCause().getMessage());
+            return;
```
Tibor17: why return? The method fail() may go to try block right after the line with expectThrows...

src/main/java/org/junit/Assert.java
Tibor17: No I dont think so it's thier decision. This is opensource and code review. The reviewer can be anyone.
The argument is that nested classes share instance; otherwise they are useless. Nested annotations are really hard to use if you want to have nice and meaningful code.

rschmitt: You can comment all you want on matters of code organization, and in accordance with Parkinson's Law of Triviality, people commonly do. My point is that the maintainers (which is not a group I belong to) have the best understanding of what the relevant convention is (if any) for new code, and what conditions would be sufficient for breaking that convention.

src/test/java/org/junit/tests/assertion/AssertionTest.java
rschmitt: See https://github.com/junit-team/junit/pull/1154#discussion-diff-31694161

src/main/java/org/junit/Assert.java
kcooney: I agree with Marc and David here. If the return type here is inferred to be something other than `Throwable` then the caller is expecting a specific exception, so the caller should provide a way to specify what is expected so they get a better error message (and one that includes the passed-in message).

If the caller doesn't care what is being thrown, but wants to assert that "something" is thrown, then we should return a `Throwable` and the caller can assert on it or do a cast on it.

src/main/java/org/junit/Assert.java
kcooney: I think there are cases where static nested classes make sense as a kind of organization tool to avoid an explosion of unrelated top-level classes in a package (see, for example, `Map.Entrry`).

I personally haven't decided if `ThrowingRunnnable` should be a top-level class. If we think it will be used elsewhere (perhaps `ErrorCollector`) then it should.

src/main/java/org/junit/Assert.java
kcooney: I don't think we've decided formally on a line length restriction, but since we recently switched to the Google Java Style Guide, we should probably use 100

See https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.5-line-wrapping

src/main/java/org/junit/Assert.java
rschmitt: I think that `ThrowingRunnnable` is a bit of a special case because it's just a shim. It gives us a place to hang the specific functional type signature we want to accept. The interface itself is only intended to be implemented by the compiler; it's not _really_ part of an API or SPI.

src/main/java/org/junit/Assert.java
rschmitt: I'm okay with 100; 80 is draconian.

src/main/java/org/junit/Assert.java
kcooney: And yet extension developers will use it when they develop more complicated APIs that reuse our code.

rschmitt: Yeah, it seems like there's no way to "encapsulate" this interface. We won't be able to move, eliminate, or rename it later.

kcooney:
One thing we should think about is whether we would want a similar API for `ErrorCollector`, and what it would look like. It would almost certainly *not* return the type of the exception being thrown, but could allow the caller to supply a `Consumer<T>` that would make assertions on the exception.

I'm not sure I would want to force code using `Assert.expectThrows()` to provide a `Consumer<T>` as well, but it would be nice if someone who understands  `Assert.expectThrows()` could reuse that knowledge when they start using `ErrorCollector`

We *may* also want `Assume.assumeThrows()`, which would also presumably not have a return value. I'm not convinced that it would be a common use case, though.

Keep the comments comming!

rschmitt:
Looking at `ErrorCollector`, it seems like you would want one of `Consumer`, `Predicate`, or `Function` (that is, backported equivalents). I'm not sure what you mean by "similar API."

`Assume#assumeThrows` is a bit of a more interesting case, because (since it returns `void`) you almost certainly _would_ want to take a `Predicate<T>` in the most general form of the method.

marcphilipp:
@kcooney I've conducted a few experiments using a `Supplier` and a `BiPredicate`. But I'm not convinced that this is the way to go. Feel free to [take a look](https://github.com/junit-team/junit-lambda/compare/junit-team:master...marcphilipp:expectThrows).

marcphilipp:
@rschmitt I really like this simpler version!

marcphilipp:
@junit-team/junit-committers We need to decide whether this belongs into "junit-core" or "junit-lambda".

I don't really dig the backport idea. What benefits would using such a backport have?

If I'm stuck with JDK 5/6/7, I could use new API that makes use of backported functional interfaces. However, the Javadoc for `ThrowingRunnable` states that I'm not supposed to implement it directly. If I would, it looked ugly:

```java
SomeException exception = expectThrows(SomeException.class, new ThrowingRunnable() {
    @Override
    public void run() throws Throwable {
        ...
    }
});
```

If I'm a lucky person who gets to use JDK8 in real-life projects already, I now have `some.backport.Function` along with `java.util.Function` in my classpath (and so on for all functional interfaces). If I'm testing code that makes use of the real functional interfaces, I will have a hard time managing imports or writing fully qualified names. I couldn't use existing consumers, suppliers, etc. (e.g. from the JDK API) with new JUnit API that uses back ported equivalents.

On the other side, we could move features that are designed to be used with JDK8 into a separate artifact, i.e. junit-lambda. That would allow interested users to quickly discover those new features. We could make use of Java 8 to implement the new features.  Existing consumers, suppliers, etc. could be used with JUnit API directly.

What do you guys think?

rschmitt:
`Supplier<String>` strikes me as extremely cumbersome for such a niche use case. Why not a lazily evaluated `CharSequence`?

marcphilipp:
Because that's what suppliers are for... e.g. java.util.logging uses them for the exact same purpose. IMHO the API would be much more clear if the parameter type is `Supplier<String>` rather than `CharSequence`.

But, since we have decided against a message for now, we don't need to discuss this right now. ;-)

rschmitt:
> We need to decide whether this belongs into "junit-core" or "junit-lambda".

There is not a single way in which this change (as of the current diff, namely d6652ad4ecd95cae8bed2a00da7f396197745ef8) would benefit from a dependency on JDK8. There is no standard library type that obviates `ThrowingRunnable`, not even in JDK8.

> If I'm a lucky person who gets to use JDK8 in real-life projects already, I now have some.backport.Function along with java.util.Function in my classpath (and so on for all functional interfaces). If I'm testing code that makes use of the real functional interfaces, I will have a hard time managing imports or writing fully qualified names.

It's rare for a call site to actually refer to a type like `Function` at all, even by importing it. Because lambda type inference is structural and not nominal (at the call site), it doesn't matter which specific `Function` type a given method accepts. It can be a JDK8 `java.util.function.Function` or a Guava `com.google.common.base.Function` or a proprietary JUnit type called `JUnitFunction`; the caller is just going to supply the same lambda or method reference.

> I couldn't use existing consumers, suppliers, etc. (e.g. from the JDK API) with new JUnit API that uses back ported equivalents.

If I have a `Supplier<T>` (nominally typed, in a local variable or method parameter) and you take a `JUnitSupplier<T>`, instead of writing:

`accept(supplier);`

I have to write:

`accept(supplier::get);`

This is hardly devastating. And remember, this is the 20% case. The 80% case won't involve an explicit intermediate function type at all.

> On the other side, we could move features that are designed to be used with JDK8 into a separate artifact, i.e. junit-lambda. That would allow interested users to quickly discover those new features. We could make use of Java 8 to implement the new features. Existing consumers, suppliers, etc. could be used with JUnit API directly.

Oracle put an enormous amount of effort into ensuring that twenty years' worth of non-lambda-aware code can be used with lambda expressions. This was one of the main motivations behind the choice of SAM type inference as the implementation strategy, instead of first-class function types like in Scala. If you break up your library into "the old way of doing things" and "the new way of doing things," you're basically doing Python 3. It's the opposite of how Java 8 migration is intended to work.

rschmitt:
> Because that's what suppliers are for... e.g. java.util.logging uses them for the exact same purpose. IMHO the API would be much more clear if the parameter type is Supplier<String> rather than CharSequence.
>
> But, since we have decided against a message for now, we don't need to discuss this right now. ;-)

Agreed on that last point, but remember: if you offer `Supplier<String>` _and_ `String` variants of every relevant method, you're doubling the surface area of your API. This means even more overloads for users to shovel through, as well as more boilerplate for maintainers and contributors to write, all for a feature that--I suspect--the vast majority of users have no interest in. A `CharSequence`, on the other hand, can be a `LazyString` (some sort of lazy wrapper around a `Supplier<String>`) _or_ a `String`.

kcooney:
Sorry this is so long, but there are many related threads above.

**Concerning using JDK8 APIs in JUnit:** I would prefer not to create JUnit4 versions of functional interfaces that exist in JDK 8. As @rschmitt pointed out, it isn't that big of a deal for users if we do, but I think as soon as we start wanting to use functional interfaces that exactly match the ones in JDK 8, that's the time to seriously consider pushing those features to JUnit-Lambda. If we use the built-in functional interfaces in JUnit, developers who already use JDK 8 will understand what those interfaces are for, and developers that are new to lambdas can search Google for the interfaces to understand how to use them and the possibilities they add to the API.

**Aside about JUnit-Lambda:** we don't want to require that users of JUnit 5 use JDK 8, because large number of users cannot use JDK 8 (including Android users and GWT users). But some contributors to JUnit want to start experimenting on what a Java test framework that uses lambdas would look like, and get a feel for what kinds of changes we would need in the core JUnit code to support those features. So we decided to create a different project for that. Yes, there are issues in terms of duplication of concepts and discoverability of the new APIs. Hopefully you all can help us resolve those issues when the JUnit-Lambda project is further along.

**Concerns about when to add this API:** The interface used in this pull doesn't have an equivalent JDK 8 version, so adding it does no harm. I don't think we should delay this feature until the release of JUnit-Lambda simply because the APIs aren't useful if you don't use JDK 8 (though I'm leaning towards moving `ThrowingRunnable` to a new package, `org.junit.function`).

The are reasons I would delay this pull include:
* We aren't sure this is the best API, and we want users to give us feedback after trying it
* We are planning on building a labmda-based and/or fluent assertion API in the near future
* We strongly feel these methods do not belong in the `Assert` class

My main concern is the first. Once we add an API to JUnit, it's hard to remove it, but it's possible we could have some Beta APIs JUnit-Lambda (possibly using an annotation like Guava's `@Beta` annotation)

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,48 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+     * @param throwableClass the expected type of the exception
+     * @param runnable       A function that is expected to throw an exception when executed
+     * @return The exception thrown by {@code runnable}
+     * @since 4.13
+     */
+    public static <T extends Throwable> T expectThrows(Class<T> throwableClass, ThrowingRunnable runnable) {
+        try {
+            runnable.run();
+        } catch (Throwable t) {
+            if (throwableClass.isInstance(t)) {
```
kcooney: this can throw a `NullPointerException` if 1throwableClass` is null. I recommend adding this to the top of the metod:

    if (throwableClass == null) {
      throw new NullPointerException("throwableClass"); // or some better message
    }

(yes, some might argue that `IllegalArgumentException` would be better, but JUnit has mostly settled on `NullPointerException` for new code (because that's what would usually be thrown if the check wasn't there), so we should be consistent.

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,48 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+                return throwableClass.cast(t);
```
kcooney: I would just do a simple cast here; you know it's safe because you just checked, and it's (slightly) more efficient and just as clear

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,48 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+            } else {
+                String mismatchMessage = String.format("Expected %s to be thrown, but %s was thrown",
```
kcooney: this isn't quite consistent with the other exception messages that we use for the other methods in this class. Instead, I think we should use `Assert.format()`:

    String mismatchMessage = format(
        "unexpected exception thrown", throwableClass.getSimpleName(), t.getClass().getSimpleName());

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,48 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+                        throwableClass.getSimpleName(), t.getClass().getSimpleName());
+
+                // The AssertionError(String, Throwable) ctor is only available on JDK7.
+                AssertionError assertionError = new AssertionError(mismatchMessage);
+                assertionError.initCause(t);
+                throw assertionError;
+            }
+        }
+        String message = String.format("Expected %s to be thrown, but nothing was thrown",
```
kcooney: nit: Please change "Expected" to "expected" to be consistent with the other messages for assertions thrown by this class

src/test/java/org/junit/tests/assertion/AssertionTest.java
```diff
@@ -674,4 +677,80 @@ public void assertNotEqualsIgnoresDeltaOnNaN() {
...
+        expectThrows(IOException.class, throwingRunnable(npe));
+    }
+
+    @Test
+    public void expectThrowsWrapsAndPropagatesUnexpectedExceptions() {
+        NullPointerException npe = new NullPointerException("inner-message");
+
+        try {
+            expectThrows(IOException.class, throwingRunnable(npe));
+        } catch (AssertionError ex) {
```
kcooney: BTW, one common convention is to name the exception "expected" to make it 100% clear that you expect it to be thrown. See https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s6.2-caught-exceptions

src/main/java/org/junit/Assert.java
rschmitt: > We aren't sure this is the best API, and we want users to give us feedback after trying it

You're never going to be _sure_ of that. But I'm not aware of any major objections, or even alternative proposals, to the latest revision. (JUnit-Lambda doesn't count because it's vaporware.)

> Once we add an API to JUnit, it's hard to remove it, but it's possible we could have some Beta APIs JUnit-Lambda (possibly using an annotation like Guava's @Beta annotation). I think it's a good API, but it's hard to know what problems it has or what improvements to add until some people start using it.

You guys make it sound like you've _never added a feature to JUnit_ and you wouldn't even know where to start. This is ridiculous. I'm adding one (1) method to `org.junit.Assert`, following simple and well-established patterns that can probably be traced back to SUnit. The thing has a surface area the size of a postage stamp, and I'm already pretty confident that we've explored virtually the entire design space already.

Furthermore, you can't just worry about the risk of releasing `expectThrows` unless you're also going to consider the risk of _not_ releasing `expectThrows`. I've already pointed out that I've seen people writing these assertions by hand, over and over again, _in situ_ in a test method. The verbosity of having to do this is annoying, but the high odds of getting it wrong actually make it dangerous, too. I've seen people of all experience levels get this wrong to varying degrees. The longer you drag your feet, the more broken tests will be written.

> We are planning on building a lambda-based and/or fluent replacement for Assert/Assume/ErrorCollector in the near future

So what? That's no reason for development on the existing `Assert` class to come to a standstill. It might be a relevant objection to a competing proposal for some sort of massive overhaul, but that's not what this PR is about.

src/main/java/org/junit/Assert.java
rschmitt: I'm starting to worry that you're stringing me along. My concern here is that the maintainers will continue asking for picayune tweaks and refactorings and generally wasting my time before informing me that they've decided that this PR will not be accepted in any form. Is there a consensus yet that this method, or something _very very similar to it_, should be added to `org.junit.Assert` _very very soon_?

kcooney: I don't think anyone is stringing you along. Unfortunately, we are talking about multiple things in the comments:

* what the API should look like
* how it should be implemented (including suggestions that some people might think border on Bikeshedding)
* when and how we should release it

Please feel free to postpone making minor changes until we agree on the other points. As you have probably seen by now, it's extremely hard to follow conversations in github when there are multiple rounds of comments and commits.

src/main/java/org/junit/Assert.java
kcooney: @rschmitt sorry if this has been frustrating.

Firstly, I apologize for not making it clear in my longer comment above that I think we should include an API like this in JUnit soon (and not wait until JUnit Lamba). While I understand the reasons why some people want to wait, I don't find the arguments presented so far convincing. I don't want to dismiss people's concerns out of hand, so I'm still leaving the door open for someone to convince me that we should wait.

Secondly, I ask that you please have patience in our process. Designing an API takes time. The size of the API is often unrelated to the amount of time it might take to ensure it's what we want to build, and make sure it's polished and well documented.

If we were developing an API that was going to be used by only a hundred people at one company, I'd be fine with what you proposed here, but by some estimates JUnit is used by millions of developers (based [this](https://www.google.com/search?q=how+many+java+developers+worldwide) and [this](http://zeroturnaround.com/rebellabs/java-tools-and-technologies-landscape-for-2014)). Taking a bit of time to ensure that this is the API that we want our developers to use is time worth spent.

Your original pull request was opened *only two days ago*, so I think it may be a bit of a stretch to say we are dragging our feet :-) 

Finally, I think you should take the comments from people as evidence of how enthusiastic people are to have an API like this as part of JUnit. We're simply trying to come to a consensus on what that API should look like, and when and how to release it.

rschmitt: > Your original pull request was opened only two days ago, and we responded on the same day, so I think it may be a bit of a stretch to say we are dragging our feet :-)

I don't think it is. There was [a nearly identical PR](https://github.com/junit-team/junit/pull/1055) _half a year ago_, and it went nowhere. And there was a [detailed proposal](https://github.com/junit-team/junit/issues/706) six months before _that_. The scary thing about #1055 is that it was closed because a strange consensus was reached that the functionality belonged in junit-lambda, at least initially. Six months later, junit-lambda has gone absolutely nowhere. But my point is that the maintainers have had over a year to meditate on this design. It would not be an act of reckless hastiness to merge this PR as-is.

> Secondly, I ask that you please have patience in our process.

That would be easier if there were a clear process. That's what I was trying to tell you when I responded to your remark that "[o]nce we add an API to JUnit, it's hard to remove it, but it's possible we could have some Beta APIs JUnit-Lambda (possibly using an annotation like Guava's @Beta annotation)." You're basically telling me that an open source project that has been around since the Clinton administration isn't sure how to _add a simple feature_. Why is this not a solved problem? Can you see how your remark can make the JUnit team look rudderless and stagnant?

Keep in mind that I've been very thorough in addressing feedback, either incorporating it or responding with (very solid, I think) counterarguments. Since I haven't heard any new feedback in a while, I'm left sitting here wondering what the problem is. What is everybody waiting for?

> Designing an API takes time. The size of the API is often unrelated to the amount of time it might take to ensure it's what we want to build, and make sure it's polished and well documented.

There's a difference between design and analysis paralysis. Seeing all of the old discarded PRs for this feature (which I only discovered recently) makes me worry that this discussion is tending towards the latter.

src/main/java/org/junit/Assert.java
rschmitt: My understanding is that (1) and (2) are largely settled, and the main disagreement around (3) has to do with junit-lambda (which I will continue to insist is a complete red herring). If you take junit-lambda out of the picture, I guess the main question is whether or not you mark the method `@Beta` (I would actually suggest `@Deprecated`) in the initial release.

dsaff: @rschmitt,

Can I ask a favor?  I'd like the chance to have a conversation with @marcphilipp and some others about how we should make choices given the upcoming plans for JUnit Lambda.  These people have invested considerable time away from their paid day jobs in order to look into these very questions, and I think it's fair to them to give them time to agree on a plan, and fair to you and other contributors to see a plan before you invest time into a particular enhancement.

Given that time to work on JUnit is, for all the maintainers, time that must be gleaned from outside of paid work and family, this conversation could take a while.  Can you give us two weeks?

If that's really too long, then you might want to try publishing this in a separate fork or repo for user feedback, because I'm afraid that JUnit has never been (nor needed to be) an organization that makes decisions much more quickly than that.

Feel free to drop me a note individually if more than two weeks goes by without an update here.

Fair enough?

rschmitt: That's fine; I have no problem with waiting, I just want to understand whose court the ball is in.

dsaff: OK, I ended up traveling for most of those two weeks, so just wanted you to know that I haven't forgotten, but will still need another week.  Sorry,

   David

rschmitt: After some experimentation on the TestNG side, I think we should have two methods:
```java
public static <T extends Throwable> T expectThrows(Class<T> throwableClass, ThrowingRunnable runnable);
public static <T extends Throwable> void assertThrows(Class<T> throwableClass, ThrowingRunnable runnable)
```
Here `assertThrows` delegates directly to `expectThrows` and discards its return value. I've identified three reasons why `assertThrows` should also be available:
1. It prevents warnings: discarding a Throwable return value can raise red flags
2. It facilitates convention: "assert" is called solely for its side effect, whereas "expect" returns something but first asserts something about its return value
3. It enhances discoverability: most people will look for methods prefixed with "assert", and indeed, the return value of expectThrows is often unneeded

Tibor17:
You can introduce one more method `expectThrows` with only the secind param and calling the method you already have but the first param will be Throwable.class. Then the assert will be 
`assertThat(expectThrows(() -> {...}), is(MyException.class))`. You should then have a test which proves generics with this construct using Java Hamcrest matcher.
Personally `expectedThrown` is fitting to the sentence when I read the statement with assertThat more than `expectThrows`:
`Assert that expected thrown is MyException.`

src/main/java/org/junit/Assert.java
marcphilipp: @rschmitt Thanks for the suggestion, I think that's the way to go! Do you have time to add it to this pull request?

After thinking this over for a couple of days/weeks I'm convinced we should get this shipped in 4.13. If we end up getting the design wrong, we'll deprecate it and give it another try. So, no more waiting for you... :smirk:

kcooney: @rschmitt I'm surprised IntelliJ would give you a warning about discarding the return value. Is the method you are calling annotated with `@CheckReturnValue `?

Can someone else with a recent version of IntelliJ IDEA verify that IntelliJ gives a warning if you call a method that returns a `<T extends Throwable>` without using the return value? It would be really unfortunate to have to have two methods that are identical except for the method name and whether there is a return value.

As for the method name, I slightly prefer `assertThrows()` vs. `checkThrows()`. The googletest C++ testing framework uses "expect" to mean "if this check values, cause the test to fail, but don't abort to test" (kind of like what `ErrorCollector` does). Also, `assertThrows` would be easier to discover.

rschmitt: @marcphilipp Sure, I'll add them to this PR.

@kcooney I'm pretty sure it's an IntelliJ inspection. We could also just suppress the warning on `expectThrows`, but I really like having `assertThrows` for discoverability and `expectThrows` to avoid muddying the waters of the current convention (currently all `assert` methods are void, right?). I'm not sure what the googletest convention has to do with anything. I can just as easily point at the [expect method](https://doc.rust-lang.org/std/option/enum.Option.html#method.expect) in Rust's Option type, which makes an assertion en passant and then returns a value.

kcooney: @rschmitt I simply mentioned googletest because it's a popular testing framework where 'check' has a different meaning than the one proposed here, and I was slightly worried about confusion. if we need to have two methods (and it sounds like we do) then it would be hard to avoid a name that already has a meaning, and the names you came up are good ones.

I'm onboard.

rschmitt: Wait, I'm confused. Are you talking about `check`, `assert`, or `expect`? You've mentioned `checkThrows` above, and there's no such method in the PR.

kcooney:
@Tibor17 I strongly feel that the caller should pass in the expected exception type into these methods. That way, JUnit can give an extremely useful message if a different exception is thrown (including a stack trace).

To get the behavior you want, users can pass in `Throwable.class` as the first parameter. We can always add the methods you are asking for to JUnit proper later.

I personally prefer `expectThrows` over `expectedThrown`, but I'll defer to @marcphilipp on method naming.

src/main/java/org/junit/Assert.java
kcooney: Typo. I meant `expect`. 

Marc and I are fine with what you proposed 9 days ago with `assertThrows` and `expectThrows`. I'm looking forward to the updated pull.

src/test/java/org/junit/tests/assertion/AssertionTest.java
rschmitt: Actually, I'd rather give that name to the parameter that is currently called `throwableClass`. It makes more sense in this context.

src/main/java/org/junit/Assert.java
rschmitt: I've already added the `assertThrows` variant, I'm just working on some minor feedback from a few weeks ago now.

"I would just do a simple cast here; you know it's safe because you just checked, and it's (slightly) more efficient and just as clear."

Believe it or not, when I use a simple cast I get a compiler warning about an unchecked cast. It goes away when I use the `cast` method. I'm not sure what to do with that information. Should I go with a simple cast anyway? If so, should I add a `@SuppressWarnings` annotation?

src/main/java/org/junit/Assert.java
rschmitt: Okay, but I'm going to put some punctuation at the end of the message. The readability difference is huge:

```
unexpected exception thrown expected:<ArrayIndexOutOfBoundsException> but was:<OutOfMemoryError>
unexpected exception thrown; expected:<ArrayIndexOutOfBoundsException> but was:<OutOfMemoryError>
```

src/test/java/org/junit/tests/assertion/AssertionTest.java
marcphilipp: +1

src/main/java/org/junit/Assert.java
rschmitt: Done

rschmitt:
Okay, I think that's everything. Any additional comments on the latest revision (7d49b908248e1737c29a8a305f3ecdcc15a2eaf0)?

marcphilipp:
LGTM!

@kcooney PTAL

src/main/java/org/junit/Assert.java
kcooney: @marcphilipp do you have thoughts about whether this should be a top-level class? If we need it in another class, or if API developers want to use our code, then they will either need to import a nested class or introduce their own version. I don't have a strong feeling, but wanted to make sure you saw this thread.

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,63 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+     * executed. If it does not throw an exception, an {@link AssertionError} is thrown. If it
+     * throws the wrong type of exception, an {@code AssertionError} is thrown describing the
+     * mismatch; the exception that was actually thrown can be obtained by calling {@link
+     * AssertionError#getCause}.
+     *
+     * @param expectedThrowable the expected type of the exception
+     * @param runnable       a function that is expected to throw an exception when executed
+     * @since 4.13
+     */
+    public static <T extends Throwable> void assertThrows(Class<T> expectedThrowable, ThrowingRunnable runnable) {
```
kcooney: I believe this can be:

    public static void assertThrows(Class<? extends Throwable> expectedThrowable, ...)

There's only a need to make the method generic if the value is used in more than once place

src/main/java/org/junit/Assert.java
```diff
@@ -955,4 +955,63 @@ public static void assertEquals(Object[] expecteds, Object[] actuals) {
...
+     * @param runnable       a function that is expected to throw an exception when executed
+     * @return the exception thrown by {@code runnable}
+     * @since 4.13
+     */
+    public static <T extends Throwable> T expectThrows(Class<T> expectedThrowable, ThrowingRunnable runnable) {
+        try {
+            runnable.run();
+        } catch (Throwable actualThrown) {
+            if (expectedThrowable.isInstance(actualThrown)) {
+                return expectedThrowable.cast(actualThrown);
```
kcooney: I have a slight preference for not using `cast()` here and instead using a simple cast and annotating with `@SuppressWarnings("unchecked")` with a comment indicating why the cast is safe.

src/test/java/org/junit/tests/assertion/AssertionTest.java
kcooney: @rschmitt please fix the imports to match the previous ordering

kcooney:
@rschmitt looks very nice. Just some minor issues.

rschmitt:
Done.

kcooney:
Thanks @rschmitt ... LGTM. Just waiting for Marc's thoughts on `ThrowingRunnable` being top-level vs. nested.

rschmitt:
Yeah, that's a tricky one. There's no way to get around it being public, and it's not really a perfect fit for any existing classes or packages. I have two suggestions:

* Make it a top-level class in `org.junit.internal`. I think it's unlikely that someone will want to use this interface for something other than a lambda-friendly API, but if that _does_ occur you can at least promote it to the public API. (Distressingly, this change would not be binary compatible; even if nobody imports `ThrowingRunnable` in their Java source and just uses lambdas, javac will still generate a class that implements it explicitly.)
* Make it a top-level class in `org.junit` and just leave it there. What's the worst that could happen? It's not like you can accidentally depend on it in production code (unless you've already added junit as a `compile`-scoped Maven dependency or something).

kcooney:
I don't think it makes sense to put it in an internal package. End users would effectively be creating anonymous classes that implement it.

I actually was thinking we should put the interface in a new package, `org.junit.function`.

rschmitt:
That sounds like a decent idea if and only if you already have two or three highly plausible examples of other classes that could go in that package. It strikes me as odd to have a package with just one class in it, even though this is Java and the packages don't matter.

marcphilipp:
Let's leave it where it is now. We can easily move it to a top-level class later on. We'd have to deprecate it in `Assert` and have it extend the top-level interface.

rschmitt:
In that case, I guess we're done! Let me know if I should squash or rebase anything.

marcphilipp:
We'll squash them ourselves, when we merge this pull request. This way, the conversations in this pull will stay readable, at least as readable as they are right now. I will merge it tonight.

dsaff:
Belated, but +1 to @marcphilipp's approach here.

marcphilipp:
Squashed and merged. Thank you very much, @rschmitt!

@rschmitt Can you please add a section about the two new methods to the [release notes](https://github.com/junit-team/junit/wiki/4.13-Release-Notes)?

rschmitt:
I'll be sure to do that. Is there a tentative release date for 4.13?

rschmitt:
Release notes updated.

kcooney:
FYI: we are proposing deleting `expectThrows` and changing `assertThrows()` to return the.m caught exception. If you have objections or concerns, please comment on pull #1396

PR #1157, opened 2015-06-04
TestName: Make 'name' field volatile

rschmitt:
This ensures that the name is published across threads correctly--for
instance, if a parallelized runner is used.

### DIFF
--- a/src/main/java/org/junit/rules/TestName.java
+++ b/src/main/java/org/junit/rules/TestName.java
@@ -25,7 +25,7 @@ import org.junit.runner.Description;
  * @since 4.7
  */
 public class TestName extends TestWatcher {
-    private String name;
+    private volatile String name;
 
     @Override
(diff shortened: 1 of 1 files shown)


kcooney:
Merged. Thanks!

I think in practice, we were probably fine, since it would be hard for a thread that did not update that field to get a reference to the rule. But making it volatile is safer and there's almost no overhead on most processors.

rschmitt:
Suppose you have a `Timeout` downstream of `TestName`. The `name` field will be set, and then `Timeout` will spawn a separate thread in which the actual test will run. Currently this _should_ be safe because it creates a new `Thread` object for every test invocation, but if this implementation detail ever changed (e.g. `Timeout` reuses threads in a thread pool) undefined behavior will result.

kcooney:
Ah, I see. I could easily imagine custom Timeout implementations that use a thread pool

PR #1158, opened 2015-06-04
RFC: TestRule potpourri

rschmitt:
This is a sketch of a few TestRules I would find useful.

* `RepeatRule` repeats tests some constant number of times.
* `ParallelRepeatRule` is a parallel version of `RepeatRule` that executes tests in the supplied thread pool.
* `RetryRule` uses an injectible retry policy to decide whether failed tests should be retried.

The intended use of the first two is simple generative tests, where test cases are generated nondeterministically and some specific invariant is asserted (e.g. round-trip tests for serialization). The `RetryRule` is intended for use with integration tests that may fail spuriously. Whether to retry is decided by another object (which should probably be some kind of `TestWatcher` instance); the intention is to support different types of retry strategies, such as:

* Retry a fixed number of times per method
* Retry a fixed number of times per class
* Budget a global number of retries
* Only retry certain exceptions
* Retry unless too many failures (within some scope--suite, class...) have been observed recently

I searched GitHub to look for prior issues/PRs like this, but the only related-looking thing I found was a class in `junit.framework`. This strikes me as pretty straightforward and useful functionality, so I want to see what the project maintainers think. Which of these suggestions are viable?

### DIFF
--- a/src/main/java/org/junit/Repeat.java
+++ b/src/main/java/org/junit/Repeat.java
@@ -1,0 +1,12 @@
+package org.junit;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Repeat {
+    int value() default 1;
+}
--- a/src/main/java/org/junit/rules/ParallelRepeatRule.java
+++ b/src/main/java/org/junit/rules/ParallelRepeatRule.java
@@ -1,0 +1,63 @@
+package org.junit.rules;
+
+import org.junit.Repeat;
+import org.junit.runner.Description;
+import org.junit.runners.model.MultipleFailureException;
+import org.junit.runners.model.Statement;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+
+public class ParallelRepeatRule implements TestRule {
+    private final ExecutorService executorService;
+
+    public ParallelRepeatRule(ExecutorService executorService) {
+        this.executorService = executorService;
+    }
+
+    public Statement apply(final Statement base, Description description) {
+        Repeat annotation = description.getAnnotation(Repeat.class);
+        final int iterations = annotation == null ? 1 : annotation.value();
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                List<Future<?>> futures = new ArrayList<Future<?>>();
+                for (int i = 0; i < iterations; i++) {
+                    futures.add(executorService.submit(new Callable<Void>() {
+                        public Void call() throws Exception {
+                            try {
+                                base.evaluate();
+                                return null;
+                            } catch (Exception ex) {
+                                throw ex;
+                            } catch (Throwable t) {
+                                throw new RuntimeException(t);
+                            }
+                        }
+                    }));
+                }
+                List<Throwable> throwables = new ArrayList<Throwable>();
+                boolean interrupted = false;
+                for (Future<?> future : futures) {
+                    try {
+                        future.get();
+                    } catch (ExecutionException ex) {
+                        throwables.add(ex.getCause());
+                    } catch (InterruptedException ex) {
+                        interrupted = true;
+                    }
+                }
+                if (interrupted) {
+                    Thread.currentThread().interrupt();
+                }
+                if (!throwables.isEmpty()) {
+                    throw new MultipleFailureException(throwables);
+                }
+            }
+        };
+    }
+}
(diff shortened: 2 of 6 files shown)


kcooney:
Thanks for the contribution!

Unfortunately, if a Rule repeats/reruns a test, you won't get a unique instance for the reruns, so fields wound not get reset but @Before methods and the rules down the rule chain would be run again. Thus would be confusing

I vote we close this pull

rschmitt:
It's starting to sound like the abstractions that JUnit provides are inadequate.

On the one hand, we've got `Runner`, which is potentially very powerful, but that power can only be accessed through subclassing. I can't compose two different `Runner`s to combine their effects.

On the other hand, we have `TestRule`, which is interface-based and composes well, but which isn't powerful enough to safely extend or modify the way test methods are executed, therefore making them largely useless for anything other than simple callbacks (e.g. `TemporaryFolder`, `ErrorCollector`, `Stopwatch`) and filters (`Verifier`). The main exception is `Timeout`.

Is there a really good reason that it should be impossible to correctly implement `ParallelRepeatRule` or `RetryRule` using a `TestRule`? How would you obtain this behavior without using `@RunWith`?

kcooney:
Currently a custom runner is the only way I know to do what you want. I agree that is an unsatisfactory situation, given that a class has one runner and can't use composition.

Possibly users should be able to specify a `Statement` factory for a test class. That factory could handle what you are trying to do with `RepeatRule` and `RetryRule` and could possibly support parameterized tests and/or class/method injection. One tricky part would be how we would allow them to be chained.

If that sounds like it would be useful, please create a new issue.

rschmitt:
Why is it that setup/teardown are run as part of `Statement#evaluate`, but instantiation is performed out-of-band? The test framework obviously knows how to instantiate test classes, it doesn't need to be told. Maybe that logic is in the wrong place? Can the `newInstance` call take place in a `Statement` too? I'm currently reading through various classes in the `BlockJUnit4ClassRunner` hierarchy to get a sense of how this is done. Inheritance =(

rschmitt:
Welp, it's impossible. `BlockJUnit4ClassRunner` and its various helper classes all assume that the test class instance can be known a priori. The `test` object is instantiated and then woven into a bunch of `Statement` and `TestRule` and `MethodRule` implementations that perform crucial jobs (such as invoking setup and teardown methods), and there's no way to fix that coupling without breaking anything that subclasses `BlockJUnit4ClassRunner`. You'd have to create _yet another test runner_ that offers this functionality.

What's interesting is that replacing the `Object` with a `Supplier` or `AtomicReference` (anything that provides indirection, really) has some interesting potential. For instance, you can implement `RetryRule` et al correctly of course, but you can _also_ implement TestNG's behavior of reusing the same instance across all invocations.

kcooney:
I don't think it's entirely impossible. The `methodBlock()` function encodes all the logic for creating a `Statement` that runs the test with a new instance, so if someone provides a `StatementFactory` for their test class, that factory could be called to do the work, and it could be passed in a `StatementFactory` that performs the steps that are now in `methodBlock` (similar to how a `TestRule` is given a `Statement` that runs the test)

That being said, I've been starting to think that if we wanted to provide more powerful extension points (like rules that can act on the class level or the method level, or allowing a test to be re-run, or allowing parameterized tests) we might need a replacement for `BlockJUnit4TestRunner`. That would sadly mean that other runners won't get the new functionality.

I suggest creating a new issue that describes the root problem you want to solve, without going into too much detail about how we might solve it. For example, you could request an extension point that would allow you to have a test re-run once if it fails, and you want to be able to do that without defining your own `Runner`

rschmitt:
> I don't think it's entirely impossible. The methodBlock() function encodes all the logic for creating a Statement that runs the test with a new instance

It _encodes_ that logic, but it doesn't actually _encapsulate_ it. The code in question is an inheritance nightmare. The `encodeBlock` method and essentially all of the methods called by it are protected, and that makes them part of the public API. _That's_ why I said that this is impossible to fix with the current testrunner. If I submitted a PR with the changes you suggest, it would be rejected immediately as a breaking change. Seriously, Google "extends BlockJUnit4ClassRunner"; it's eye-opening. Here are two results I found immediately that override the specific methods we're discussing:

http://saltnlight5.blogspot.com/2012/09/enhancing-spring-test-framework-with.html
http://andyonsoftware.blogspot.com/2011/05/custom-junit-runner.html

In any event, I think the whole "Runner" design is a dead end. It's doing too much, and it's too monolithic to reasonably extend and compose. I think an interesting question is how many pieces we can cut the Runner up into.

kcooney:
*The code in question is an inheritance nightmare*

I actually agree. Making changes to it without breaking people is extremely hard. Allowing users to use delegation vs. inheritance as been a common theme for JUnit4, but we have a long way to go. If I could devote 3 months to JUnit and had someone to pair with, I would love to eliminate the need for most code outside JUnit itself write a Runner.

Please do file a feature request. Github review threads are horrible for capturing conversations, as you've probably seen in your other pull.

rschmitt:
Ok, will do. This PR was a spike intended to facilitate just this kind of discussion.

PR #1163, opened 2015-06-09
Linked Throwable

Jire:
Added link to throwable and emphasized


kcooney:
Thanks for the contribution!

We generally don't link to classes under `java.lang`. See http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#styleguide

The generated JavaDoc already has a link to `Throwable ` so I am leaning towards closing this pull

PR #1164, opened 2015-06-10
Upgrade plugins to latest version

hansjoachim:
Upgraded all plugins to the latest available version. Additionally, I added version numbers for a couple which didn't set an explicit version number.

I haven't tested that all plugins still work as expected, since some of them are a bit tricky for me to check. Notable examples are the release- and gpg-plugin. I have verified that `mvn clean install` runs succesfully.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -347,6 +347,22 @@
                     </archive>
                 </configuration>
             </plugin>
+            <plugin>
+                <artifactId>maven-clean-plugin</artifactId>
+                <version>2.6.1</version>
+            </plugin>
+            <plugin>
+                <artifactId>maven-deploy-plugin</artifactId>
+                <version>2.8.2</version>
+            </plugin>
+            <plugin>
+                <artifactId>maven-install-plugin</artifactId>
+                <version>2.5.2</version>
+            </plugin>
+            <plugin>
+                <artifactId>maven-resources-plugin</artifactId>
+                <version>2.7</version>
+            </plugin>
         </plugins>
     </build>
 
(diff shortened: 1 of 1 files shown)


marcphilipp:
Thanks!

Tibor17:
@hansjoachim 
You could update surefire-junit47. I was waiting for that but another committer wanted to split my PR so I did not want to continue.

hansjoachim:
>Better not to update these since they are managed by Maven itself.

True, maven sets default version numbers for these plugins. It also displays warnings if they are not explicitly set when running ` mvn versions:display-plugin-updates`, so I've gotten into the habit of adding them.
```
[WARNING] The following plugins do not have their version specified:
[WARNING]   maven-clean-plugin ........................ (from super-pom) 2.6.1
[WARNING]   maven-deploy-plugin ....................... (from super-pom) 2.8.2
[WARNING]   maven-install-plugin ...................... (from super-pom) 2.5.2
[WARNING]   maven-resources-plugin ...................... (from super-pom) 2.7
```
One might argue that different versions of maven might set different defaults so it is nice to set what ends up being used, but I doubt this will cause problems in practice. I don't have any strong opinions on this, so if you are opposed to listing these plugins I can remove them again. :)

>You could update surefire-junit47.

Sure, can do. Not sure why it didn't show up among the other available updates though.

hansjoachim:
>>You could update surefire-junit47.

>Sure, can do. Not sure why it didn't show up among the other available updates though.

@Tibor17 Looked into this now. I guess the reason it didn't show up was that it is only used for the fast-tests profile.

However, I'm confused because if I run `mvn clean verify -P fast-tests` it just seems to hang after a couple of test suites. This happens both before and after changing the version number of surefire-junit47, so I while I can upgrade it, I'm not able to verify it works as intended either before or after. Presumably it has worked at some point, so it might be just me having bad luck.

PR #1170, opened 2015-07-01
Junit5: Fix #1169 Throwing MultipleFailureException with empty list of throwables forces test to fail

### LINKED ISSUE #1169: Throwing MultipleFailureException with empty list of throwables forces test to pass

``` java
@Test
  makeTestSucceed();
  assertEquals(1, 2);
}
}
```

ShengyuanLu:
The `MultipleFailureException` instance thrown by test case is collected, instead of collect `getFailures()` only;
The case `ParentRunnerTest#parentRunnerTestMethods` is modified to verify the behavior. 

Thanks!


ShengyuanLu:
Close this pull request, because I used wrong 'master' branch. 
New pull request to **'junit5'** submitted.
Sorry for the confuse!

PR #1171, opened 2015-07-01
Fix #1169 Throwing MultipleFailureException with empty list of throwables forces test to fail

### LINKED ISSUE #1169: Throwing MultipleFailureException with empty list of throwables forces test to pass

ShengyuanLu:
The `MultipleFailureException` instance thrown by test case is collected, instead of collect `getFailures()` only;
The case `ParentRunnerTest.parentRunnerTestMethods()` is modified to verify the behavior.

Thanks!

### DIFF
--- a/src/main/java/org/junit/internal/runners/model/EachTestNotifier.java
+++ b/src/main/java/org/junit/internal/runners/model/EachTestNotifier.java
@@ -25,6 +25,9 @@ public class EachTestNotifier {
     }
 
     private void addMultipleFailureException(MultipleFailureException mfe) {
+        if(mfe.getFailures().isEmpty()) {
+            addFailure(mfe);
+        }
         for (Throwable each : mfe.getFailures()) {
             addFailure(each);
         }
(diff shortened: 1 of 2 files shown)


kcooney:
I didn't realize that fixing the handling of `MultipleFailureException` would be so easy. I still slightly prefer fixing this in the constructor by throwing `IllegalArguemenyException` but I could be convinced to accept this solution.

@marcphilipp what do you think?

ShengyuanLu:
@kcooney Throwing `IllegalArguemenyException` in constructor works fine, except `@Test(expected = MultipleFailureException.class' will be fail.

And I am making anther fix for it.  

Thanks for your advice!

ShengyuanLu:
[Latest pull request] (https://github.com/junit-team/junit/pull/1173) for the fix was submitted.

kcooney:
@ShengyuanLu instead of creating new pull requests each time you make a change, you can make changes on the same branch, push to your github repo, and ping the thread.

marcphilipp:
I think having the constructor throw an `IllegalArgumentException` when an empty list of errors is passed in would be the better solution, because it does not require changing the handling of `MultipleFailureException` at all.

@ShengyuanLu Do you have time to submit another pull request that only changes the constructor of `MultipleFailureException` and adds a test for it?

PR #1172, opened 2015-07-01
Let maven replacer plugin use sourceDirectory property.

avandeursen:
The old maven replacer plugin was configured with a relative link to the source path.

This caused a failure of

    mvn -f path-to-junit/pom.xml

The modified pom file uses the `project.build.sourceDirectory` property, which solves this problem.


Tibor17:
+1

marcphilipp:
Thanks!

PR #1173, opened 2015-07-02
Fix #1169 Throwing MultipleFailureException with empty list of throwables forces test to fail, and modify MultipleFailureException.getMessage()

### LINKED ISSUE #1169: Throwing MultipleFailureException with empty list of throwables forces test to pass

ShengyuanLu:
Fix #1169 Throwing `MultipleFailureException` with empty list of throwables forces test to fail, and modify `MultipleFailureException.getMessage()`.

Note: 
- @kcooney Throwing `IllegalArguementException` in constructor works fine, except `@Test(expected = MultipleFailureException.class)` will be fail.

- The [previous pull-request](https://github.com/ShengyuanLu/junit/commit/71b17fdde564535e51ccdf2390b0cb3bd1ca342d) is buggy, this is latest one. 

Thanks!

### DIFF
--- a/src/main/java/org/junit/internal/runners/model/EachTestNotifier.java
+++ b/src/main/java/org/junit/internal/runners/model/EachTestNotifier.java
@@ -25,6 +25,9 @@ public class EachTestNotifier {
     }
 
     private void addMultipleFailureException(MultipleFailureException mfe) {
+        if(mfe.getFailures().isEmpty()) {
+            notifier.fireTestFailure(new Failure(description, mfe));
+        }
         for (Throwable each : mfe.getFailures()) {
             addFailure(each);
         }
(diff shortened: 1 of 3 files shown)


kcooney:
@ShengyuanLu the description doesn't match the code changes. Did you mean to add a `throw new IllegalArgumentException("list cannot be empty")` to the `MultipleFailureException` constructor?

marcphilipp:
Closing, see comment at https://github.com/junit-team/junit/pull/1171#issuecomment-118124288.

ShengyuanLu:
@marcphilipp @kcooney 
The fix is updated: https://github.com/junit-team/junit/pull/1175. An `IllegalArgumentException` will be thrown in this case.

Thanks you!

PR #1174, opened 2015-07-02
Avoid calling a deprecated constructor in the JavaDoc sample for the …

cnauroth:
…Timeout rule.

In the JavaDocs for the `Timeout` rule, there is a code sample that uses a constructor that has been deprecated.  This patch is a change in JavaDocs only to avoid using the deprecated constructor, and hence prevent spreading usage of the deprecated constructor by developers copy-pasting the sample.  Thank you!


marcphilipp:
Thanks! :-)

cnauroth:
@marcphilipp , thanks for the quick merge!  :-)

PR #1175, opened 2015-07-03
Fix #1169 An IllegalArgumentException will be thrown when empty list of throwables is passed to MultipleFailureException.

### LINKED ISSUE #1169: Throwing MultipleFailureException with empty list of throwables forces test to pass

ShengyuanLu:
@marcphilipp @kcooney 
Thanks for your advice!

Fix: An `IllegalArgumentException` will be thrown when empty list of throwables is passed to `MultipleFailureException`.

Files changed: https://github.com/junit-team/junit/pull/1175/files

### DIFF
--- a/src/main/java/org/junit/runners/model/MultipleFailureException.java
+++ b/src/main/java/org/junit/runners/model/MultipleFailureException.java
@@ -22,6 +22,9 @@ public class MultipleFailureException extends Exception {
     private final List<Throwable> fErrors;
 
     public MultipleFailureException(List<Throwable> errors) {
+        if(errors.isEmpty()) {
+            throw new IllegalArgumentException("Argument error list cannot be empty");
+        }
         this.fErrors = new ArrayList<Throwable>(errors);
     }
 
--- a/src/main/java/org/junit/runners/model/MultipleFailureException.java
+++ b/src/main/java/org/junit/runners/model/MultipleFailureException.java
@@ -22,7 +22,7 @@ public class MultipleFailureException extends Exception {
     private final List<Throwable> fErrors;
 
     public MultipleFailureException(List<Throwable> errors) {
-        if(errors.isEmpty()) {
+        if (errors.isEmpty()) {
             throw new IllegalArgumentException("Argument error list cannot be empty");
         }
         this.fErrors = new ArrayList<Throwable>(errors);
--- a/src/test/java/org/junit/tests/assertion/MultipleFailureExceptionTest.java
+++ b/src/test/java/org/junit/tests/assertion/MultipleFailureExceptionTest.java
@@ -1,6 +1,7 @@
 package org.junit.tests.assertion;
 
 import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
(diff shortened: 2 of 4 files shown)


src/main/java/org/junit/runners/model/MultipleFailureException.java
```diff
@@ -22,6 +22,9 @@
     private final List<Throwable> fErrors;
 
     public MultipleFailureException(List<Throwable> errors) {
+        if(errors.isEmpty()) {
```
kcooney: Syyle nit: we put spaces after keywords like "if"

src/test/java/org/junit/tests/assertion/MultipleFailureExceptionTest.java
```diff
@@ -67,6 +68,16 @@ public void assertEmptyThrowsMutipleFailureExceptionForManyThrowables() throws E
         }
     }
 
+    @Test
+    public void assertEmptyErrorListConstructorFailure() throws Exception {
+        try {
+            new MultipleFailureException(Collections.<Throwable>emptyList());
+            fail();
```
kcooney: The `fail()`methid throws `AssertionError` which is an `Exception`. Either use `ExpectedException` or change your catch clause to catch the exact exception you are expecting

src/test/java/org/junit/tests/assertion/MultipleFailureExceptionTest.java
```diff
@@ -67,6 +68,16 @@ public void assertEmptyThrowsMutipleFailureExceptionForManyThrowables() throws E
...
+        } catch (Exception expected) {
+            assertThat(expected, instanceOf(IllegalArgumentException.class));
+            assertTrue(expected.getMessage().equals("Argument error list cannot be empty"));
```
kcooney: This won't give a useful message if it fails. You can use Hamcrest to assert that the message contains the expected text

src/main/java/org/junit/runners/model/MultipleFailureException.java
ShengyuanLu: Yes, I will add a space after `if`.

src/test/java/org/junit/tests/assertion/MultipleFailureExceptionTest.java
ShengyuanLu: I will catch `IllegalArgumentException` instead of  `Exception`, although 'AssertionError' is-not-a `Exception` :)

src/test/java/org/junit/tests/assertion/MultipleFailureExceptionTest.java
ShengyuanLu: Nice catch :+1: 
`containsString` will be used instead here.

src/test/java/org/junit/tests/assertion/MultipleFailureExceptionTest.java
kcooney: Sorry, haven't had my coffee yet :-)

My overall point is the code is easier to read if it's consistent and follows conventional idioms.

Alsi, if you only catch what you expect to be thrown, then if something else is thrown you get a useful stack trace in the failure

ShengyuanLu: Got your point and agree!
I am updating my pull request.

ShengyuanLu:
Pull request was updated.

kcooney:
Thanks! Merged. @ShengyuanLu can you please update the JUnit5 release notes?

kcooney:
BTW, I squashed the commits on my side. The commit is here:

https://github.com/junit-team/junit/commit/31f72d37e12a134de67d89bf5504df43e0858867

ShengyuanLu:
Regarding to updating the JUnit5 release notes, I couldn't find `doc\ReleaseNotes5.md`.
Should I create `ReleaseNotes5.md` and.txt/html files?

Thanks for your time for reviewing my fix,  **Really**!

kcooney:
Hmm...I don't see release notes for JUnit 5 either

@junit-team/junit-committers did we not start the release notes?

marcphilipp:
No, we haven't started them yet. Please let's put them here from now on:
https://github.com/junit-team/junit/wiki/5.0-Release-Notes

ShengyuanLu:
@marcphilipp The issue has been added to https://github.com/junit-team/junit/wiki/5.0-Release-Notes.

marcphilipp:
Thanks!

PR #1176, opened 2015-07-03
MethodSorter METHOD_ORDER allowing to specify the execution order of test methods

adamsiemion:
- added attribute order to the Test annotation for specyfing the execution order
- created a new MethodSorter (METHOD_ORDER) and added it to the MethodSorters enum
- created a unit test for this feature

### DIFF
--- a/src/main/java/org/junit/Test.java
+++ b/src/main/java/org/junit/Test.java
@@ -95,4 +95,10 @@ public @interface Test {
      * </p>
      */
     long timeout() default 0L;
+
+    /**
+     * Optionally specify the execution order of a test method. Methods are executed from the lowest <code>order</code>
+     * to the highest.
+     */
+    int order() default 0;
 }
--- a/src/main/java/org/junit/internal/MethodSorter.java
+++ b/src/main/java/org/junit/internal/MethodSorter.java
@@ -5,6 +5,7 @@ import java.util.Arrays;
 import java.util.Comparator;
 
 import org.junit.FixMethodOrder;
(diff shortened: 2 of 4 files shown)


tobiasKaminsky:
This would be a great feature for me as it is more intuitive to have the tests executed in the same order as they are written.
Of course this only matters if the tests needs to be in a specific order.
Currently I avoid this problem by adding numbers to the front of the tests and use NAME_ORDER.

(And I know that a testsuite should work without a specific order, but in this case it is not possible :/ )

cardil:
:+1: for this feature. BTW. Random order, just like Ruby rspec, whould be nice to!! Mayby I will do that in seperate PR

tobiasKaminsky:
@cardil As far as I have understand and read the current sorting oder is random, but always the same for a specific computer and java machine.

cardil:
@tobiasKaminsky Yup. hashcodes for strings doesn't change beetween JVM executions. So, the truely random order would be nice.

BTW. This discusion is offtopic :-P so I've created new issue for that: #1194

kcooney:
There are many ways that people could want to specify the ordering of methods, so I personally would prefer that we added an extension point that allowed you to define your own class that does the ordering or sorting.

See #1130 for one way we could do that

kcooney:
I don't support specifying an ordering in the `Test` annotation. The JUnit team has long encouraged tests to not depend on ordering. While I can see why some projects might need to specify an ordering, I don't think adding support to the `Test` annotation would encourage it too strongly.

PR #1183, opened 2015-07-22
Use Unix-style line endings.

rohitagr:
Furthermore, add a .gitattributes file to enforce Unix-style linebreaks on all text files. Currently, most files are using Unix-style linebreaks, but some use Windows-style CRLF.


kcooney:
This will, unfortunately, result in this change being associated with all of the changes for these files, which makes "blame" less useful.

rschmitt:
`git blame -w`

rohitagr:
As @rschmitt said, this is not a problem from the command line, but it appears github's blame doesn't support ignoring whitespace. Should I abandon this?

kcooney:
It looks like it should be possible to "grandfather" the existing files that have CRLF endings, but require all other existing and new files to use LF endings. Can we do that?

kcooney:
I think `.gitattributes` supports specifying an override for the line endings for specific files, so I'd like to do that.

marcphilipp:
Replaced by #1405.

PR #1187, opened 2015-08-07
Access java.lang.management.* via reflection

rohitagr:
Several platforms, including Android and j2objc, do not provide the java
management APIs and crash or otherwise fail when encountering code that
imports or uses these APIs directly.


kcooney:
Thanks for doing this. I'm surprised no one reported previously that these APIs don't work on Android.

The changes LGTM

@marcphilipp any concerns?

rohitagr:
They generally don't cause problems if not used on Android, but proguard complains heavily (and dies with the default config). On j2objc the imports always cause problems.

rohitagr:
Is anything blocking this?

PR #1188, opened 2015-08-07
Updated answer to question in faq.html

gk2go:
I've updated answer as discussed on JUnit's Yahoo! Group. 
@dsaff,  I think it is now under ~100 columns.
@kcooney, is this pull request now submitted against junit-team/junit?

UPDATED ANSWER:
If you are using JDK 1.3 or higher, you can use reflection to subvert the access control mechanism with the aid of the <a href="http://sourceforge.net/projects/privaccessor/">PrivilegedAccessor</a>. For details on how to use it, read <a href="http://www.onjava.com/pub/a/onjava/2003/11/12/reflection.html">this article</a>.
If you are using JDK 1.6 or higher and you annotate your tests with @Test, you can use <a href="http://dp4j.com">Dp4j</a> to inject reflection in your test methods. For details on how to use it, see <a href="http://dp4j.com/testscript">this test script</a>.

### DIFF
--- a/src/site/fml/faq.fml
+++ b/src/site/fml/faq.fml
@@ -720,11 +720,13 @@ public void testIndexOutOfBoundsExceptionNotRaised()
       For details on how to use it,
       read <a
       href="http://www.onjava.com/pub/a/onjava/2003/11/12/reflection.html">this
-      article</a>.<br>
-	  If you are using JDK 1.6 or higher and you annotate your tests with @Test,
-	  you can use <a href="http://dp4j.com">Dp4j</a> to inject reflection in your
-	  test methods. For details on how to use it, see <a 
-	  href="http://dp4j.com/testscript">this test script</a>.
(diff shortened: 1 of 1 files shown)


dsaff:
@gk2go, this appears to have broken the build, see above.

src/site/fml/faq.fml
```diff
@@ -720,7 +720,11 @@ public void testIndexOutOfBoundsExceptionNotRaised()
       For details on how to use it,
       read <a
       href="http://www.onjava.com/pub/a/onjava/2003/11/12/reflection.html">this
-      article</a>.
+      article</a>.<br>
```
kcooney: I think the `<br>` tag is what is breaking the build. Instead, use `<p>` and `</p>` around paragraphs.

gk2go:
@dsaff @kcooney, shall I close the pull request now?

PR #1190, opened 2015-08-12
Fixes #1189: JUnit4TestAdapter blocks JUnit-3-style suites that are included via @RunWith(Suite.class)

### LINKED ISSUE #1189: JUnit4TestAdapter blocks JUnit-3-style suites that are included via @RunWith(Suite.class)
The intention of junit.framework.JUnit4TestAdapter is to allow a JUnit-3-style test runner to run JUnit-4-style tests.

However, due to a bug in JUnit, this doesn't work when a JUnit-4-style test suite (@RunWith(Suite.class)) includes a JUnit-3-style suite class.
Happened to us in https://bugs.eclipse.org/bugs/show_bug.cgi?id=474777

To reproduce in JUnit itself:
- remove the @RunWith(AllTests.class) annotation from org.junit.tests.junit3compatibility.OldTests
- run the whole org.junit.tests.AllTests:
  - if you run it with a JUnit-4-style runner, it passes
  - but with a JUnit-3-style runner that calls the suite() method, it fails with:
```
AllTests (org.junit.tests)
org.junit.tests.AllTests
org.junit.tests.junit3compatibility.OldTests
initializationError(org.junit.tests.junit3compatibility.OldTests)
java.lang.Exception: No runnable methods
	at org.junit.runners.BlockJUnit4ClassRunner.validateInstanceMethods(BlockJUnit4ClassRunner.java:202)
	at org.junit.runners.BlockJUnit4ClassRunner.collectInitializationErrors(BlockJUnit4ClassRunner.java:138)
	at org.junit.runners.ParentRunner.validate(ParentRunner.java:416)
	at org.junit.runners.ParentRunner.<init>(ParentRunner.java:84)
	at org.junit.runners.BlockJUnit4ClassRunner.<init>(BlockJUnit4ClassRunner.java:68)
	at org.junit.internal.builders.JUnit4Builder.runnerForClass(JUnit4Builder.java:10)
	at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:59)
	at org.junit.internal.builders.AllDefaultPossibilitiesBuilder.runnerForClass(AllDefaultPossibilitiesBuilder.java:26)
	at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:59)
	at org.junit.runners.model.RunnerBuilder.runners(RunnerBuilder.java:101)
	at org.junit.runners.model.RunnerBuilder.runners(RunnerBuilder.java:87)
	at org.junit.runners.Suite.<init>(Suite.java:102)
	at org.junit.runners.Suite.<init>(Suite.java:70)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:39)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:27)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:501)
	at org.junit.internal.builders.AnnotatedBuilder.buildRunner(AnnotatedBuilder.java:108)
	at org.junit.internal.builders.AnnotatedBuilder.runnerForClass(AnnotatedBuilder.java:86)
	at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:59)
	at org.junit.internal.builders.AllDefaultPossibilitiesBuilder.runnerForClass(AllDefaultPossibilitiesBuilder.java:26)
	at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:59)
	at org.junit.internal.requests.ClassRequest.getRunner(ClassRequest.java:33)
	at junit.framework.JUnit4TestAdapter.<init>(JUnit4TestAdapter.java:40)
	at junit.framework.JUnit4TestAdapter.<init>(JUnit4TestAdapter.java:34)
	at org.junit.tests.AllTests.suite(AllTests.java:246)
...
```
I'm preparing a pull request.

mkeller:


src/main/java/org/junit/internal/builders/AllDefaultPossibilitiesBuilder.java
```diff
@@ -40,7 +40,10 @@ protected JUnit3Builder junit3Builder() {
     }
 
     protected AnnotatedBuilder annotatedBuilder() {
-        return new AnnotatedBuilder(this);
+        // For compatibility with Request#classes(...) and the Suite runner,
+        // deeper nesting levels always need to support suite() methods:
+    	RunnerBuilder suiteBuilder = canUseSuiteMethod ? this : new AllDefaultPossibilitiesBuilder(true);
```
marcphilipp: The indentation seems off here.

src/test/java/org/junit/tests/junit3compatibility/JUnit4TestAdapterTest.java
```diff
@@ -0,0 +1,86 @@
...
+    @Suite.SuiteClasses(TestSuite3.class)
+    public static class TestSuite4ForTestSuite3 {
+    }
+    
+    /**
+     * Test a JUnit-4-style test suite that contains a test
+     * class that is a JUnit-3-style test suite.
+     */
+    @Test
+    public void testSuite4ForTestSuite3() {
```
marcphilipp: We don't normally use Javadoc to describe what a test tests but rather try to express that in the method name. Can you try to do that for all the tests in this class?

marcphilipp:
@dsaff Do you have time to take a look at this?

kcooney:
I worry that this change may have unintended consequences. For one, ifsomeone subclassed`AllDefaultPossibilitiesRunnerBuilder` this change would break them.

I am having problems wrapping my head around the problem (possibly due to lack of sleep :-). Can you include a simple reproducible test case in the bug?

src/main/java/org/junit/internal/builders/AllDefaultPossibilitiesBuilder.java
```diff
@@ -40,7 +40,10 @@ protected JUnit3Builder junit3Builder() {
     }
 
     protected AnnotatedBuilder annotatedBuilder() {
-        return new AnnotatedBuilder(this);
+        // For compatibility with Request#classes(...) and the Suite runner,
+        // deeper nesting levels always need to support suite() methods:
+        RunnerBuilder suiteBuilder = canUseSuiteMethod ? this : new AllDefaultPossibilitiesBuilder(true);
```
kcooney: As far as I can tell, the only way `canUseSuiteMethod` can be false is if you either explicitly created `AllDefaultPossibilitiesBuilder` and passed in false, or if you used `Request.classWithoutSuiteMethod()`. In either case, not using the suite method sounds like expected behavior.

Am I missing something

mkeller: AllDefaultPossibilitiesBuilder is clearly an internal class without any API, so nobody playing by the rules can be broken by my proposed fix.

There is no use case for the current behavior of `new AllDefaultPossibilitiesBuilder(false)`. It's just an implementation bug that should be fixed in JUnit 4. The Javadoc of Request#classWithoutSuiteMethod() supports that view by saying that it "will run all the tests in a class. If the class has a suite() method, it will be ignored."
=> Only the suite method of the given class should be ignored.

The only reason why that special request exists is to avoid the endless nesting that would otherwise occur if a test class uses the JUnit4TestAdapter in its suite() method (according to the API of JUnit4TestAdapter, see the added Javadoc in my fix).

mkeller: If you don't like the additional AllDefaultPossibilitiesBuilder or the additional RunnerBuilder, you can also take this variant, which solves the problem by replacing the bad AllDefaultPossibilitiesBuilder#canUseSuiteMethod with a field that holds the class for which to block the suite() method:
https://github.com/mkeller/junit/commit/5fea7c6e8f8c3cbfc142c0bc62c471e3079a0031

kcooney:
Unfortunately, many people directly use JUnit internal classes. Just in a few seconds I found this: http://www.programcreek.com/java-api-examples/index.php?api=org.junit.internal.builders.AllDefaultPossibilitiesBuilder

We have a very large number of users, so a breaking change that should affect no one will affect someone.

What's wrong with my proposal?

mkeller:
> What's wrong with my proposal?

It could probably be cleaned up and would work, but it would just be wrong to keep the bad constructor and hack around it to implement JUnit4TestAdapter properly. It doesn't make sense to have internal packages and then still treat them as untouchable API. To consistently take the wrong way, wouldn't you have to do the same dance with ClassRequest as well (to avoid breaking users that rely on undocumented behavior)? Once you're in the mud, you never get out again.

kcooney:
@mkeller regardless of whether we believe that other third-party projects should depend on code in internal packages, they do (and I don't blame them, since multiple times we made classes internal that should have been external). Since we know that this particular class is used by many projects, we should consider the cost of breaking them vs. the cost of having a clean internal API.

How about we compromise and add a second constructor to `AllDefaultPossibilitiesBuilder` that takes in a class instance, and update `ClassRequest` to pass in the class? That way we get the behavior you want, a mostly clean API, and we won't break anyone who creates or subclasses `AllDefaultPossibilitiesBuilder`.

I am less concerned about the behavior change in `ClassRequest`; I think the API makes the intent clear, and we wouldn't be causing compile-time breakages. It still could cause different tests to run then ran before the change, so I would recommend making these changes in the JUnit5 branch.

kcooney:
I'm closing this in favor of #1344

Issue #1182, opened 2015-07-09
Rename leftover fields with f-prefix


marcphilipp:
Due to serialization issues (see #976) we have decided not to rename certain fields in classes that implement `Serializable`. We should increment the `serialVersionUID` of those classes and rename the fields in 5.0.

Here's a list of classes:
- ArrayComparisonFailure.java - junit/src/main/java/org/junit/internal
- AssumptionViolatedException.java - junit/src/main/java/org/junit/internal
- ComparisonFailure.java - junit/src/main/java/org/junit
- Description.java - junit/src/main/java/org/junit/runner
- Failure.java - junit/src/main/java/org/junit/runner/notification
- InitializationError.java - junit/src/main/java/org/junit/internal/runners
- InitializationError.java - junit/src/main/java/org/junit/runners/model
- MaxHistory.java - junit/src/main/java/org/junit/experimental/max
- MultipleFailureException.java - junit/src/main/java/org/junit/runners/model

kcooney:
This would likely break people

PR #1191, opened 2015-08-18
Removef "f" prefixes from field names

alexpanov:
Fixed issue #1182, updated plugin versions


src/main/java/org/junit/ComparisonFailure.java
```diff
@@ -16,15 +16,15 @@
...
      */
     private static final int MAX_CONTEXT_LENGTH = 20;
-    private static final long serialVersionUID = 1L;
+    private static final long serialVersionUID = 2L;
 
     /*
      * We have to use the f prefix until the next major release to ensure
      * serialization compatibility. 
      * See https://github.com/junit-team/junit/issues/976
      */
```
marcphilipp: We should remove these comments, too, shouldn't we?

alexpanov: @marcphilipp Yes. I have removed this comment and all the others that reference the same issue. I have also found another class with an f-prefixed field name and fixed it. Please take a look.

kcooney:
Although I appreciate the effort, and I'm very much not a fan of having classes with field names starting with `f` I don't think we should submit this change.

In an early RC for 4.12 we tried renaming the fields, and it broke at least one IDE due to the serialization incompatibilities. I suspect numerous IDEs and build tools would be broken by the serialization changes, and many projects would be broken that access private fields by reflection (yes, it's ugly, and we prefer people ask us to provide extension points, but many projects continue to access fields).

All combined, these changes would delay the migration to JUnit5.

The benefits of renaming the fields are quite small (minor readability improvements).

I recommend we not rename fields in classes unless we need to make other changes to the classes that cannot be made in a way that is backwards and forwards compatible with serialization. When that happens, we can more easily weigh the befits of adding the features vs the cost of breaking people.

The JUnit Lambda effort has as one of it's goals "Decouple test execution and reporting from test definition and provisioning". Hopefully that will involve writing a test execution API that isn't as fragile with respect to serialization. When that API exists and is adopted, it should be safer to rename these fields.

PR #1195, opened 2015-08-27
Remove src/main/java/junit/runner/Version.java

mureinik:
This file is autogenerated from
src/main/java/junit/runner/Version.java.template during the build.

Keeping it in the SCM just generates noise, as you will may see it as
modified when you run mvn install. This is especially true for
newcommers, and may create a negative first impression of the project.


marcphilipp:
Thanks!

PR #1199, opened 2015-09-04
With ExpectedException, document that there should be no statements after the exceptional one.

eamonnmcmanus:
Add text to the documentation of ExpectedException that points out that it is a mistake to have statements after the one that is expected to throw the exception. A common bug pattern is to have verification statements after that statement, like Mockito.verifyNoMoreInteractions. These will of course never be executed, but an imperfect understanding of exactly how ExpectedException works leads people not to realize that.


kcooney:
Thanks for the contribution!

I wonder if we should go further and deprecated this class in favor of https://github.com/junit-team/junit/commit/bdb1799a6b58c4ab40c418111333f4e8e1816e7ee (or at the very least suggest it as an option)

eamonnmcmanus:
GitHub linkified `bdb1799` as a broken link but if I understand correctly it is a change that introduces a method `Assert.assertThrows(Class<? extends Throwable> expectedThrowable, ThrowingRunnable runnable)`. That is often much better than ExpectedException, especially with Java 8 where it is nicely succinct. However, it doesn't completely replace ExpectedException because it doesn't allow you to assert things about the exception message or its cause.

kcooney:
@eamonnmcmanus I fixed the link

The `expectThrows()` method returns the exception that was thrown by the lamba. You can make assertions on the exception message either with simple JUnit assert methods or with an assertion framework like Truth.

eamonnmcmanus:
I hadn't picked up on `expectThrows`. In that case, I think using `assertThrows`/`expectThrows` is clearly better than using `ExpectedException` on Java 8 or later. With earlier versions of Java the verbosity of anonymous classes makes that less clear. So probably deprecation would be too strong. What do you think of the [wording](https://github.com/junit-team/junit/pull/1199/files#diff-2bf153b68f2b7e3f28a61be21d98cdd3R17) I've added to the beginning of `ExpectedException`?

src/main/java/org/junit/rules/ExpectedException.java
```diff
@@ -14,7 +14,10 @@
 
 /**
  * The {@code ExpectedException} rule allows you to verify that your code
- * throws a specific exception.
+ * throws a specific exception. Note that {@link org.junit.Assert#assertThrows Assert.assertThrows}
```
kcooney: I new wording looks good, but probably belongs in aa new paragraph, probably near the bottom.of the Javadoc.

I am also not sure if the link tags will work if you don't specify parameters.

Finally, I think you can remove the parenthesis around the last sentence.

src/main/java/org/junit/rules/ExpectedException.java
```diff
@@ -34,16 +41,19 @@
...
+ * <p>You have to add the {@code ExpectedException} rule to your test.
  * This doesn't affect your existing tests (see {@code throwsNothing()}).
- * After specifiying the type of the expected exception your test is
+ * After specifying the type of the expected exception your test is
  * successful when such an exception is thrown and it fails if a
  * different or no exception is thrown.
  *
- * <p>
- * Instead of specifying the exception's type you can characterize the
- * expected exception based on other criterias, too:
```
eamonnmcmanus: Couldn't resist fixing "criterias" while I was here.

src/main/java/org/junit/rules/ExpectedException.java
eamonnmcmanus: Link tags without parameters do work, but only if there are no overloads. Since that would be a fragile state of affairs, I've added the parameters. I also followed your other suggestions, except for putting the new text at the end. If we're trying to encourage people to move to `assertThrows` then we probably don't want to bury that suggestion at the end of the text? Putting it at the top feels more like the effect we would get with `@deprecated`, though I admit it is a bit clunky.

sgz-xiaowu:
来自我的华为手机-------- 原始邮件 --------主题：Re: [junit] With ExpectedException, document that there should be no statements after the exceptional one. (#1199)发件人：蒩monn McManus 收件人：junit-team/junit 抄送：I hadn't picked up on expectThrows. In that case, I think using assertThrows/expectThrows is clearly better than using ExpectedException on Java 8 or later. With earlier versions of Java the verbosity of anonymous classes makes that less clear. So probably deprecation would be too strong. What do you think of the wording I've added to the beginning of ExpectedException?

src/main/java/org/junit/rules/ExpectedException.java
stefanbirkner: I agree with @eamonnmcmanus that the note about `assertThrows` should be at the top, because it is an important message for all Java 8 users.

src/main/java/org/junit/rules/ExpectedException.java
```diff
@@ -14,7 +14,14 @@
 
 /**
  * The {@code ExpectedException} rule allows you to verify that your code
- * throws a specific exception.
+ * throws a specific exception. Note that
+ * {@link org.junit.Assert#assertThrows(java.lang.Class, org.junit.Assert.ThrowingRunnable)
+ * Assert.assertThrows}
+ * is often a better choice since it allows you to express exactly where you
```
stefanbirkner: We should mention that this is only the case for Java 8. With Java < 8 you have to use anonymous classes that are hard to read.

eamonnmcmanus: OK, how about the text now?

PR #1201, opened 2015-09-17
Fix Samuraism's logo path

yusuke:


jlink:
I had already fixed it myself. Thanks for the contribution!

PR #1202, opened 2015-09-17
Skip symlinks in recursive temp-dir cleanup, and add test-case

choliver:
BUG: TemporaryFolder.delete() follows symlinks when recursing, which can lead to all sorts of havoc.

The equivalent bug was fixed in Apache IO here (http://svn.apache.org/viewvc/commons/proper/io/trunk/src/main/java/org/apache/commons/io/FileUtils.java?r1=661658&r2=684715&diff_format=h).  I'm not sure whether Files.isSymbolicLink() is sufficient here, or whether to just copy their implementation.

### DIFF
--- a/src/main/java/org/junit/rules/TemporaryFolder.java
+++ b/src/main/java/org/junit/rules/TemporaryFolder.java
@@ -4,6 +4,7 @@ import static org.junit.Assert.fail;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
 
 import org.junit.Rule;
 
@@ -268,7 +269,9 @@ public class TemporaryFolder extends ExternalResource {
         File[] files = file.listFiles();
         if (files != null) {
             for (File each : files) {
-                result = result && recursiveDelete(each);
+                if (!Files.isSymbolicLink(each.toPath())) {
+                    result = result && recursiveDelete(each);
+                }
             }
         }
         return result && file.delete();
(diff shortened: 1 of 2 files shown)


choliver:
I'm not sure what to do about the animal-sniffer NPEs, I haven't used this before.  Will look into it, but if anyone could advise that would be great!

choliver:
Looks like it's the same issue as here: https://github.com/mojohaus/animal-sniffer/issues/8.

marcphilipp:
The Files class is not available on JDK 6.

choliver:
@marcphilipp - Aha.  I'm a first-time JUnit contributor; so could you advise on whether licensing issues will affect our ability to copy-and-paste (etc.) from Apache Commons vs. having to reinvent the wheel?

marcphilipp:
Well, technically, when you copy code from Apache Commons it is licensed under a different licence (Apache License, I assume) than JUnit. So we needed to ship that part of the code under that licence, too. That would make things complicated and we would only consider it if it were a really important feature.

Taking a step back: You create symlinks in a TemporaryFolder?

choliver:
Not intentionally, but yes.  I spent the best part of a day this week trying to understand why my integration test (that spins up a Python virtualenv) was destroying my local Python installation.

One would assume that "delete folder" would have the same semantics as "rm -rf", so the current behaviour is both surprising and dangerous.  So I'd consider this a bugfix rather than a feature.

(That said, I'm somewhat surprised that I seem to be the first to hit this issue!)

kcooney:
I am closing this in favor of #1303

Issue #1194, opened 2015-08-26
Truely random test order


cardil:
By default junit executes test in default order that is based on method name hash code. In java hashcodes of string doesn't change between executions, so the order is random, but constant for given computer. That could lead to unpredictable errors.

To fix this enchantment, there should be truly random order option to explicitly specify in test case using `@FixMethodOrder`.

This will be similar to Rspec `--order random` option. Ref: http://blog.davidchelimsky.net/blog/2012/01/04/rspec-28-is-released/

Jahhein:
Might I suggest knuth shuffle for determining run order?
info-
https://en.wikipedia.org/wiki/Fisher–Yates_shuffle

source code-
http://algs4.cs.princeton.edu/11model/Knuth.java.html

ffbit:
From [Test execution order](https://github.com/junit-team/junit/wiki/Test-execution-order)
> By design, JUnit does not specify the execution order of test method invocations. Until now, the methods were simply invoked in the order returned by the reflection API. However, using the JVM order is unwise since the Java platform does not specify any particular order, and in fact JDK 7 returns a more or less random order. Of course, well-written test code would not assume any order, but some do, and a predictable failure is better than a random failure on certain platforms.
> 
> From version 4.11, JUnit will by default use a deterministic, but not predictable, order (MethodSorters.DEFAULT). To change the test execution order simply annotate your test class using @FixMethodOrder and specify one of the available MethodSorters:

> @FixMethodOrder(MethodSorters.JVM): Leaves the test methods in the order returned by the JVM. This order may vary from run to run.

> @FixMethodOrder(MethodSorters.NAME_ASCENDING): Sorts the test methods by method name, in lexicographic order.

Endron:
The problem here is that there are actually two test smells that are interconnected with each other:
1. Tests depending on the execution order
2. Non repeatable tests

Putting tests into a really random order would help with the first. As no test can predict what happened before he can not depend on any other test to have been executed before. By changing the order with each run you would also find tests accidentally depending on the execution order. 

[...]

PR #1204, opened 2015-09-27
Added "RANDOM" method sorting method.

mrnichols76:
Closes issue #1194.

This commit creates an additional method sorting option called "RANDOM," which uses a random number generator (java.util.Random) to cause JUnit to execute tests in random order. It also adds another annotation, "@Seed," which allows the random number generator's seed to be set to a pre-defined value (so a previous random ordering can be applied repeatedly). If no "@Seed" annotation is included, JUnit will generate a seed. In any event, if "RANDOM" ordering is applied, the seed used will be written to stdout so that a failed randomly ordered test run (even one using a JUnit-supplied seed) can be re-run in the same order.

The code uses the Knuth/Fisher-Yates shuffle algorithm to do the reordering.

### DIFF
--- a/src/main/java/org/junit/internal/MethodSorter.java
+++ b/src/main/java/org/junit/internal/MethodSorter.java
@@ -1,12 +1,24 @@
 package org.junit.internal;
 
+import org.junit.FixMethodOrder;
+import org.junit.Seed;
+import org.junit.runners.MethodSorters;
+
 import java.lang.reflect.Method;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Comparator;
-
-import org.junit.FixMethodOrder;
+import java.util.Random;
 
 public class MethodSorter {
+
+    /* 
+     * Used for java.util.Random seed generation.
+     * This is the same magic number used in the Java API
+     * for java.util.Random.
+     */
+    private static volatile long seedUniquifier = 8682522807148012L;
+
     /**
      * DEFAULT sort order
      */
@@ -49,15 +61,29 @@ public class MethodSorter {
      *      (non-)bug #7023180</a>
      */
     public static Method[] getDeclaredMethods(Class<?> clazz) {
-        Comparator<Method> comparator = getSorter(clazz.getAnnotation(FixMethodOrder.class));
+        FixMethodOrder order = clazz.getAnnotation(FixMethodOrder.class);
+        Comparator<Method> comparator = getSorter(order);
 
         Method[] methods = clazz.getDeclaredMethods();
         if (comparator != null) {
             Arrays.sort(methods, comparator);
         }
+        if(order != null && order.value() == MethodSorters.RANDOM) {
+            Seed s = clazz.getAnnotation(Seed.class);
+            long seed = s == null ? generateSeed() : s.value();
+            System.out.printf("Class: %s, Random ordering seed = 0x%xL%n", clazz.getName(), seed);
+            Random r = new Random(seed);
+            Collections.shuffle(Arrays.asList(methods), r);
+        }
 
         return methods;
     }
+
+    // Same seed generation algorithm used in JDK 6
+    private static long generateSeed() {
+        return ++seedUniquifier + System.nanoTime();
+    }
+
 
     private MethodSorter() {
     }
--- a/src/main/java/org/junit/runners/MethodSorters.java
+++ b/src/main/java/org/junit/runners/MethodSorters.java
@@ -1,9 +1,10 @@
 package org.junit.runners;
 
+import org.junit.internal.MethodSorter;
+
 import java.lang.reflect.Method;
 import java.util.Comparator;
 
-import org.junit.internal.MethodSorter;
 
 /**
  * Sort the methods into a specified execution order.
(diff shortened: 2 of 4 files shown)


kcooney:
Thanks for the contribution.

While some people might want the random seed specified in code, others would prefer to specify it via an environment variable or system property. Still others would want the random seed to be fixed for some period of time (perhaps a day) and change at regular intervals, so if a particular order causes a failure you have time to diagnose the cause and fix it.

IMHO, a random order that changes every time you run the test is quite problematic. It would be easy to be in the middle of a coding session, run the tests, notice the failure, and start undoing changes trying to find out what you did to "break" the tests, only to discover later that the breakage had nothing to do with your code.

For these changes and many more, I would strongly prefer an API that would allow people to shuffle tests that requires that the developer (or a third-party library) to specify the ordering. See, for example, pull #1130

src/main/java/org/junit/runners/MethodSorters.java
```diff
@@ -25,6 +26,13 @@
     JVM(null),
 
     /**
+     * Sorts the test methods in pseudo-random order.
+     * This sort option generates and displays (to stdout) a
+     * seed value for the random number generator.
+     */
+    RANDOM(null),
```
kcooney: BTW, if you want the ordering provided by `RANDOM` to be consistent when you pass the same seed, then you'll need to actually order the methods in a deterministic way before you do the shuffling.

src/main/java/org/junit/internal/MethodSorter.java
```diff
@@ -49,16 +60,49 @@ public int compare(Method m1, Method m2) {
...
+     * This seed is then displayed on stdout.
+     * @param methods the array of methods
+     * @param className the name of this class
+     */
+    private static void randomize(Method[] methods, String className, 
+        long seed) {
+        System.out.printf("Class: %s, Random ordering seed = 0x%xL%n", 
+            className, seed);
+        Random r = new Random(seed);
+        for(int i=methods.length-1; i>0; i--) {
```
kcooney: BTW, you could instead do:

  `Collections.shuffle(Arrays.asList(methods), r);`

src/test/java/org/junit/internal/MethodSorterTest.java
```diff
@@ -179,4 +182,98 @@ public void testAscendingMethodSorter() {
...
+
+        void delta() {
+        }
+
+        void epsilon() {
+        }
+    }
+
+    @Test
+    public void testRandomMethodSorter() {
```
kcooney: This test will occasionally fail

src/main/java/org/junit/internal/MethodSorter.java
mrnichols76: Yes. That's much more readable. I've used shuffle before, but I didn't realize there was a version that allowed you to supply your own Random.

src/main/java/org/junit/runners/MethodSorters.java
mrnichols76: I just pushed up a commit that fixes this (just had to change line 33 to use the DEFAULT ordering--since in MethodSorter, the existing line to do the sort is before the shuffle takes place). Thanks for the comments, by the way.

src/test/java/org/junit/internal/MethodSorterTest.java
mrnichols76: You mean if the random shuffling happens to not change the ordering at all?

kcooney:
@mrnichols76 Just to be clear, I commented on some of the smaller issues in the code, but I still personally feel that this is not the best approach for solving this issue.

@marcphilipp what are your thoughts?

mrnichols76:
@kcooney The way you proposed in pull #1130 is more general/flexible (and I must admit that I had not seen it until you pointed it out to me). It's better in a lot of ways.

However this pull (1204) has the virtue of providing the requested randomization feature with minimal changes to the existing codebase, so I think it's still worth considering.

marcphilipp:
I'm afraid I won't have time to look at this and #1130 in depth before this weekend.

cardil:
I think that passing a seed as a `@Seed` annotation is not ok, it will require to change the code. The code should be unchanged especially if one is hunting a dependency bug in unit test. I prefer to use system option like rspec does. I was referring here https://github.com/junit-team/junit/issues/1194#issuecomment-143024939

It will also should be settable by Maven and Gradle plugins, to be usable by developers.

Second thing is that I think that random method order sorting is ok, but I that most of dependency bugs in tests are between different test classes. It would be nice that single seed can be reused by all test classes in project and to set random method sorting as default.

Maybe better way would be to give Maven and Gradle chance to set seed externally?
What do you think?

mrnichols76:
@cardil I like the idea of passing it in externally.  As you say, it would be more useful if you could apply the same seed to multiple files. I hadn't thought about that when I wrote my code, mostly because I was thinking about the way that my students and I use JUnit (I teach Java programming). Do you think it would be better to use an environment variable for this or a command-line option? (or maybe both?)

I still think there ought to be an option to allow someone to put the seed in a test's annotation, though, because some users of IDEs may not want to bother tinkering with the details of build files or environment variables. Having a "@Seed" annotation is like adding "@Ignore" tag to a test--you can switch it on or switch it off. Yes, it changes the test's source file, but it doesn't really change the code of the test.

kcooney:
These kinds of reasonable debates on how to pick and override the seed are exactly why I would prefer an extension point. Let third-party library developers design an API and work with integration with build tools (for selecting and overriding the seed, as well as possibly including the seed in the test report).

kcooney:
I propose we close this. I still stand by my concerns I stated earlier.

marcphilipp:
Agreed.

PR #1209, opened 2015-10-15
createTestClass with throws InitializationError

vorburger:
I'm @Override createTestClass() and in my implementation things could go wrong and it would thus be handy if createTestClass had with throws InitializationError, just like the called constructor ParentRunner anyway already does.


kcooney:
This change would break any other custom runners that call this method.

PR #1214, opened 2015-10-27
Add evaluatePostcheck method to ExpectedException rule.

PiotrSliwa:
# Problem
Hi *,

there is a known issue with JUnit testing. Sometimes there is a need for some assertions after an expected exception is thrown by tested method. To visualize the problem let's consider below snippet:

```java
class Dummy {
    
    private boolean called = false;

    public boolean hasMethodBeenCalled() {
        return called;
    }
    
    public void throwingMethod() throws Exception {
        called = false;   // Lol, we made a bug here - let's cross figers our unit test catch it
        throw new Exception();
    }
}

public class TestedTest {

    private final Dummy sut = new Dummy();

    @Rule
    public ExpectedException expectedException = ExpectedException.none();
    
    @Test
    public void shallThrowException() throws Exception {
        expectedException.expect(Exception.class);
        sut.throwingMethod();
        assertTrue(sut.hasMethodBeenCalled());  // Shoot, it did not - the test passes :c
    }
}
```

This is certainly what one could expect - test's flow never reaches the assertion as the preceding method throws an exception breaking it. We can assume that a check for `sut.hasMethodBeenCalled()` is somehow crucial and needs to be tested. In order to do so, a developer is forced to modify his test. One of the options is to use `try {} finally {}` expression:

```java
@Test
public void shallThrowException() throws Exception {
    expectedException.expect(Exception.class);
    try {
        sut.throwingMethod();
    }
    finally {
        assertTrue(sut.hasMethodBeenCalled());
    }
}
```

It works but is highly illegible resulting in a message:

```
Expected: an instance of java.lang.Exception
     but: <java.lang.AssertionError> is a java.lang.AssertionError
```
(or `Unexpected exception, expected<java.lang.Exception> but was<java.lang.AssertionError>` when using `@Test(expected = ...)` annotation instead of `ExpectedException` which is hardly readable whatsoever).

Thus, it forces to adopt longer `try {} catch {}` with `fail` call after `sut`:

```java
@Test
public void shallThrowException() {
    try {
        sut.throwingMethod();
        fail("Expected exception has not been thrown!");
    }
    catch (Exception ex) {
        assertTrue(sut.hasMethodBeenCalled());
    }
}
```

which finally works as intended. However, it means that a programmer has to implement it each time by his own and it's basically what `ExpectedException` is for.

# Solution
In these commits I provided `evaluatePostcheck` method to `ExpectedException`. The considered test will look like this now:

```java
@Test
public void shallThrowException() throws Exception {
    expectedException.expect(Exception.class);
    expectedException.evaluatePostcheck(() -> {
        assertTrue(sut.hasMethodBeenCalled());  // Java 8 lambda can be used as well
    });
    sut.throwingMethod();
}
```

`evaluatePostcheck` accepts `Runnable` interface as an argument. It is invoked after JUnit verifies thrown exception allowing developer to easily perform checks/assertions of the tested class's state despite broken flow.

Such a check is especially useful when mocks are used - often it's necessary to check whether a mock's method is called before the tested method throws.

### DIFF
--- a/src/main/java/org/junit/rules/ExpectedException.java
+++ b/src/main/java/org/junit/rules/ExpectedException.java
@@ -248,10 +248,11 @@ public class ExpectedException implements TestRule {
      * public void throwsExceptionWithSpecificType() {
      *     thrown.expect(NullPointerException.class);
      *     thrown.evaluatePostcheck(new Runnable() {
-     *         @Override
+     *         &#064;Override
      *         public void run() {
-     *             // additional checks (e.g. whether a mock was called before
-     *             // the test threw the exception)
+     *             // Additional checks (e.g. whether a mock was called before
+     *             // the test threw an exception). This will be called after
+     *             // the exception is caught and verified.
      *         }
      *     });
      *     throw new NullPointerException();
--- a/src/main/java/org/junit/rules/ExpectedException.java
+++ b/src/main/java/org/junit/rules/ExpectedException.java
@@ -115,6 +115,14 @@ public class ExpectedException implements TestRule {
 
     private String missingExceptionMessage= "Expected test to throw %s";
 
+    private Runnable postThrowCheck = EmptyCheck.instance;
+
+    private static class EmptyCheck implements Runnable {
+        public static final EmptyCheck instance = new EmptyCheck();
+        private EmptyCheck() { }
+        public void run() { }
+    }
+
     private ExpectedException() {
     }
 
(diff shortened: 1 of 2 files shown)


marcphilipp:
Have you looked at Assert#assertThrows in current master?

PiotrSliwa:
No, to be honest - I have not seen it yet. It partially solves the problem, however as I can see it still does not allow to verify exception's contents (e.g. message and cause) so if one wants to do it, he/she still needs to use `ExpectedException` rule and the feature can be useful. Nevertheless, the `assertThrows` looks really nice.

kcooney:
`Assert.expectThrows()` returns the thrown exception, which would allow you to check the message and cause.

PiotrSliwa:
Right, I can see it now. So, I guess this pull request depends on politics applying to exception testing by JUnit - is `@Rule ExpectedException` still going to be maintained (in which case this commit seems to be useful) or will be deprecated by introduction of `Assert.expectThrows`.

kcooney:
iMHO it is easy to use `ExpectedEcception` incorrectly. Too often I have seen code that doesn't make sense (especially if the use of the Rule gets moved to a helper method.

stefanbirkner:
`ExpectedException` is still maintained, but IMHO it should not be extended by features that are designed for using Lambda expressions. That's because with Java 8 there are better ways for handling exceptions in test. e.g. `Assert.expectThrows()`

kcooney:
+1 on Stefan's comment.

@PiotrSliwa does `Assert.expectThrows` meet your needs? If so, I think we should not merge this pull.

PiotrSliwa:
I agree with your points, plus `Assert.expectThrows` works for me, therefore I think we can close the pull request. Thank you for your comments:)

Issue #1127, opened 2015-04-19
Show contents actual array when array lengths differ.


sf105:
assertArrayEquals() will short-circuit and only report lengths if they differ. I find this insufficient for easy diagnosis. Can the assertion show the actual result? Happy to provide a pull request if this is accepted.

marcphilipp:
I was about to ask whether you have considered using Hamcrest… ;-)

Having said that, I would support such an enhancement.

sf105:
I did :), but I was doing some training and didn't want to pull in the extra library.

PostalBear:
Hi folks, if you don't mind I would propose these pull requests. 
I used feature of Hamcrest matcher to print content of array if length is different, as it was proposed by @marcphilipp.

PS sorry for duplicated pull request, I had problem with my Git :(

kcooney:
@PostalBear I see the commits but no pull request.

I don't think Marc was suggesting that the fix in JUnit would use Hamcrest; I think he was wondering if Hamcrest  ave the error message that the original poster expected.

FYI, we will most likely be removing dependencies from JUnit to Hamcrest, so if we change JUnit to resolve this issue, we should find a fix that doesn't use Hamcrest APIs.

PostalBear:
@kcooney I got your point, indeed it is possible to achieve what original poster wanted by using Hamcrest matcher (since I did it same way). 
In such a case I will take a look to other low priority issues, instead of dealing with this one. 

PS What do you think about closing this issue ? I'm asking cause it could be easily solved without any modifications of JUnit code at all.

kcooney:
I don't have a strong feeling for whether or not we should change the current behavior. If we do, I think we should do it in the junit5 branch.

We did do some optimizations on this code in 4.12, so it's possible the JUnit version is faster than the Hamcrest one.

atulagrawal:
PR for this issue has been merged. Can you please close this issue. Thanks.
Ref - https://github.com/junit-team/junit4/pull/1300

marcphilipp:
Thanks, @atulagrawal!

PR #1216, opened 2015-10-31
A sort of fix for issue #1127

tomwhoiscontrary:
In issue #1127, @sf105 noted that when comparing two arrays which differ in length, assertArrayEquals() will only report that they differ in length, which he found insufficient for easy diagnosis. As a fix, he suggested printing the complete actual array contents.

Rather than doing that, this PR changes assertArrayEquals() to do the usual array comparison even when arrays differ in length, producing a failure message which combines the difference in length and the first difference in content. I think this should ease diagnosis, as @sf105 wanted, and it doesn't require making big changes to the code, adding Hamcrest, introducing a new behaviour (JUnit doesn't print complete array contents anywhere else), or blowing people's IDEs up when they compare multi-megabyte byte arrays.

### DIFF
--- a/src/main/java/org/junit/internal/ComparisonCriteria.java
+++ b/src/main/java/org/junit/internal/ComparisonCriteria.java
@@ -70,8 +70,8 @@ public abstract class ComparisonCriteria {
         }
 
         if (actualsLength != expectedsLength) {
-            Object expected = getArrayElementOrSentinel(expecteds, expectedsLength, prefixLength);
-            Object actual = getArrayElementOrSentinel(actuals, actualsLength, prefixLength);
+            Object expected = getToStringableArrayElement(expecteds, expectedsLength, prefixLength);
+            Object actual = getToStringableArrayElement(actuals, actualsLength, prefixLength);
             try {
                 Assert.assertEquals(expected, actual);
             } catch (AssertionError e) {
@@ -80,15 +80,37 @@ public abstract class ComparisonCriteria {
         }
     }
 
-    private static final Object END_OF_ARRAY_SENTINEL = new Object() {
-        @Override
-        public String toString() {
-            return "end of array";
-        }
-    };
+    private static final Object END_OF_ARRAY_SENTINEL = objectWithToString("end of array");
 
-    private Object getArrayElementOrSentinel(Object array, int length, int index) {
-        return index < length ? Array.get(array, index) : END_OF_ARRAY_SENTINEL;
+    private Object getToStringableArrayElement(Object array, int length, int index) {
+        if (index < length) {
+            Object element = Array.get(array, index);
+            if (isArray(element)) {
+                return objectWithToString(componentTypeName(element.getClass()) + "[" + Array.getLength(element) + "]");
+            } else {
+                return element;
+            }
+        } else {
+            return END_OF_ARRAY_SENTINEL;
+        }
+    }
+
+    private static Object objectWithToString(final String string) {
+        return new Object() {
+            @Override
+            public String toString() {
+                return string;
+            }
+        };
+    }
+
+    private String componentTypeName(Class<?> arrayClass) {
+        Class<?> componentType = arrayClass.getComponentType();
+        if (componentType.isArray()) {
+            return componentTypeName(componentType) + "[]";
+        } else {
+            return componentType.getName();
+        }
     }
 
     private boolean isArray(Object expected) {
--- a/src/main/java/org/junit/internal/ComparisonCriteria.java
+++ b/src/main/java/org/junit/internal/ComparisonCriteria.java
@@ -34,10 +34,22 @@ public abstract class ComparisonCriteria {
         }
         String header = message == null ? "" : message + ": ";
 
-        int expectedsLength = assertArraysAreSameLength(expecteds,
-                actuals, header);
+        if (expecteds == null) {
+            Assert.fail(header + "expected array was null");
+        }
+        if (actuals == null) {
+            Assert.fail(header + "actual array was null");
+        }
 
-        for (int i = 0; i < expectedsLength; i++) {
+        int actualsLength = Array.getLength(actuals);
+        int expectedsLength = Array.getLength(expecteds);
+        if (actualsLength != expectedsLength) {
+            header = header + "array lengths differed, expected.length="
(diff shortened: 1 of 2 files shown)


src/main/java/org/junit/internal/ComparisonCriteria.java
```diff
@@ -56,27 +68,31 @@ public void arrayEquals(String message, Object expecteds, Object actuals)
...
         }
-    }
 
-    private boolean isArray(Object expected) {
-        return expected != null && expected.getClass().isArray();
+        if (actualsLength != expectedsLength) {
+            Object expected = getArrayElementOrSentinel(expecteds, expectedsLength, prefixLength);
+            Object actual = getArrayElementOrSentinel(actuals, actualsLength, prefixLength);
+            try {
+                Assert.assertEquals(expected, actual);
```
kcooney: The previous code just used `Assert.fail()` here. Any reason we should do something different? I seen to recall this code being a bit sensitive when handling nested arrays.

src/test/java/org/junit/tests/assertion/AssertionTest.java
```diff
@@ -83,11 +98,29 @@ public void arraysActualNullMessage() {
     }
 
     @Test
-    public void arraysDifferentLengthMessage() {
+    public void arraysDifferentLengthDifferingAtStartMessage() {
+        try {
+            assertArrayEquals("not equal", (new Object[]{true}), (new Object[]{false, true}));
+        } catch (AssertionError exception) {
+            assertEquals("not equal: array lengths differed, expected.length=1 actual.length=2; arrays first differed at element [0]; expected:<true> but was:<false>", exception.getMessage());
+        }
```
kcooney: Before this line add a `return` statement, then after the `catch` block call `fail()` so we know an exception was thrown.

This will be so much easier if we could use lamdas and `Assert.expectThrows()` ...

kcooney:
I really like this solution!

Can you add tests to verify the failure message for arrays  of arrays where the second nested array has differing lengths?

tomwhoiscontrary:
Will do. I did think briefly about the nested arrays case, and i am 70% sure it will just work. It should definitely have tests, though!

src/test/java/org/junit/tests/assertion/AssertionTest.java
tomwhoiscontrary: Funny you should say that. I initially wrote the tests in the way you suggest, then noticed that none of the other tests in this class have that pattern. Rather, for every message test, there is a corresponding failure test that checks that the assertion really does fail. On the one hand, that does leave the message tests brittle, but on the other hand, it means each test only tests one thing - either the fact that the assertion fails, or the message it produces.

My tests follow the existing pattern. I'm happy to change them to break the pattern, but perhaps it would be better to refactor all of the tests in this class to work that way.

src/main/java/org/junit/internal/ComparisonCriteria.java
tomwhoiscontrary: @kcooney I need to do an assertEquals to get the right failure message. I could use fail, but then i would have to construct the failure message here, and that would duplicate the code in assertEquals.

kcooney: I will let @marcphilipp decide, but if the previous code used fail with a hard-coded message, then I think we should do the same here, to minimize the observable change on behavior.

src/test/java/org/junit/tests/assertion/AssertionTest.java
kcooney: I saw that, too. Although I do like the idea of a test only testing one thing, I don't like tests that can pass for the wrong reason. Also, the way the tests are structured, you have to write two tests for each use case (violating Don't Repeat Yourself). Finally, if the tests were written the way I suggest, they would only be testing one thing: verifying that with the given input, `assertArrayEquals()` throws the expected exception.

`Assert.expectThrows ()` would be perfect here if we could use lamdas. We could use it here without lamdas, or we could add the `return` and `fail()` calls.

tomwhoiscontrary:
Okay, i've changed the tests as suggested - being more careful to verify failure in each case, and adding tests for multidimensional arrays.

The testing did bring up a case where the message is poor:

```
        Object[] expecteds = new Object[][]{{}, new Object[]{true}};
        Object[] actuals = new Object[][]{{}};
        assertArrayEquals(expecteds, actuals);
```

The message is:

```
array lengths differed, expected.length=2 actual.length=1; arrays first differed at element [1]; expected:<[Ljava.lang.Object;@2b71fc7e> but was:<end of array>
```

Using the toString of an array in the message is not really very useful (although still better than what you get now, which is nothing). But i'm not sure what would be better, or how to make that happen.

tomwhoiscontrary:
Okay! @kcooney, i think i've got good failure messages for all kinds of arrays now. Where the mismatching element is itself an array, the element comparison part of the message looks like:

```
// primitive array:
... expected:<int[1]> but was:<end of array>
// object array:
... expected:<java.lang.Object[1]> but was:<end of array>
// multidimensional array:
... expected:<java.lang.Object[][1]> but was:<end of array> // multidimensional array
```

That is, the type of the array, and its length. That should give users a bit of a clue as to where the arrays are different, without overwhelming them with detail.

I played around with dumping the complete array contents, but it requires a disproportionate amount of code, and, again, risks blowing people's IDEs up when they compare *arrays which contain* multi-megabyte byte arrays!

Apologies for the delay, i was hibernating. Can you (or anyone else) suggest any further improvements?

marcphilipp:
I am closing this pull request because it targets the `junit5` branch which we will delete in the next few days. Please open a pull request targeting `master` instead.

tomwhoiscontrary:
@marcphilipp I'll happily raise a new PR against master if there's a chance that it will actually be merged.

Would it be okay to make the PR by merging this into master, or would you prefer a completely new branch that doesn't involve the junit5 branch in its history?

marcphilipp:
@junit-team/junit-committers Do you guys have objections to such a pull request for master?

kcooney:
@marcphilipp No objections

marcphilipp:
@tomwhoiscontrary Feel free to provide a new PR against master. Can you rebase your commits on master first?

tomwhoiscontrary:
Will do.

PR #1217, opened 2015-11-03
Add doc: Don't use ExcpectedException with Java 8.

stefanbirkner:
The additional documentation has two objectives.

* Users should be aware that there is a better alternative for Java 8.
* Avoid issues for adding Lambda based features to this rule. See #1214.


kcooney:
Instead should we merge https://github.com/junit-team/junit/pull/1199 ?

stefanbirkner:
You're right. I didn't remember that pull request.

Issue #1078, opened 2015-01-31
@Test annotations on interface default methods


nwillc:
I was hoping to make and interface with @Test annotated default methods that my tests could implement.  Didn't appear to see these tests.

kcooney:
JUnit currently requires JDK5. Most likely, JUnit 5.0 would require JDK 6. We don't want to require more recent JDKs, because many people can't use the latest JDK (including people on Android).

Can you do this without default methos?

josephw:
`@Test` isn't marked as [`@Inherited`](http://docs.oracle.com/javase/8/docs/api/java/lang/annotation/Inherited.html); that might be all that's necessary to implement this, but it would be a change in behaviour.

kcooney:
Not sure if making `@Test` inherited would break people. What happens today if you have a base class with an abstract method annotated with `@Test` and you override it in a subclass without annotating the overridden.method?

nwillc:
Inheriting from a base abstract works today.  It's how I implemented
http://nwillc.github.io/jdk_contract_tests

marcphilipp:
@kcooney The method is run as a test regardless whether it is annotated with `@Test` in the subclass or not.

marcphilipp:
@josephw Adding `@Inherited` won't help since JUnit only scans super classes for methods with annotations:
https://github.com/junit-team/junit/blob/master/src/main/java/org/junit/runners/model/TestClass.java#L64

marcphilipp:
We'd probably need to scan interfaces for test methods, too.

Stephan202:
Not 100% sure about Java 8's new *default methods* on interfaces, but the documentation for [`@Inherited`](https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/Inherited.html) states:

> Note that this meta-annotation type has no effect if the annotated type is used to annotate anything other than a class. Note also that this meta-annotation only causes annotations to be inherited from superclasses; annotations on implemented interfaces have no effect.

See also [this SO question](https://stackoverflow.com/questions/4745798/why-java-classes-do-not-inherit-annotations-from-implemented-interfaces). Would be good to verify how default methods are treated.

Stephan202:
Wait, I just realized that you want subclasses to implement the tests, without having to annotate them as a `@Test`. In that scenario default methods of course don't apply. So in conclusion: `@Inherited` won't help here.

Stephan202:
[...]

PR #1222, opened 2015-11-22
Make JUnit look for `@Test` annotated methods in interfaces as well

chrisvest:
Previously, the JUnit4 runner looked for `@Test` annotated methods to run, in the concrete test class, and in all of its super-classes.

This change makes JUnit also look for `@Test` annotated methods in all the interfaces that a test class implements.

For instance, the following now works, in that the `abstractBehaviour()` test method is now executed as part of the `ConcreteTest`:

```java
    public interface AbstractTests {
        @Test
        public void abstractBehaviour();
    }
    public class ConcreteTest implements AbstractTests {
        public void abstractBehaviour() {
            // assertions...
        }
    }
```

This is particularly useful on Java8, because it allows us to use interfaces as test mix-ins:

```java
    public interface CollectionTestFixture {
        Collection<Object> create();
    }
    public interface CollectionAddition extends CollectionTestFixture {
        @Test
        default void mustContainAddedItem() {
            Object obj = new Object();
            Collection<Object> coll = create();
            coll.add(obj);
            assertTrue(coll.contains(obj));
        }
    }
    public interface CollectionRemove extends CollectionTestFixture {
        @Test
        default void removeOfAddedItemMustSucceed() {
            Object obj = new Object();
            Collection<Object> coll = create();
            coll.add(obj);
            assertTrue(coll.remove(obj));
        }
    }
    public class ArrayListTest implements CollectionAddition, CollectionRemove {
        @Override
        public Collection<Object> create() {
            return new ArrayList<>();
        }
    }
    public class LinkedListTest implements CollectionAddition, CollectionRemove {
        @Override
        public Collection<Object> create() {
            return new LinkedList<>();
        }
    }
```

Of course, the above example is most likely better written as a parameterised test, but there might be other cases where this mix-in pattern is more appropriate.

### DIFF
--- a/src/main/java/org/junit/runners/model/TestClass.java
+++ b/src/main/java/org/junit/runners/model/TestClass.java
@@ -10,6 +10,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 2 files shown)


Tibor17:
@junit-team/junit-committers
This seems to be very desirable.
I could use this in parameterized tests.
Still the test classes in JUnit do not make sense for OOP. If we have trivial tests witout any instance fields (possibly rule annotation on method) then we need only interface. The JUnit can make Proxy (Java 1.3) and instantiate or aggregate several interfaces like in this example.
Just random thinking...
When you want to release 4.13 with this PR?

kcooney:
FYI: this idea was discussed in #1078 

I suggest reading that discussion, and adding your thoughts there

chrisvest:
@kcooney Thoughts added.

PR #1227, opened 2015-12-07
Catch SecurityException in addition to IOException.

### LINKED ISSUE #1213: java.security.AccessControlException: access denied ("java.util.PropertyPermission" "user.home" "read") running under security manager, because JUnit forgot to use doPrivileged
If you perform actions which cause permission checks with the security manager (and you're doing some action which is not exploitable by the caller), then you're supposed to use doPrivileged.

`BaseTestRunner` in JUnit has this:

        private static File getPreferencesFile() {
            String home = System.getProperty("user.home");  // <- not using doPrivileged
            return new File(home, "junit.properties");
        }

When running our test suite with the security manager enabled, we get a failure:

    Caused by: java.security.AccessControlException: access denied ("java.util.PropertyPermission" "user.home" "read")
        at java.security.AccessControlContext.checkPermission(AccessControlContext.java:457)
        at java.security.AccessController.checkPermission(AccessController.java:884)
        at java.lang.SecurityManager.checkPermission(SecurityManager.java:549)
        at java.lang.SecurityManager.checkPropertyAccess(SecurityManager.java:1294)
        at java.lang.System.getProperty(System.java:717)
        at junit.runner.BaseTestRunner.getPreferencesFile(BaseTestRunner.java:225)
        at junit.runner.BaseTestRunner.readPreferences(BaseTestRunner.java:232)
        at junit.runner.BaseTestRunner.getPreferences(BaseTestRunner.java:51)
        at junit.runner.BaseTestRunner.getPreference(BaseTestRunner.java:246)
        at junit.runner.BaseTestRunner.getPreference(BaseTestRunner.java:250)
        at junit.runner.BaseTestRunner.<clinit>(BaseTestRunner.java:324)

This occurs irrespective of the fact that our security policy allows reading any system property, because something further down the stack, i.e. something inside IDEA's JUnit launcher (perhaps some dynamically-generated bytecode?) is "completely" untrustworthy and thus even the "grant to all code sources" section of the policy does not apply to it.

At the moment, our workaround for this is to give AllPermission to the IDEA installation, but this is not really satisfactory because every developer tends to install it in a different location depending on their own conventions, what platform they're on, whether they have admin access on the box, etc.

If JUnit would add a doPrivileged block here (and to any other place where it seems appropriate) then we wouldn't have to do this and our rule which says JUnit is completely trusted would be sufficient.

trejkaz:
Fixes #1213, not by adding the doPrivileged, but by coping with lack of access to the file by pretending it doesn't exist.

### DIFF
--- a/src/main/java/junit/runner/BaseTestRunner.java
+++ b/src/main/java/junit/runner/BaseTestRunner.java
@@ -233,6 +233,7 @@ public abstract class BaseTestRunner implements TestListener {
             setPreferences(new Properties(getPreferences()));
             getPreferences().load(is);
         } catch (IOException ignored) {
+        } catch (SecurityException ignored) {
         } finally {
             try {
                 if (is != null) {


kcooney:
@trejkaz could you please update the release notes (in the wiki) for 4.13 to document this?

trejkaz:
Done. Hopefully that entry is good enough.

PR #1230, opened 2015-12-21
capture log4j logs rule

sapirgolan:
Adding a new rule that capture all logs that were written using log4j

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,6 +113,12 @@
             <artifactId>hamcrest-core</artifactId>
             <version>1.3</version>
         </dependency>
+        <dependency>
+            <groupId>log4j</groupId>
+            <artifactId>log4j</artifactId>
+            <version>1.2.17</version>
+            <optional>true</optional>
+        </dependency>
     </dependencies>
 
     <build>
--- a/src/main/java/org/junit/rules/Logging.java
+++ b/src/main/java/org/junit/rules/Logging.java
@@ -1,0 +1,131 @@
+package org.junit.rules;
+
+import org.apache.log4j.AppenderSkeleton;
+import org.apache.log4j.Level;
+import org.apache.log4j.Logger;
+import org.apache.log4j.spi.Filter;
+import org.apache.log4j.spi.LoggingEvent;
+import org.apache.log4j.varia.LevelRangeFilter;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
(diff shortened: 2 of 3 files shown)


kcooney:
Thanks for the contribution

We generally try to avoid adding any external dependencies to JUnit. Hamcrest was an exception, and that caused all kinds of problems.

sapirgolan:
Hi kcooney,

I'm aware of that, I read your POM and saw only a single dependency, Hamcrest.
However, This depedency has the 'optioanl' property set, meaning that users have to specify it if they want to use it - https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html

If someone is already using log4J he will have the dependency in this mvn module

Tibor17:
@sapirgolan is right.
@kcooney log4j is needed only to compile junit. One way or another, it is guaranteed that this rule is used by log4j user who already owns log4j dependency in his classpath; otherwise `ClassNotFound` is thrown. The normal user who does not use this rule would not have class loading issue. I am only facing a problem in OSGi, for instance PAX; they should exclude package but I guess log4j is used everywhere, so no issue.

Tibor17:
hamcrest cannot be compared with log4j because hamcrest affected the everything, however log4j only this rule in the least invasive way.
@sapirgolan I am using slf4j. Can I use this rule as well?

sapirgolan:
@Tibor17 , slf4j is abstraction for various logging frameworks, where log4j is only one of them.
When using slf4j, you need to plug in the desired logging framework at deployment time.

If you plug in log4j the rule will work. But if you plug another framework, java.util.logging for example, it will not work.

kcooney:
Personally, I would lean against JUnit having even a compile-time dependency on a third-party logging framework. Third-party APIs change in incompatible ways.

Secondly, one of the core principles of JUnit is APIs over functionality. See "Prefer Extensions Over Features" at https://github.com/junit-team/junit-lambda/wiki/Core-Principles

stefanbirkner:
I would like to close this pull request. It does not fit into the scope of JUnit which is the core testing framework and the base for building elaborate support for testing other library. Therefore Rules have been introduced. Everyone can create a library with rules and publish it. E.g. I have done this with [System Rules](http://stefanbirkner.github.io/system-rules/).

marcphilipp:
I agree with @kcooney and @stefanbirkner.

Issue #1233, opened 2016-01-02
Jdk9 Early Access Build 99 issue


nhojpatrick:
Installed Java 9 Early Access Build 99, trying to build JUnit and identified issue with maven-jar-plugin 2.6. Will raise with maven-jar-plugin and then submit pull request with updates so JUnit builds with Java 9.

marcphilipp:
Please do, thanks for checking!

nhojpatrick:
Fixes awaiting release;
https://issues.apache.org/jira/browse/MJAR-206
https://issues.apache.org/jira/browse/MJAVADOC-443

Pull request into maven-plugin;
https://github.com/apache/maven-plugins/pull/76

nhojpatrick:
See pull request https://github.com/junit-team/junit/pull/1234

stefanbirkner:
The plugins are now working fine with JDK 9. The only problem currently is that `StackTracesTest` is failing. I have a look at it. For more details see the [builds for my jdk9 branch](https://travis-ci.org/stefanbirkner/junit/branches).

stefanbirkner:
JUnit can now be build with JDK 9. See 9b061ea8c96fa6cba0ac9d7cfd5e8ffbd030b34a.

PR #1234, opened 2016-01-03
Bugfix/jdk9 build fixes

nhojpatrick:
Linked to https://github.com/junit-team/junit/issues/1233

With Java 9, it drops support for Java 5 i.e. -source 1.5 and -target 1.5. So this pull request updates updates the min Java version needed to 1.6.

Two maven plugins, maven-jar-plugin and maven-javadoc-plugin also have issues with Java 9.

Once https://github.com/apache/maven-plugins/pull/76 is merged it will depending on a fixed version of plexus-archiver (see https://github.com/codehaus-plexus/plexus-archiver/pull/12).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -278,7 +278,7 @@
                 in jar archive target/junit-*-javadoc.jar.
                 -->
                 <artifactId>maven-javadoc-plugin</artifactId>
-                <version>2.10.3</version>
+                <version>2.10.4-SNAPSHOT</version>
                 <configuration>
                     <stylesheetfile>${basedir}/src/main/javadoc/stylesheet.css</stylesheetfile>
                     <show>protected</show>
@@ -337,7 +337,7 @@
             </plugin>
             <plugin>
                 <artifactId>maven-jar-plugin</artifactId>
-                <version>2.6</version>
+                <version>3.0.0-SNAPSHOT</version>
                 <configuration>
                     <archive>
                         <addMavenDescriptor>false</addMavenDescriptor>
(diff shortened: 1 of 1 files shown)


kcooney:
@nhojpatrick thanks for the fix! We don't want to change the required JDK version on the master branch, because master will be used for JUnit  4.13. I believe we changed the required JDK version on the junit5 branch. Is the bug reproducible there? If so, can you please redo these changes on that branch?

marcphilipp:
@kcooney The junit5 branch will most probably never be released since we're working on JUnit 5 (alpha) in the junit-lambda repository.

kcooney:
@marcphilipp I thought we agreed to review the changes on the junit5 branch?

In any case, we shouldn't upgrade the required JDK version in a 4.x release.

Luckily, JDK 9 isn't scheduled to be released until September, so we have time.

nhojpatrick:
Tried the junit-team:junit:junit5 branch but it uses older versions of the plugins than junit-team:junit:master does.

Checkout junit-team:junit-lambda:master but don't have gradle installed so will leave it for the time being.

kcooney:
We haven't done a merge from master to junit5 in a while. I personally see some value in some of the changes in the junit5 branch. Will discuss with the other maintainers off-thread

nhojpatrick:
Regarding the September comment, it does seam a long way away. But what about other projects that might want to use JUnit and also support Java 9... when Java 9 is generally available.

Let me know what is decided regarding junit5 and also Java 9, and when I've free time I'll try and help out.

kcooney:
@nhojpatrick unless I'm misunderstanding the changes, they only affect the ability to build JUnit; Java 9 users should be able to use the prebuilt 4.12 jars, correct? If so, I think we can wait until we can use non-snapshot jars of these plugins.

Tibor17:
@nhojpatrick javac 1.6 was here discussed several times. JUnit4 will never have it for sure.
Regarding JDK 9 this does not touch users because ordinal jar will work in JDK9 and therefore we do not need JDK9 to build JUnit. I think Travis is good enough with JDK 7/8 until JUnit4 has died.

kcooney:
since the jars can still be used, I vote we close this.

marcphilipp:
Agreed.

PR #1261, opened 2016-03-18
made some refactor

fcalderonnearsoft:
Using implementations types like `ArrayList` are not recommended, changed to interface `List`.

In `BaseTestRunner` it looks cleaner `!fgFilterStack` than `fgFilterStack == false`

### DIFF
--- a/src/main/java/junit/runner/BaseTestRunner.java
+++ b/src/main/java/junit/runner/BaseTestRunner.java
@@ -298,7 +298,7 @@ public abstract class BaseTestRunner implements TestListener {
     }
 
     protected static boolean showStackRaw() {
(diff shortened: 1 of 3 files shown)


kcooney:
Thanks for the contribution.

Unfortubately, changing the method signature of public methods could break callers.

fcalderonnearsoft:
Thanks for review @kcooney 

I added commit undoing those.

kcooney:
Thanks. Whether the new code is more clear than the original is a bit subjective. I'm inclined to leave the code as-is, but perhaps another maintenanee will feel differently

fcalderonnearsoft:
Sure. Thanks for you review and feedback :+1:

PR #1265, opened 2016-03-22
Fix typo in assertEmptyThrowsMultipleFailureExceptionForManyThrowables()

ar-vargas:
at MultipleFailureExceptionTest.

I had problems with the other branch and I had to delete it.
Sorry for opening a new pull request.

### DIFF
--- a/src/main/java/org/junit/runner/JUnitCommandLineParseResult.java
+++ b/src/main/java/org/junit/runner/JUnitCommandLineParseResult.java
@@ -15,11 +15,6 @@ class JUnitCommandLineParseResult {
     private final List<Throwable> parserErrors = new ArrayList<Throwable>();
 
     /**
-     * Do not use. Testing purposes only.
-     */
-    JUnitCommandLineParseResult() {}
-
-    /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/junit/runner/JUnitCommandLineParseResult.java
```diff
@@ -15,11 +15,6 @@
     private final List<Throwable> parserErrors = new ArrayList<Throwable>();
 
     /**
-     * Do not use. Testing purposes only.
-     */
-    JUnitCommandLineParseResult() {}
```
kcooney: This is not redundant. It is a package-scope constror, preventing code outside of this package from constructing an instance via the constructor.

stefanbirkner:
I would like to merge ff7311b only. @kcooney  Do you agree?

marcphilipp:
+1

kcooney:
I agree with @stefanbirkner

marcphilipp:
I cherry-picked ff7311b and pushed it to master.

@ar-vargas Thanks!

PR #1272, opened 2016-03-28
Replace ArrayList for primitive array in copyArray method

avargas-nearsoft:
at org.junit.runner.JUnitCommandLineParseResult
According to https://github.com/junit-team/junit4/issues/1271

### DIFF
--- a/src/main/java/org/junit/runner/JUnitCommandLineParseResult.java
+++ b/src/main/java/org/junit/runner/JUnitCommandLineParseResult.java
@@ -85,13 +85,11 @@ class JUnitCommandLineParseResult {
     }
 
     private String[] copyArray(String[] args, int from, int to) {
-        ArrayList<String> result = new ArrayList<String>();
-
+        String[] result = new String[to - from];
         for (int j = from; j != to; ++j) {
-            result.add(args[j]);
+            result[j - from] = args[j];
         }
-
-        return result.toArray(new String[result.size()]);
+        return result;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/junit/runner/JUnitCommandLineParseResult.java
```diff
@@ -85,13 +85,11 @@ private void parseArgs(String[] args) {
     }
 
     private String[] copyArray(String[] args, int from, int to) {
-        ArrayList<String> result = new ArrayList<String>();
-
+        String [] result = new String[to-from];
```
kcooney: Style-nit: we put spaces around operators like -

src/main/java/org/junit/runner/JUnitCommandLineParseResult.java
```diff
@@ -85,13 +85,11 @@ private void parseArgs(String[] args) {
     }
 
     private String[] copyArray(String[] args, int from, int to) {
-        ArrayList<String> result = new ArrayList<String>();
-
+        String [] result = new String[to-from];
         for (int j = from; j != to; ++j) {
-            result.add(args[j]);
+            result [j-from] = args[j];
```
kcooney: We don't put spaces when indexing into an array. Please change to:

    result[j - from] = args[j];

avargas-nearsoft:
Updated.

marcphilipp:
Thanks!

PR #1274, opened 2016-04-05
Test cleanup

paulduffin:
While working on some other enhancements I noted that the test structure (particularly in org.junit tree) is a bit chaotic what with all the tests being in a flat list in org.junit.AllTests and ExperimentalTests with some tests duplicated, some tests missing, some missing tests broken, or using a JUnit3 test instead of the JUnit4 one.. There was also a comment in the AllTests class saying that it needed to be cleaned up which linked to a sourceforge page that no longer existed so cleaning it up seemed long overdue.

I split the work into 3 separate commits:
- The first one removes the duplicates and uses the correct test, this defines the baseline of the tests that must be run after the tests have been reorganized.
- The second one is the bigger one, it simply restructures the tests and doesn't change the tests that are run. It is very easy to see that it runs the same number of tests as after the first commit. The numbers in the commit message are based off IDEA's count not maven's but it has been tested in both.
- The third one adds in the missing tests and fixes a broken test.


marcphilipp:
There was definitely a bit of chaos there, thanks for cleaning that up!

@junit-team/junit-committers I think we should merge this. Any objections?

kcooney:
Nice work! No objections.

It's sad that we need to manually list the test classes. Unfortunately, if we have Maven find the tests and run them, there is a risk that someone could introduce a bug that leads to all tests being reportes as passing tests.

marcphilipp:
@paulduffin Thanks!

paulduffin:
I think that allowing Maven to find the tests and run them would not work because it will find and run tests that are intended to fail.

kcooney:
Good point. Perhaps we could have a custom classpath scanning Suite runner that supports an `@IntendedToFail` annotation. We could have tests for this suite written in JUnit3-style just to be safe.

marcphilipp:
Since this pull request is closed, I'd suggest we should open a new issue to discuss this, shall we?

Issue #1270, opened 2016-03-28
Used only first* from methods marked @Parameters


sirikid:
I think it might be useful to have more than one method marked @ Parameters.
\* lexicographically first

marcphilipp:
Can you please provide a compelling example? ;-)

sirikid:
In the first method I create test data using a PRNG, and in the second I read from a file previously detected edge cases.

kcooney:
@sirikid why do you need two methods annotated with `@Parameters` to do that? You could have one method that does both things (then it's 100% clear what JUnit would do).

PR #1275, opened 2016-04-06
Multiple parameterized methods

sirikid:
#1270


marcphilipp:
I like this idea!

If we decide to do this, the default value for `name` in `@Parameters` does not make sense anymore when there are multiple parameter factory methods. We could change the logic to use `{method}[{index}]` when there are multiple parameter factory methods and use `{index}` when there is only one (to preserve backwards compatibility).

In addition, the Javadoc needs to be updated... but we should first decide whether we want this feature.

@junit-team/junit-committers What do you guys think about this?

kcooney:
@marcphilipp please see my comment on the bug

PR #1279, opened 2016-04-08
In getRunner method of FilterRequest class,when runner is already an ErrorReporingRunner,skip filtering.

### LINKED ISSUE #1277: FilterRequest may hide the real failure cause(exception) of a test
# Description
FilterRequest may hide the real failure cause(exception) of a test

# Steps To Reproduce
## 1. Write the following test code
<pre>
@RunWith(PowerMockRunner.class)
@PrepareForTest(value = { ClassUnderTest.class })
public class BarTest {

	@Test
	public void test() {
		System.out.println("hello world");
	}
}
</pre>

## 3.In eclipse,select the test method,and run as junit

# The expected result
The root cause,namely,javassist failure in stacktrace be shown.
In the case above,Exception javassist.bytecode.InterfaceMethodrefInfo cannot be cast to javassist.bytecode.MethodrefInfo should be seen.

# the current result
<pre>
BarTest.test
initializationError(org.junit.runner.manipulation.Filter)
java.lang.Exception: No tests found matching [{ExactMatcher:fDisplayName=test], {ExactMatcher:fDisplayName=test(BarTest)], {LeadingIdentifierMatcher:fClassName=BarTest,fLeadingIdentifier=test]] from org.junit.internal.requests.ClassRequest@72d1ad2e
	at org.junit.internal.requests.FilterRequest.getRunner(FilterRequest.java:40)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestLoader.createFilteredTest(JUnit4TestLoader.java:77)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestLoader.createTest(JUnit4TestLoader.java:68)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestLoader.loadTests(JUnit4TestLoader.java:43)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:444)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:675)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:382)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:192)

</pre>

# The reason and my suggested solution

Here is the getRunner code of the org.junit.internal.requests.FilterRequest class.

<pre>
    public Runner getRunner() {
        try {
            Runner runner = request .getRunner();
            fFilter.apply( runner);
            return runner ;
        } catch (NoTestsRemainException e ) {
            return new ErrorReportingRunner(Filter.class, new Exception(String
                    . format("No tests found matching %s from %s", fFilter
                            .describe(), request.toString())));
        }
    }
</pre>

It may be better  to be changed  into the following.
<pre>
    public Runner getRunner() {
        try {
            Runner runner = request .getRunner();

            if (runner instanceof ErrorReportingRunner) {
              // runner is already an ErrorReportingRunner
                return runner;
           }
            fFilter.apply( runner);
            return runner ;
        } catch (NoTestsRemainException e ) {
            return new ErrorReportingRunner(Filter.class, new Exception(String
                    . format("No tests found matching %s from %s", fFilter
                            .describe(), request.toString())));
        }
    }
</pre>

chang-chao:
fix #1277

### DIFF
--- a/src/main/java/org/junit/internal/requests/FilterRequest.java
+++ b/src/main/java/org/junit/internal/requests/FilterRequest.java
@@ -34,6 +34,10 @@ public final class FilterRequest extends Request {
     public Runner getRunner() {
         try {
             Runner runner = request.getRunner();
+            if (runner instanceof ErrorReportingRunner) {
+                // runner is already an ErrorReportingRunner
+                return runner;
+            }
             fFilter.apply(runner);
             return runner;
         } catch (NoTestsRemainException e) {


kcooney:
Are you sure this fixes the issue? The default implementation of `Filter.apply()` is a no-op if the runner doesn't implement `Filterable`, and `ErrorReporingRunner` doesn't implement `Filterable`.

Is it possible to write a test that creates a filtered request for a test class or runner that simulates the problem?

chang-chao:
@kcooney Thanks for responding.
In my case,the filter that was applied was not a default filter,but a customized one named [org.eclipse.jdt.internal.junit4.runner.SubForestFilter](https://git.eclipse.org/r/#/c/33189/3/org.eclipse.jdt.junit4.runtime/src/org/eclipse/jdt/internal/junit4/runner/SubForestFilter.java).
We should not assume that all filter implementations have the same behavior as the default one,right?

kcooney:
@chang-chao Could you find the code for `SubForestFilter`?

Custom filters override `Filter.shouldRun()`. I don't think `Filter.apply()` was meant to be overridden; any code overriding it should be sure to not filter runners that are not `Filterable`.

marcphilipp:
https://github.com/eclipse/eclipse.jdt.ui/blob/master/org.eclipse.jdt.junit4.runtime/src/org/eclipse/jdt/internal/junit4/runner/SubForestFilter.java

kcooney:
I think the real problem is with the Eclipse filter. It is explicitly throwing `NoTestsRemainException` in `apply()`. If that `throw` call was removed, In think he problem will be fixed and the filter would still work correctly.

The problem with fixing it in `FilterRequest` is that there are many other places in JUnit (and user's code)/that call `apply()`, and many of those places expect `apply()` on a runner that isn't `Filterable` to be a no-op.

chang-chao:
A bug is filed against eclipse,So this PR is closed.
https://bugs.eclipse.org/bugs/show_bug.cgi?id=491752

PR #1283, opened 2016-04-24
Added a method in ExpectedException to perform other assertions after the expected exception is thrown.

manishgo:
This enables the user to perform assertions for the system under test, at the time when exception was thrown.


kcooney:
Have you seen https://github.com/junit-team/junit4/commit/bdb1799a6b58c4ab40c418111333f4e8e1816e7e ?

manishgo:
Ya I just saw it. Looks like ExpectedException is going to be deprecated. If that is the case, this pull request looks redundant.

kcooney:
Most likely, yes.

Thanks for the pull.request! Let us know if you have other ideas for improvements

PR #1285, opened 2016-04-27
Refactor ErrorReportingRunner

alb-i986:
Commits are separate just for the sake of reviewing; needs to be squashed before merging.

This is just a refactoring, no functional changes involved.

### DIFF
--- a/src/main/java/org/junit/internal/runners/ErrorReportingRunner.java
+++ b/src/main/java/org/junit/internal/runners/ErrorReportingRunner.java
@@ -74,7 +74,7 @@ public class ErrorReportingRunner extends Runner {
 
     private Description createDescription() {
         Description description = Description.createSuiteDescription(classNames);
-        for (int i = 0; i < causes.size(); i++) {
+        for (Throwable each : causes) {
             description.addChild(describeCause());
         }
         return description;
--- a/src/main/java/org/junit/internal/runners/ErrorReportingRunner.java
+++ b/src/main/java/org/junit/internal/runners/ErrorReportingRunner.java
@@ -16,7 +16,7 @@ public class ErrorReportingRunner extends Runner {
     private final Description description;
 
     public ErrorReportingRunner(Class<?> testClass, Throwable cause) {
-        this(cause, new Class<?>[] { testClass });
+        this(cause, testClass);
     }
     
     public ErrorReportingRunner(Throwable cause, Class<?>... testClasses) {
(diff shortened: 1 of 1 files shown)


marcphilipp:
Thanks for the refactoring! However, I would rather revert 06f5a94a540428130d659fbde900d6373750a28d and f7e435aee120e117d73f45352abb4de0bd5cc3c2 because IMHO readability outweighs performance optimizations in this case.

stefanbirkner:
I don't have a strong opinion about f7e435a. There are some reasons for keeping the old version and some for using the new one.

I agree with @marcphilipp on not merging 06f5a94. The other three commits are improvements that should definitely be merged.

kcooney:
Since `Description ` instances are mutable, f7e435a is an externally visible  change. I don't know if it would break anyone,  but since the others are luke-warm about it, we should revert it.

I agree on 06f5a94

kcooney:
Also for 7474220 please don't move `describeCause` because it makes it harder to see the diffs.

marcphilipp:
@alb-i986 Do you want to do the suggested changes or shall i take over?

alb-i986:
@marcphilipp ya sorry, I'll do as soon as I can. Cheers

alb-i986:
About  06f5a94:

Apart from optimization, I think it also helps readability.
As soon as I see a for-each loop, I expect to see the item to be used inside the block. But here it isn't. That's a little bit "unsettling" I think.

About f7e435a:

> Since Description instances are mutable, f7e435a is an externally visible change. I don't know if it would break anyone, but since the others are luke-warm about it, we should revert it.

@kcooney
I see what you mean, good point.

That's unfortunate, IMO. Descriptions should be immutable. There's no point in editing a Description after a Runner builds it and returns it, right?
If you think it's worth it, I'll create an issue for that (maybe in junit5?).

> Also for 7474220 please don't move describeCause because it makes it harder to see the diffs.

If you mean that it's harder to do the code review, ya, I understand. I should have done the move in a different commit, sorry about that.
Anyway, I was following the Clean code principle "Place a (private) method right after the method using it", to improve readability.
Are you happy with the move in a separate commit?

kcooney:
@alb-i986 

> That's unfortunate, IMO. Descriptions should be immutable. 

I agree. We did look into making `Description` immutable (with proper builders) but there was a fair amount of work involved in updating the code to use immutable descriptions (particularly in the filtering code) and the changes might break people, so the effort just lost steam.

JUnit5 doesn't use `Description` internally, so there is no point to creating an issue for this in JUnit5


As for the other changes in this pull, some of them are changing the code from one style to a different style; the benefits of those kinds of changes are somewhat subjective. I can see your point about 06f5a94 from a readability standpoint (though the performance benefits, if any, are small). I don't have a strong feeling about where private methods should be put before or after the method where they are called, but since two people already agreed on the current style (the author and the reviewer) I suggest we should leave `describeChild` where is was.

alb-i986:
Should be good to be squashed now.
Thanks

marcphilipp:
Thanks! 👍

PR #1287, opened 2016-04-29
 Unused code removed

Novanic:
Unused code removed (within package "org.junit.internal" which is probably not intended to be a part of the public api)


kcooney:
Thanks for the contribution

Unfortunately, these classes and methods are public, so we cannot remove them in JUnit 4.13

Novanic:
Hm, I thought at least the "internal" package isn't public.

The changes within class ErrorReportingRunner should be safe, because only private methods are affected. There is also a very strange loop in that class which causes that the same message/description is added multiple times. Could you please take a look at it?

Thanks.

kcooney:
We've broken people in the past by making changes to internal packages, so we've decided to deprecate APIs instead of removing them, and only removing deprecated APIs in major releases.

The `ErrorReporingRunner` code you changed is a bit odd. Perhaps you could evaluate the history of that class; maybe there was a bad merge. We also have two other recent pulls with changes for that class.

PR #1291, opened 2016-05-12
Tests expecting AssumptionViolatedException should be marked as passed, not skipped

### LINKED ISSUE #1290: Tests annotated with `@Test(expected = AssumptionViolatedException.class)` will be marked as skipped instead of passing
When using `@Test(expected = AssumptionViolatedException.class)` the test will actually be marked as skipped instead of passing, as if the AssumptionViolatedException were thrown up in the stack.
In fact that's what seems to happen: https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/internal/runners/statements/ExpectException.java#L22

I've found this bug while doing #1289, while inspecting our skipped unit tests.
See e.g. https://github.com/junit-team/junit4/blob/master/src/test/java/org/junit/tests/experimental/AssumptionTest.java#L80

But maybe this is actually working as expected. I mean, no one should ever expect AssumptionViolatedException. It can happen only to us, in our unit tests. Wherever we need to, we have to use a try/catch block instead.

@stefanbirkner You seem to have worked on assumptions quite a lot, can you please advise?

alb-i986:
Tests annotated with `@Test(expected = AssumptionViolatedException.class)`
which throw AssumptionViolatedException should be marked as passing, not skipped.

Fixes #1290


QinJiangbo:
:sunglasses:

src/test/java/org/junit/internal/runners/statements/ExpectExceptionTest.java
```diff
@@ -0,0 +1,29 @@
...
+import org.junit.internal.AssumptionViolatedException;
+import org.junit.runners.model.Statement;
+
+import static org.junit.Assert.*;
+
+public class ExpectExceptionTest {
+
+    @Test
+    public void givenWeAreExpectingAssumptionViolatedExceptionAndAStatementThrowingAssumptionViolatedException() {
+        ExpectException sut = new ExpectException(new StatementThrowingAssumptionViolatedException(), AssumptionViolatedException.class);
```
kcooney: Can we instead test this in `org.junit.tests.running.methods.ExpectedTest` in the same way we test the expected attributes? We get a bit more confidence if when test features like this by writing tests that run specific tests and verify JUnit gives the proper result for those test runs.

alb-i986: Sure I'll do that. But I also would like to keep this unit test, which is really a unit test, small.

marcphilipp:
This fix looks fine. However, I'm not sure this use case makes sense.

@alb-i986 You had a test in your codebase that used `@Test(expected = AssumptionViolatedException.class)`?

src/test/java/org/junit/internal/runners/statements/ExpectExceptionTest.java
kcooney: Sure, having both kinds of tests is fine. The integration test is necessary to verify the bug is fixed; the unit test verifies the contract of the `ExpectException` class.

If we keep this test class, please have the class Javadoc say where the integration tests are (using`@link`)

A few suggested  changes:

1. Could you update this test to use https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/internal/runners/statements/Fail.java ? That way, you can see what exception the `Statement` is throwing by looking at the test method.
2. Extract a local variable for the expression you pass to the `ExpectException` constructor and call that variable `delegate`. 
3. I don't think we use `sut` as a variable name very often. If we don't, I suggest  naming it `expectException` so future readers wont be confused by the name.

src/test/java/org/junit/internal/runners/statements/ExpectExceptionTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+        try {
+            sut.evaluate();
+            // then no exception should be thrown
+        } catch (Throwable e) {
+            fail("should not throw anything, but was thrown: " + e);
```
kcooney: No need to catch unexpected exceptions; if `evaluate` throws an exception then the test would fail.

src/test/java/org/junit/internal/runners/statements/ExpectExceptionTest.java
```diff
@@ -0,0 +1,29 @@
...
+import org.junit.Test;
+import org.junit.internal.AssumptionViolatedException;
+import org.junit.runners.model.Statement;
+
+import static org.junit.Assert.*;
+
+public class ExpectExceptionTest {
+
+    @Test
+    public void givenWeAreExpectingAssumptionViolatedExceptionAndAStatementThrowingAssumptionViolatedException() {
```
kcooney: I like descriptive method names but this is a bit long. How about:

`whenExpectingAssumptionViolatedExceptionStatementsThrowingItShouldPass`

src/test/java/org/junit/internal/runners/statements/ExpectExceptionTest.java
kcooney: Also, if you are adding unit tests, please add a test that makes sure that tests that expect `AssumptionViolatedException` but do not throw it should fail

kcooney:
@marcphilipp we have tests like this in JUnit. See #1290

alb-i986:
> You had a test in your codebase that used @Test(expected = AssumptionViolatedException.class)?

As Kevin said, no I don't, I've found one in JUnit itself:

https://github.com/junit-team/junit4/blob/master/src/test/java/org/junit/tests/experimental/AssumptionTest.java#L80

I share your concern in #1290, where I was commenting 

> But maybe this is actually working as expected. I mean, no one should ever expect AssumptionViolatedException. It can happen only to us, in our unit tests. Wherever we need to, we have to use a try/catch block instead.

However, I think the scenario may happen outside as well, although rare.
So this should make sense.

src/test/java/org/junit/internal/runners/statements/ExpectExceptionTest.java
alb-i986: Ya right! Tricky uh? :) Testing JUnit with JUnit is tricky itself

kcooney:
I agree that no one *should* ever expect `AssumptionViolatedException` but the behavior they get if they do may be unexpected.

This change seems safe. The tests were previously marked as skipped, and now they will be marked as passing. Then again, some users might expect that *any* test method with a violated assumption should be marked as skipped.

Another option is to have tests annotated with `Test(expected = AssumptionViolatedException.class)`  fail with a message indicating that this construct is ambiguous, and directing them to use `assertThrows` or `expectThrows`.

src/test/java/org/junit/internal/runners/statements/ExpectExceptionTest.java
alb-i986: agreed

src/test/java/org/junit/internal/runners/statements/ExpectExceptionTest.java
alb-i986: > Could you update this test to use https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/internal/runners/statements/Fail.java ?

done

> Extract a local variable for the expression you pass to the ExpectException constructor and call that variable delegate

Done, but I gave it a more meaningful name, although a bit longer: `statementThrowingAssumptionViolatedException`

> I don't think we use sut as a variable name very often. If we don't, I suggest naming it expectException so future readers wont be confused by the name

Done but I would suggest to consider adopting it.
I think I took it from xunitpatterns, see e.g. http://xunitpatterns.com/Test%20Stub.html (scroll to "Motivating Example"):

```
   public void testDisplayCurrentTime_AtMidnight() {
      // fixture setup
      TimeDisplay sut = new TimeDisplay();
      // exercise sut
      String result = sut.getCurrentTimeAsHtmlFragment();
      // verify direct output
      String expectedTimeString = "<span class=\"tinyBoldText\">Midnight</span>";
      assertEquals( expectedTimeString, result);
   }
```

I find it a great name as it clearly stands out among the many variables (the DOCs, following Meszaros's terminology) which may be defined in a unit test.

It also makes it extremely easy to distinguish the "when" step: it's the one (and only, in theory) of the form `sut.method(inputs)`. In the example above, `sut.getCurrentTimeAsHtmlFragment();`

> Also, if you are adding unit tests, please add a test that makes sure that tests that expect AssumptionViolatedException but do not throw it should fail

Done. I've also added a couple more scenarios:
- scenario where the given statement passes
- scenario where the statement throws subclass of AssumptionViolatedEx

marcphilipp:
I tend to think that we should mark them as passing.

@junit-team/junit-committers Opinions?

src/test/java/org/junit/internal/runners/statements/ExpectExceptionTest.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Integration tests can be found in {@link org.junit.tests.running.methods.ExpectedTest}.
+ * See e.g. {@link org.junit.tests.running.methods.ExpectedTest#expectsAssumptionViolatedException()}
+ */
+public class ExpectExceptionTest {
+
+    @Test
+    public void whenExpectingAssumptionViolatedExceptionStatementsThrowingItShouldPass() {
+        Statement statementThrowingAssumptionViolatedException = new Fail(new AssumptionViolatedException("expected"));
+        ExpectException expectException = new ExpectException(statementThrowingAssumptionViolatedException, AssumptionViolatedException.class);
```
kcooney: If you want to keep the long variable name for the Statement arg then please wrap this line to make it readable:

    ExpectException expectException = new ExpectException(
            statementThrowingAssumptionViolatedException,
            AssumptionViolatedException.class);
 
But I really prefer `delegate` or  `delegateStatement` because that is what the role of the Statement is for the sut. The variable is declared one line up, so the long name isn't needed and makes the code harder to scan IMHO.

src/test/java/org/junit/internal/runners/statements/ExpectExceptionTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Test
+    public void whenExpectingAssumptionViolatedExceptionStatementsThrowingDifferentExceptionShouldFail() {
+        Statement statementThrowingSomeException = new Fail(new SomeException("not expected"));
+        ExpectException expectException = new ExpectException(statementThrowingSomeException, AssumptionViolatedException.class);
+
+        try {
+            expectException.evaluate();
+            fail("should throw 'Unexpected exception' when statement throws an exception which is not the one expected");
+        } catch (Exception e) {
```
kcooney: Is there a more specific exception that can be caught?

src/test/java/org/junit/internal/runners/statements/ExpectExceptionTest.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    @Test
+    public void whenExpectingAssumptionViolatedExceptionStatementsPassingShouldFail() throws Exception {
+        ExpectException expectException = new ExpectException(new PassingStatement(), AssumptionViolatedException.class);
+
+        try {
+            expectException.evaluate();
+            fail("ExpectException should throw when the given statement passes");
```
kcooney: This call throws `AssertionError` so this won't work. You want:

    try {
        expectException.evaluate();
    }  catch (AssertionError e) {
        assertThat(e.getMessage(), containsString("Expected exception: " + AssumptionViolatedException.class.getName()));
        return;
    }
    fail("ExpectException should throw when the given statement passes");

kcooney:
Sorry for nit-picking about the tests, but I found the original unit tests to be hard to understand

kcooney:
Marc, I agree that this is a good fix.

src/test/java/org/junit/internal/runners/statements/ExpectExceptionTest.java
alb-i986: Unfortunately not, at the minute. I was going to open a new PR to change the exception thrown in case of unexpected exception to be of a new type `UnexpectedException extends AssertionError`.

src/test/java/org/junit/internal/runners/statements/ExpectExceptionTest.java
alb-i986: In the catch block I verify the exception message, so it works, but I agree it's not nice.
Your solution is correct but a little hard to read.

As in the other comment, the best would be to have the SUT throw a more specific exception. I'm not sure about the name though..`ExpectedException`?

kcooney: The idiom.I wrote above is quite common in JUnit's tests, so let's go with that. We can't simply rely on checking the exception  message because it also can result in tests passing that shouldn't pass, plus the failure messages you get if the test fails are confusing.

I don't think a new exception  type is worth it, bu the other maintainers might disagree with me and agree with you :-(

Verifying that a method or constructor call throws a specific exception would be easier if we could use lambdas and `assertThrows`

marcphilipp: I agree with Kevin. Let's not introduce a new exception class.

@alb-i986 Please follow @kcooney's suggestion and rewrite the try-catch-block accordingly.

kcooney:
Merged

alb-i986:
@kcooney 
Sure, done.
Cheers

PR #1292, opened 2016-05-13
Fix ResultMatchers#hasFailureContaining

alb-i986:
It should *not* match when the given PrintableResult has no failures.

Please note that this change replaces the dependency hamcrest-core with hamcrest-library. I'm not sure if you are happy with that. If you are, we probably also need to update another couple of lines in the pom, see e.g. https://github.com/junit-team/junit4/blob/master/pom.xml#L308

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,7 +110,7 @@
     <dependencies>
         <dependency>
             <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest-library</artifactId>
+            <artifactId>hamcrest-core</artifactId>
             <version>1.3</version>
         </dependency>
     </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -110,7 +110,7 @@
     <dependencies>
         <dependency>
             <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest-core</artifactId>
+            <artifactId>hamcrest-library</artifactId>
             <version>1.3</version>
         </dependency>
     </dependencies>
--- a/src/main/java/org/junit/experimental/results/ResultMatchers.java
+++ b/src/main/java/org/junit/experimental/results/ResultMatchers.java
@@ -5,8 +5,6 @@ import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 import org.hamcrest.TypeSafeMatcher;
 
-import static org.hamcrest.CoreMatchers.equalTo;
-
 /**
  * Matchers on a PrintableResult, to enable JUnit self-tests.
  * For example:
@@ -36,21 +34,14 @@ public class ResultMatchers {
      * Matches if there are {@code count} failures
      */
     public static Matcher<PrintableResult> failureCountIs(final int count) {
-        return failureCount(equalTo(count));
-    }
-
-    /**
-     * Matches if the number of failures matches {@code countMatcher}
-     */
-    public static Matcher<PrintableResult> failureCount(final Matcher<Integer> countMatcher) {
         return new TypeSafeMatcher<PrintableResult>() {
             public void describeTo(Description description) {
-                description.appendText("has a number of failures matching " + countMatcher);
+                description.appendText("has " + count + " failures");
             }
 
             @Override
             public boolean matchesSafely(PrintableResult item) {
-                return countMatcher.matches(item.failureCount());
+                return item.failureCount() == count;
             }
         };
     }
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -110,7 +110,7 @@
     <dependencies>
         <dependency>
             <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest-core</artifactId>
+            <artifactId>hamcrest-library</artifactId>
```
marcphilipp: I don't think we should change this dependency.

marcphilipp:
@junit-team/junit-committers Does anyone know why `ResultMatchers` is in `src/main/java` instead of `src/test/java`?

kcooney:
My guess is that `ResultMatchers` is there so external users writing Rules and other test extensions could use it to verify test their code works in actual tests.

In any case, it's too late to move it now

marcphilipp:
That's true, unfortunately.

@alb-i986 Can you try to solve this without using `hamcrest-library`?

alb-i986:
> Can you try to solve this without using hamcrest-library?

Done, but why's that?

marcphilipp:
I don't want to force regular users and IDEs like Eclipse (who bundle JUnit) to jump through hoops just so we can use `greaterThan`. Does that make sense?

marcphilipp:
This looks good to me now. Thanks, @alb-i986! 👍 

@kcooney What do you think?

alb-i986:
I'm not sure what you mean by "jump through hoops", but ok, as it turned out I could do without it.

But like at least we could add hamcrest-library as a test dependency for ourselves. But I'm not sure if that's easy to configure.

marcphilipp:
Yeah, that would make sense and should be easy to configure.

src/main/java/org/junit/experimental/results/ResultMatchers.java
```diff
@@ -34,14 +36,21 @@ public ResultMatchers() {
      * Matches if there are {@code count} failures
      */
     public static Matcher<PrintableResult> failureCountIs(final int count) {
+        return failureCount(equalTo(count));
+    }
+
+    /**
+     * Matches if the number of failures matches {@code countMatcher}
+     */
+    public static Matcher<PrintableResult> failureCount(final Matcher<Integer> countMatcher) {
```
kcooney: I'm curious why we need an overload that takes a `Matcher<Integer>`. Can you give an example of where this would be more useful than the version that takes an `int`?

Also, shouldn't the parameter have a type of `Matcher<? super Integer>` (see http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#arrayWithSize(org.hamcrest.Matcher) for one example in Hamcrest)

src/main/java/org/junit/experimental/results/ResultMatchers.java
```diff
@@ -34,14 +36,21 @@ public ResultMatchers() {
...
         return new TypeSafeMatcher<PrintableResult>() {
             public void describeTo(Description description) {
-                description.appendText("has " + count + " failures");
+                description.appendText("has a number of failures matching " + countMatcher);
```
kcooney: I personally don't see anything wrong with the existing description. In theory, changing the description could break some users of `ResultMatchers.failureCountIs()`. Is there some motivation for the change?

src/main/java/org/junit/experimental/results/ResultMatchers.java
alb-i986: The existing `failureCountIs(final int count)` only allows to do an equals match, whereas taking a Matcher of Integer allows clients to do any kind of integer matching, eg. <, >, >=, <=.
It came from the original implementation of this PR 9f5dfd8c1a9a006bc2e5d93a9bc770f2449e7afa

But since I later changed implementation (69e8a8a6b3d3d1074476463e85dcfef3b9feb0c7), it's not needed for this PR. I've removed it and I'll open a new PR for that, as I think it still makes sense on its own.

Please let me know your thoughts

src/main/java/org/junit/experimental/results/ResultMatchers.java
alb-i986: I understand your concern but the original string doesn't work well on failure with greaterThan and similar:

- Expected: has a value equal to or greater than <2> failures
- Expected: has a value greater than <2> failures
- Expected: has a value less than <1> failures

Sounds better:

    Expected: has a number of failures matching a value equal to or greater than <2>

But we could have the original failureCountIs override describeTo of the new overloaded version..

kcooney:

alb-i986:
Done:

https://github.com/junit-team/junit4/wiki/4.13-Release-Notes#pull-request-1292-fix-resultmatchershasfailurecontaining

PR #1293, opened 2016-05-13
MultipleFailureException: expose the stacktrace of the nested exceptions in getMessage

alb-i986:
This makes it possible to debug tests failing with a MultipleFailureException made up of many exceptions.

### DIFF
--- a/src/test/java/org/junit/tests/assertion/MultipleFailureExceptionTest.java
+++ b/src/test/java/org/junit/tests/assertion/MultipleFailureExceptionTest.java
@@ -7,12 +7,10 @@ import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.startsWith;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.lang.annotation.AnnotationFormatError;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


src/test/java/org/junit/tests/assertion/MultipleFailureExceptionTest.java
```diff
@@ -62,12 +65,28 @@ public void assertEmptyThrowsMultipleFailureExceptionForManyThrowables() throws
...
         } catch (MultipleFailureException expected) {
             assertThat(expected.getFailures(), equalTo(errors));
-            assertTrue(expected.getMessage().startsWith("There were 2 errors:\n"));
-            assertTrue(expected.getMessage().contains("ExpectedException(basil)\n"));
-            assertTrue(expected.getMessage().contains("RuntimeException(garlic)"));
+            assertThat(expected.getMessage(), allOf(
+                    startsWith("There were 2 errors:\n"),
+                    containsString("ExpectedException: basil\n"),
+                    containsString("RuntimeException: garlic\n")
+            ));
```
marcphilipp: Nice! 👍

marcphilipp:
I'm not sure about this change. I see your point that the stacktraces are quite hidden at the moment when debugging -- but they are accessible using `getFailures()`, right?

kcooney:
Can you provide before and after examples of failing tests run with `JUnitCore` (and possibly maven or gradle)?

alb-i986:
Sorry, my bad, you're right: the stacktraces of the encapsulated exceptions are actually displayed by both maven and gradle.
In fact I see that [EachTestNotifier is unpacking the MultipleFailureException and firing multiple Failure's](https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/internal/runners/model/EachTestNotifier.java#L21).

This is a branch I made a while ago, so I'm not sure if at that time things were different, or if it was just an oversight.
Either way, I'll just close this.
Cheers

Issue #98, opened 2010-04-27
assumeTrue() does not work with expected exceptions


jonas22:
assumeTrue() does not work if a test has an expected exception.

Example code:

```
import static org.junit.Assume.*;
import org.junit.Test;

public class TestAssume {
    @Test(expected = IllegalArgumentException.class)
    public void test() {
        assumeTrue(false);
    }
}
```

This exception is thrown by JUnit 4.8.1:

```
java.lang.Exception: Unexpected exception, expected<java.lang.IllegalArgumentException> but was<org.junit.internal.AssumptionViolatedException>
    at org.junit.internal.runners.statements.ExpectException.evaluate(ExpectException.java:28)
    at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:76)
    at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:50)
    at org.junit.runners.ParentRunner$3.run(ParentRunner.java:193)
    at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:52)
    at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:191)
    [...]
```

PR #1294, opened 2016-05-13
Add Result#getAssumptionFailureCount

alb-i986:
My original intent was to fix unit test assumeWithExpectedException: it was throwing AssumptionViolatedException thus being skipped (see #98).

It led me to adding Result#getAssumptionFailureCount.

### DIFF
--- a/src/main/java/org/junit/runner/Result.java
+++ b/src/main/java/org/junit/runner/Result.java
@@ -28,6 +28,7 @@ public class Result implements Serializable {
             ObjectStreamClass.lookup(SerializedForm.class).getFields();
     private final AtomicInteger count;
     private final AtomicInteger ignoreCount;
+    private final AtomicInteger assumptionFailureCount;
     private final CopyOnWriteArrayList<Failure> failures;
     private final AtomicLong runTime;
     private final AtomicLong startTime;
@@ -38,6 +39,7 @@ public class Result implements Serializable {
     public Result() {
         count = new AtomicInteger();
         ignoreCount = new AtomicInteger();
+        assumptionFailureCount = new AtomicInteger();
         failures = new CopyOnWriteArrayList<Failure>();
         runTime = new AtomicLong();
         startTime = new AtomicLong();
@@ -46,6 +48,7 @@ public class Result implements Serializable {
     private Result(SerializedForm serializedForm) {
         count = serializedForm.fCount;
         ignoreCount = serializedForm.fIgnoreCount;
+        assumptionFailureCount = serializedForm.assumptionFailureCount;
         failures = new CopyOnWriteArrayList<Failure>(serializedForm.fFailures);
         runTime = new AtomicLong(serializedForm.fRunTime);
         startTime = new AtomicLong(serializedForm.fStartTime);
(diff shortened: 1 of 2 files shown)


src/test/java/org/junit/tests/experimental/AssumptionTest.java
```diff
@@ -188,9 +188,18 @@ public void failingAssumptionInConstructorIgnoresClass() {
         assertThat(testResult(AssumptionFailureInConstructor.class), isSuccessful());
     }
 
-    @Test(expected = IllegalArgumentException.class)
-    public void assumeWithExpectedException() {
-        assumeTrue(false);
+    public static class TestClassWithAssumptionFailure {
+
+        @Test(expected = IllegalArgumentException.class)
```
marcphilipp: I know it was there before, but the `expected` parameter is confusing and should be removed.

alb-i986: It was part of the scenario. It comes the example given in the bug report: #98

kcooney:
This changes the serialized form of this class. Do we know if IDEs or build tools serialize it?

marcphilipp:
Good catch, @kcooney! I guess it's okay to *add* an instance variable, though. On the JVM with the old version of the class this field should just get ignored. But we should write a test for that scenario.

marcphilipp:
@alb-i986 Can you add such a test? Please let me know if you need help.

kcooney:
Here is how we worked around this problem last time: f6149a9236d

Due to that fix we might be safe.

src/main/java/org/junit/runner/Result.java
```diff
@@ -156,13 +166,15 @@ public RunListener createListener() {
         private static final long serialVersionUID = 1L;
         private final AtomicInteger fCount;
         private final AtomicInteger fIgnoreCount;
+        private final AtomicInteger assumptionFailureCount;
```
kcooney: I'm reasonably sure that if you add this, you'll be breaking serialization. It's possible that new code will be able to read objects serialized with old code (the tests in `ResultTests.java` check for that) but code linked against 4.12 won't be able to read `Result` objects serialized by versions after 4.12. That _may_ break some tests or IDEs.

Could we instead not add this field, and have the `Result(SerializedForm)` constructor set a field to indicate that the assumption failure count cannot be determined?

kcooney:
According to http://www.jguru.com/faq/view.jsp?EID=5064 we can safely add new fields while still allowing code running with older versions of the class to read serialized data written by the newer version. So we just need to make sure that the class behaves well when it is deserialized from a binary form that was written by an older version (which didn't know about this new field). I suggest when this happens, calling `getAssumptionFailureCount()` should throw an exception.

src/main/java/org/junit/runner/Result.java
```diff
@@ -87,6 +90,13 @@ public int getIgnoreCount() {
     }
 
     /**
+     * @return the number of tests skipped because of an assumption failure
```
kcooney: Could you please add:

    @since 4.13

kcooney:
@alb-i986 did you have any time to fix the issues no this pull soon? If not, it won't make 4.13

alb-i986:
@kcooney when is 4.13 due?

kcooney:
@alb-i986 we don't have a due date, but it has been over two years since 4.12 was released, and there are few remaining open issues targeted to 4.13

marcphilipp:
@alb-i986 Any news? Please let us know if you can do the requested changes.

kcooney:
Replaces by #1530

PR #1289, opened 2016-05-12
Fix mvn surefire config: better test output

alb-i986:
Mvn test output is now more detailed and helpful.

> Fix mvn surefire config: force surefire-junit47 as the surefire provider
> Previously surefire-junit3 was used.
> See also https://maven.apache.org/surefire/maven-surefire-plugin/examples/providers.html

Before:

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.junit.tests.AllTests
Tests run: 940, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 8.58 sec - in org.junit.tests.AllTests

Results :

Tests run: 940, Failures: 0, Errors: 0, Skipped: 0
```

After:

```
[INFO] --- maven-surefire-plugin:2.19.1:test (default-test) @ junit ---

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.junit.tests.assertion.AssertionTest
Tests run: 86, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.028 sec - in org.junit.tests.assertion.AssertionTest
Running org.junit.tests.assertion.ComparisonFailureTest
Tests run: 12, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.001 sec - in org.junit.tests.assertion.ComparisonFailureTest
Running org.junit.tests.assertion.MultipleFailureExceptionTest
Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0 sec - in org.junit.tests.assertion.MultipleFailureExceptionTest
Running org.junit.tests.deprecated.JUnit4ClassRunnerTest
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0 sec - in org.junit.tests.deprecated.JUnit4ClassRunnerTest
Running org.junit.tests.description.AnnotatedDescriptionTest
Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.01 sec - in org.junit.tests.description.AnnotatedDescriptionTest
Running org.junit.tests.description.SuiteDescriptionTest
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0 sec - in org.junit.tests.description.SuiteDescriptionTest
Running org.junit.tests.description.TestDescriptionMethodNameTest
[...]

PR #1299, opened 2016-05-14
refactor  AssumptionTests

alb-i986:
Use a try/catch block instead of the attribute 'expected' of the @Test annotation.
See bug #1290.
See discussion on #1289.


marcphilipp:
Thanks, @alb-i986! 👍

PR #1300, opened 2016-05-15
 A sort of fix for issue #1127 (take 2)

tomwhoiscontrary:
*This is substantially the same change as in #1216, ported onto master [as discussed](https://github.com/junit-team/junit4/pull/1216#issuecomment-219213799). I will repeat the PR blurb for ease of reading ...*

In issue #1127, @sf105 noted that when comparing two arrays which differ in length, assertArrayEquals() will only report that they differ in length, which he found insufficient for easy diagnosis. As a fix, he suggested printing the complete actual array contents.

Rather than doing that, this PR changes assertArrayEquals() to do the usual array comparison even when arrays differ in length, producing a failure message which combines the difference in length and the first difference in content. I think this should ease diagnosis, as @sf105 wanted, and it doesn't require making big changes to the code, adding Hamcrest, introducing a new behaviour (JUnit doesn't print complete array contents anywhere else), or blowing people's IDEs up when they compare multi-megabyte byte arrays.

### DIFF
--- a/src/main/java/org/junit/internal/ComparisonCriteria.java
+++ b/src/main/java/org/junit/internal/ComparisonCriteria.java
@@ -41,9 +41,23 @@ public abstract class ComparisonCriteria {
 
         // Only include the user-provided message in the outer exception.
         String exceptionMessage = outer ? header : "";
-        int expectedsLength = assertArraysAreSameLength(expecteds, actuals, exceptionMessage);
 
-        for (int i = 0; i < expectedsLength; i++) {
+        if (expecteds == null) {
+            Assert.fail(exceptionMessage + "expected array was null");
+        }
+        if (actuals == null) {
+            Assert.fail(exceptionMessage + "actual array was null");
+        }
+
+        int actualsLength = Array.getLength(actuals);
+        int expectedsLength = Array.getLength(expecteds);
+        if (actualsLength != expectedsLength) {
+            header = header + "array lengths differed, expected.length="
+                    + expectedsLength + " actual.length=" + actualsLength + "; ";
+        }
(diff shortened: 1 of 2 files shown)


src/main/java/org/junit/internal/ComparisonCriteria.java
```diff
@@ -41,9 +41,23 @@ private void arrayEquals(String message, Object expecteds, Object actuals, boole
...
+            Assert.fail(exceptionMessage + "expected array was null");
+        }
+        if (actuals == null) {
+            Assert.fail(exceptionMessage + "actual array was null");
+        }
+
+        int actualsLength = Array.getLength(actuals);
+        int expectedsLength = Array.getLength(expecteds);
+        if (actualsLength != expectedsLength) {
+            header = header + "array lengths differed, expected.length="
```
marcphilipp: `+=`?

marcphilipp:
This change looks really useful.

@junit-team/junit-committers Please take a look! :-)

kcooney:
I agree with Marc

src/main/java/org/junit/internal/ComparisonCriteria.java
tomwhoiscontrary: Can do. I always get a bit creeped out by doing += on strings, for some reason!

kcooney:
Merged. Thanks!

@tomwhoiscontrary could you please update the release notes at https://github.com/junit-team/junit4/wiki/4.13-Release-Notes ?

tomwhoiscontrary:
@kcooney Thanks. Release note added - let me know if it needs improving.

kcooney:
@tomwhoiscontrary looks great. Thanks!

PR #1302, opened 2016-05-18
Overload ResultMatchers.failureCountIs: take a Matcher of integer

alb-i986:
This allows clients to assert that the failure count is greaterThan, lessThan, etc.

### DIFF
--- a/src/main/java/org/junit/experimental/results/ResultMatchers.java
+++ b/src/main/java/org/junit/experimental/results/ResultMatchers.java
@@ -46,6 +46,8 @@ public class ResultMatchers {
 
     /**
      * Matches if the number of failures matches {@code countMatcher}
+     *
+     * @since 4.13
      */
     public static Matcher<PrintableResult> failureCountIs(Matcher<? super Integer> countMatcher) {
         return new FailureCountMatcher(countMatcher);
--- a/src/main/java/org/junit/experimental/results/ResultMatchers.java
+++ b/src/main/java/org/junit/experimental/results/ResultMatchers.java
@@ -5,6 +5,8 @@ import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 import org.hamcrest.TypeSafeMatcher;
 
+import static org.hamcrest.CoreMatchers.equalTo;
+
 /**
  * Matchers on a PrintableResult, to enable JUnit self-tests.
  * For example:
@@ -34,16 +36,19 @@ public class ResultMatchers {
      * Matches if there are {@code count} failures
      */
     public static Matcher<PrintableResult> failureCountIs(final int count) {
-        return new TypeSafeMatcher<PrintableResult>() {
+        return new FailureCountMatcher(equalTo(count)) {
+            @Override
             public void describeTo(Description description) {
                 description.appendText("has " + count + " failures");
             }
-
-            @Override
-            public boolean matchesSafely(PrintableResult item) {
-                return item.failureCount() == count;
-            }
         };
+    }
+
+    /**
+     * Matches if the number of failures matches {@code countMatcher}
+     */
+    public static Matcher<PrintableResult> failureCountIs(Matcher<? super Integer> countMatcher) {
+        return new FailureCountMatcher(countMatcher);
     }
 
     /**
@@ -75,5 +80,22 @@ public class ResultMatchers {
                 description.appendText("has failure containing " + string);
             }
         };
+    }
+
+    private static class FailureCountMatcher extends TypeSafeMatcher<PrintableResult> {
+        private final Matcher<? super Integer> countMatcher;
+
+        public FailureCountMatcher(Matcher<? super Integer> countMatcher) {
+            this.countMatcher = countMatcher;
+        }
+
+        public void describeTo(Description description) {
+            description.appendText("has a number of failures matching " + countMatcher);
+        }
+
+        @Override
+        public boolean matchesSafely(PrintableResult item) {
+            return countMatcher.matches(item.failureCount());
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/test/java/org/junit/tests/experimental/results/ResultMatchersTest.java
```diff
@@ -20,4 +29,29 @@ public void hasFailuresDescriptionReflectsInput(int i) {
...
+        PrintableResult resultWithOneFailure = new PrintableResult(Collections.singletonList(
+                new Failure(Description.createTestDescription("class", "name"), new RuntimeException("failure 1"))));
+
+        assertThat(ResultMatchers.failureCountIs(equalTo(3)).matches(resultWithOneFailure), is(false));
+        assertThat(ResultMatchers.failureCountIs(equalTo(1)).matches(resultWithOneFailure), is(true));
+    }
+
+    @Test
+    public void failureCountIsTakingMatcherShouldHaveGoodDescription() {
+        Matcher<Integer> matcherTestStub = new BaseMatcher<Integer>() {
```
kcooney: Can we make this a `Matcher<Number>` to ensure that the code compiles?

src/main/java/org/junit/experimental/results/ResultMatchers.java
```diff
@@ -34,19 +36,22 @@ public ResultMatchers() {
...
-            public boolean matchesSafely(PrintableResult item) {
-                return item.failureCount() == count;
-            }
         };
     }
 
     /**
+     * Matches if the number of failures matches {@code countMatcher}
+     */
+    public static Matcher<PrintableResult> failureCountIs(Matcher<? super Integer> countMatcher) {
```
kcooney: Should this be named `failureCountMatches()`? 

Also, please add `@since 4.13` to the Javadoc

src/test/java/org/junit/tests/experimental/results/ResultMatchersTest.java
alb-i986: What's wrong?
Compilation looks fine to me:
https://travis-ci.org/junit-team/junit4/builds/131156153

src/main/java/org/junit/experimental/results/ResultMatchers.java
alb-i986: failureCountIs reads more natural

    assertThat(result, failureCountIs(greaterThan(2))

    assertThat(result, failureCountMatches(greaterThan(2))

About the since annotation, willl do, thanks.

src/test/java/org/junit/tests/experimental/results/ResultMatchersTest.java
kcooney: Nothing is wrong but nothing verifies that you can use this new API with `Matcher<Number>` or  `Matcher<Object>`

alb-i986: I see what you mean.

I understand this may be a bit OT  but if we take one step back, are we sure we need a lower bound?

    failureCountIs(Matcher<? super Integer>)

I think we need quite the opposite: an upper bound, so as to allow any Matcher<Integer> and _below_, more specific; we do not want to allow something generic like Matcher<Object>, do we?

See also https://docs.oracle.com/javase/tutorial/java/generics/wildcardGuidelines.html:

> An "in" variable is defined with an upper bounded wildcard, using the extends keyword.

But since Integer is final, we may just go for no bound at all: `Matcher<Integer>` or at most `Matcher<? extends Comparable<Integer>>`, which happens to be the [return type of greaterThan](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#greaterThan%28T%29).

        ResultMatchers.failureCountIs(allOf(equalTo(2), greaterThan(2)));

The above line compiles with both `Matcher<Integer>` and `Matcher<? extends Comparable<Integer>>`.

Then, again, I understand that `Matcher<? super Integer>` is what hamcrest uses in cases like this (e.g. in [arrayWithSize](http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#arrayWithSize%28org.hamcrest.Matcher%29)) , so I'll understand if you prefer to comply.

kcooney: I am sure you want an upper bound. That is what `hasSize()` uses.

src/main/java/org/junit/experimental/results/ResultMatchers.java
marcphilipp: I think it should just be called `failureCount`. That's how Hamcrest usually does it. If you want a more readable version you can use `is` as syntactic sugar, i.e.

```java
assertThat(result, failureCount(is(greaterThan(2))));
```

src/test/java/org/junit/tests/experimental/results/ResultMatchersTest.java
marcphilipp: I absolutely agree with @kcooney.

kcooney: @alb-i986 Can the type of `matcherTestStub` be changed to `Matcher<Number>`?

Otherwise this LGTM

kcooney:
@marcphilipp at one point we talked about reducing dependencies from JUnit to Hamcrest. Is that still the plan, or did plans change with JUnit Lambda?

marcphilipp:
As far as I remember that was at a time when we thought we would evolve the JUnit 4 codebase into JUnit 5. Since that plan has become obsolete I don't see a reason why we shouldn't improve existing Hamcrest-based APIs.

kcooney:
@marcphilipp Does that mean we should remove the deprecations we did in #1150 ?

marcphilipp:
On second thought, I don't think we should add additional APIs that rely on Hamcrest. I'm not sure about Hamcrest's progress towards a new version, but IMHO since `ResultMatchers´ should never have been a public API in the first place, I think we should leave it as is.

kcooney:
Closing 

Thanks for the contribution, @alb-i986; sorry we couldn't merge it

PR #1303, opened 2016-05-21
Prevent following symbolic links when deleting temporary directories.

varming:
Following symbolic links when removing files can lead to the removal of files outside the directory structure rooted in the temporary folder, and it can lead to unbounded recursion if a symbolic link points to a directory where the link is directly reachable from.

Preventing following symbolic links is pretty easy with JDK5 and later, but writing a test is ugly because there is no java interface to create symbolic links until NIO2 in JDK7. So I'll hold off with a test unless a reviewer insists.


kcooney:
LGTM

@junit-team/junit-committers any objections to merging this?

marcphilipp:
LGTM, too!

kcooney:
Merged. Thanks!

Issue #1223, opened 2015-11-24
TemporaryFolder doesn't work for parallel test execution in several JVMs


tryge:
Our observation is that the TemporaryFolder is deleted during tests when we run our gradle build with the `--parallel` flag.

We have several thousand unit tests that in part run in parallel in several JVMs (spawned by gradle). What I think happens is that when `junit` retrieves a temporary file and deletes it in order to create a directory instead using the same name, between the point where the file is deleted and the directory is created another test in another JVM chooses the same name for the temporary folder. The first test to finish execution deletes the whole directory and the second throws an exception because the directory (or a file in it) doesn't exist anymore.

One option for us is to switch to `Files.createTempDirectory`, but I like how `TemporaryFolder` works so I'm in favour of keeping it (and not changing all of our tests...)

marcphilipp:
This issue has been addressed in #1304.

jzgithub1:
Was this issue every fixed for certain?

PR #1304, opened 2016-05-22
Retry TemporaryFolder.newFolder's call to mkdir if unsuccessful

varming:
If java.io.File.mkdir returns false, then a new directory has not been created, causing problems when running tests in parallel (see issue #1223).

### DIFF
--- a/src/main/java/org/junit/rules/TemporaryFolder.java
+++ b/src/main/java/org/junit/rules/TemporaryFolder.java
@@ -44,6 +44,9 @@ public class TemporaryFolder extends ExternalResource {
     private final boolean assureDeletion;
     private File folder;
 
+    private static final int TEMP_DIR_ATTEMPTS = 10000;
+    private static final String TMP_PREFIX = "junit";
+
     /**
      * Create a temporary folder which uses system default temporary-file 
      * directory to create temporary resources.
@@ -160,7 +163,7 @@ public class TemporaryFolder extends ExternalResource {
      * Returns a new fresh file with a random name under the temporary folder.
      */
     public File newFile() throws IOException {
-        return File.createTempFile("junit", null, getRoot());
+        return File.createTempFile(TMP_PREFIX, null, getRoot());
     }
 
     /**
@@ -216,10 +219,24 @@ public class TemporaryFolder extends ExternalResource {
     }
 
     private File createTemporaryFolderIn(File parentFolder) throws IOException {
-        File createdFolder = File.createTempFile("junit", "", parentFolder);
-        createdFolder.delete();
-        createdFolder.mkdir();
-        return createdFolder;
+        File createdFolder = null;
+        for (int i = 0; i < TEMP_DIR_ATTEMPTS; ++i) {
+            // Use createTempFile to get a suitable folder name.
+            String suffix = ".tmp";
+            File tmpFile = File.createTempFile(TMP_PREFIX, suffix, parentFolder);
+            String tmpName = tmpFile.getName();
+            // Discard suffix of tmpName.
+            String folderName = tmpName.substring(0, tmpName.length() - suffix.length());
+            createdFolder = new File(parentFolder, folderName);
+            if (createdFolder.mkdir()) {
+                tmpFile.delete();
+                return createdFolder;
+            }
+            tmpFile.delete();
+        }
+        throw new IOException("Unable to create temporary directory in: "
+            + parentFolder.toString() + ". Tried " + TEMP_DIR_ATTEMPTS + " times. "
+            + "Last attempted to create: " + createdFolder.toString());
     }
 
     /**


kcooney:
I don't feel comfortable having an infinite loop. We should limit the attempts.

According to some of the comments at http://stackoverflow.com/a/617438 Java might not delete files right away. One of the commenters there suggested a fix and that fix should limit collisions as well.

varming:
Guava does 10,000 attempts and use a new timestamp on every attempt, before throwing an exception. We can do something similar and throw an IOException if all attempts fail.

kcooney:
I was thinking of something like this:

    for  (int i = 0; i < TEMP_DIR_ATTEMPTS; ++i) {
        File tmp = File.createTempFile("junit", "", parentFolder);
        File createdFolder = new File(temp.getPath + ".d");
        if (createdFolder.mkdir()) {
            tmp.delete()
            return createdFolder;
         }
         tmp.delete();
    }
    ...

varming:
What is benefit of calling createTempFile when we cannot use the resulting file name directly?

varming:
The is one thing I really do not like about repeated executions of:
```
    File tmp = File.createTempFile("junit", ".tmp", parentFolder);
    String path = temp.getPath();
    File createdFolder = new File(path().substring(0, path.length() - 4));
    if (createdFolder.mkdir()) {
```
It might be that every call to File.createTempFile returns a file with the same name, say "junitFoo.tmp", yet "junitFoo" already exists. Then we do 10,000 rounds for nothing.

kcooney:
Calling `createTempFile` gives us a guaranteed unique path that no other thread or process can take, all without doing lots of looping (looping makes the test run time unpredictable so can result in flaky timeouts). Once we have a path we can remove the extension and create the directory. I doubt be would need any looping, but we can try a few times just in case.

kcooney:
According to[ the Javadoc](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#createTempFile(java.lang.String,%20java.lang.String,%20java.io.File)) `File.createTempFile()` won't return the same path every time.

varming:
Okay, you got me to look at the JDK code. When calling File.createTempFile JDK7 and later picks a random long X and check if junitX.tmp exists. If it exists then it starts over in a loop. Finally it attempts to create the file by calling into the JVM and from there into libc:open with flags O_CREAT | O_RDRW | O_EXCL (see http://hg.openjdk.java.net/jdk7u/jdk7u/jdk/file/f51368baecd9/src/share/classes/java/io/File.java).

In the NIO2 implementation of Files.createTempDirectory the JDK also picks a random long X and attempts to create a directory junitX. If it already exists then they loop (see http://hg.openjdk.java.net/jdk7u/jdk7u/jdk/file/f51368baecd9/src/share/classes/java/nio/file/TempFileHelper.java#l80)

I am doing pretty much the same as the NIO2 implementation, except I am using a nano timestamp instead SecureRandom.nextLong, and I bound the number of iterations, similar to what Guava does. I can switch to SecureRandom to be more like NIO2 if you like, but I fail to see any benefit of calling File.createTempFile before attempting to create the directory.

kcooney:
I really rather reuse the existing JDK implementation so we get the benefit of any bug fixes. So I would  prefer either the solution I gave above or your original solution (but with a limitation on the number of iteration).

varming:
Looks like Github is being very slow today. I pushed an update to use File.createTempFile as requested, but it will probably take a bit of time before it shows up. The current commit is: https://github.com/varming/junit4/commit/3c03251bbc458ced409018ca9b614f55a6df4af4

src/main/java/org/junit/rules/TemporaryFolder.java
```diff
@@ -216,10 +219,23 @@ public File newFolder() throws IOException {
...
-        File createdFolder = File.createTempFile("junit", "", parentFolder);
-        createdFolder.delete();
-        createdFolder.mkdir();
-        return createdFolder;
+        File createdFolder = null;
+        for (int i = 0; i < TEMP_DIR_ATTEMPTS; ++i) {
+            // Use createTempFile to get a suitable folder name.
+            final String suffix = ".tmp";
+            final File tmpFile = File.createTempFile(TMP_PREFIX, suffix, parentFolder);
+            tmpFile.delete();
```
kcooney: In my proposal, the temporary file was deleted after the temporary directory was created. That way we wouldn't have a recent condition if the user is using an older JDK where `File.createTempFile()` was sometimes returning paths that were returned from previous calls.

src/main/java/org/junit/rules/TemporaryFolder.java
```diff
@@ -216,10 +219,23 @@ public File newFolder() throws IOException {
...
 
     private File createTemporaryFolderIn(File parentFolder) throws IOException {
-        File createdFolder = File.createTempFile("junit", "", parentFolder);
-        createdFolder.delete();
-        createdFolder.mkdir();
-        return createdFolder;
+        File createdFolder = null;
+        for (int i = 0; i < TEMP_DIR_ATTEMPTS; ++i) {
+            // Use createTempFile to get a suitable folder name.
+            final String suffix = ".tmp";
```
kcooney: Style-nit: we don't mark local variables as `final` unless the compiler forces us to (like when the local is used in an anonymous class).

src/main/java/org/junit/rules/TemporaryFolder.java
varming: You pointed out that the specification of File.createTempFile says "Neither this method nor any of its variants will return the same abstract pathname again in the current invocation of the virtual machine". I checked JDK6 (it returns random filenames: http://hg.openjdk.java.net/jdk6/jdk6/jdk/file/b72e7b89dda9/src/share/classes/java/io/File.java#l1795), and GNU Classpath (they use a mix of a timestamp and a global counter to prevent returning the same name in the lifetime of the VM). Oracle's JDK5 is not public, but I doubt there were significant changes.

I'll remove the finals.

src/main/java/org/junit/rules/TemporaryFolder.java
```diff
@@ -216,10 +219,23 @@ public File newFolder() throws IOException {
...
-        File createdFolder = File.createTempFile("junit", "", parentFolder);
-        createdFolder.delete();
-        createdFolder.mkdir();
-        return createdFolder;
+        File createdFolder = null;
+        for (int i = 0; i < TEMP_DIR_ATTEMPTS; ++i) {
+            // Use createTempFile to get a suitable folder name.
+            String suffix = ".tmp";
+            File tmpFile = File.createTempFile(TMP_PREFIX, suffix, parentFolder);
+            tmpFile.delete();
```
kcooney: Please move this after creating a new temporary directory. It would protect us from race conditions  because no other JUnit thread or process could "steal" the path between the delete and the create (that is our theory as to the root cause of the bug).

Also, the code for `createTempFile()` did change.  See this blog post:

http://www.veracode.com/blog/2009/01/how-boring-flaws-become-interesting/

kcooney:
Sorry for yet more back and forth, but we keep getting into situations where I suggest something and you do something subtlety different. If you make the change I propose (and only that change) I can merge this tonight.

kcooney:
Merged. Thanks!

marcphilipp:
@varming Please add a description of this improvement to the [draft of the release notes](https://github.com/junit-team/junit4/wiki/4.13-Release-Notes).

PR #1305, opened 2016-05-24
Add ErrorCollector.checkThrows()

kcooney:


marcphilipp:
Nice! 👍 

I think it might be helpful if `checkThrows` returned the (expected) exception, like `Assert.expectThrows()` does. More detailed checks could then be performed using `ErrorCollector.checkThat()`. What do you think?

kcooney:
@marcphilipp What would it return if the lambda did not throw that exception?

marcphilipp:
Point taken.

marcphilipp:
LGTM! Feel free to merge (and add it to the release notes). 😄

kcooney:
Merged and added to the release notes

marcphilipp:
Thanks!

PR #1306, opened 2016-05-24
Improve JavaDoc of RunnerBuilder; Update reference to Google's code style

PeterWippermann:
Clarified the JavaDoc in RunnerBuilder that I found a bit confusing while debugging the code.
In the course of that change I also found that the URL to Google's code style was out-of-date.

### DIFF
--- a/src/main/java/org/junit/runners/model/RunnerBuilder.java
+++ b/src/main/java/org/junit/runners/model/RunnerBuilder.java
@@ -26,10 +26,10 @@ import org.junit.runner.Runner;
  *
  * <pre>
  * public TextFileSuite(Class testClass, RunnerBuilder builder) {
- *   // ...
- *   for (String className : readClassNames())
- *     addRunner(builder.runnerForClass(Class.forName(className)));
- *   // ...
+ *     // ...
+ *     for (String className : readClassNames())
+ *         addRunner(builder.runnerForClass(Class.forName(className)));
+ *     // ...
  * }
  * </pre>
  *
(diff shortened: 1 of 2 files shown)


kcooney:
While we do follow Google Java style, we use a four space indent, not a two space indent

src/main/java/org/junit/runners/model/RunnerBuilder.java
```diff
@@ -37,72 +37,73 @@
...
-     * @return a Runner
-     */
-    public Runner safeRunnerForClass(Class<?> testClass) {
-        try {
-            return runnerForClass(testClass);
-        } catch (Throwable e) {
-            return new ErrorReportingRunner(testClass, e);
-        }
+  /**
+   * Always returns a runner or null.
```
kcooney: In practice, I believe callers that call this won't see a `null` return value

PeterWippermann:
I adapted my code formatting and also put a remark in the "coding style" policy file

src/main/java/org/junit/runners/model/RunnerBuilder.java
PeterWippermann: @kcooney 
While debugging the code I saw `null` values.
This explains the null check in [AllDefaultPossibilitiesBuilder](https://github.com/junit-team/junit4/blob/24b8ee0bec2f2761b479bdd989275f19597955a3/src/main/java/org/junit/internal/builders/AllDefaultPossibilitiesBuilder.java#L25-L30)

kcooney:
It's a bit hard to see what you are changing with the entire file being reformatted at the same time. Please configure your IDE not to reformat lines that you didn't change.

In any case, it looks like the change is to document that `safeRunnerForClass()` can return `null`. While we do have internal implementations of `RunnerBuilder` that return `null`, no `RunnerBuilder` passed to a `Runner` constructor will return `null` (this is done by an undocumented collaboration between `AllDefaultPossibilitiesBuilder` and `JUnit4Builder`).

Yes, this is pretty ugly, and I wished the internal builders used a different (private) interface, but if we document that this can return `null` then the callers will wonder "what do I do if I get a `null`" until they eventually realize it can't happen.

PeterWippermann:
@kcooney Well, what if we add your statement to the JavaDoc then?
"no RunnerBuilder passed to a Runner constructor will return null"

kcooney:
@PeterWippermann Perhaps  "Note that some of the internal JUnit implementations of RunnerBuilder will return {@xode null} from this method,  but no RunnerBuilder passed to a Runner constructor will return {@xode null} from this method."

(Yuck. What a confusing API)

PeterWippermann:
@kcooney What do you think, ready for merging? ;-)

kcooney:
Merged. Thanks!

PR #1308, opened 2016-05-27
Refactor RuleChain: remove duplication

alb-i986:
The algorithm for running the rules in the chain is the same as the one in RunRules: re-use that!


kcooney:
Merged. Thanks!

Issue #1309, opened 2016-05-27
Regression introduced in TemporaryFolder?


alb-i986:
#1304 has possibly introduced a regression.

If you run our test suite, you get a tmp file leftover in the project root dir.

```
$ mvn clean test
[..]
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        junit6648638820331972951/

nothing added to commit but untracked files present (use "git add" to track)
```

I tracked down the unit tests causing that to be in TimeoutRuleTest:
- timeUnitTimeout
- longTimeout

Both run the test class AbstractTimeoutTest, which uses a TemporaryFolder.
Given the usage it makes:

```
@Rule
public final TemporaryFolder tmpFile = new TemporaryFolder();

@Test
public [..] {
[..]
File tmp = tmpFile.newFile();
[..]
```

.. I would expect the tmp file to be created in the system tmp dir.
But it is created in the project root dir.

New version:

      for (int i = 0; i < TEMP_DIR_ATTEMPTS; ++i) {
            // Use createTempFile to get a suitable folder name.
            String suffix = ".tmp";
            File tmpFile = File.createTempFile(TMP_PREFIX, suffix, parentFolder);
            String tmpName = tmpFile.getName();
            // Discard suffix of tmpName.
            String folderName = tmpName.substring(0, tmpName.length() - suffix.length());
            createdFolder = new File(parentFolder, folderName);
            if (createdFolder.mkdir()) {
                tmpFile.delete();
                return createdFolder;
            }
            tmpFile.delete();
        }

Old version:

          File createdFolder = File.createTempFile("junit", "", parentFolder);
          createdFolder.delete();
          createdFolder.mkdir();
         return createdFolder;

I guess the issue lies in the following line:

    createdFolder = new File(parentFolder, folderName);

In fact, while `File.createTempFile` in case of a null parent as arg, places the file in the system tmp dir, the constructor of File doesn't have such a behavior.

varming:
Oops. Looks like the File(String) constructor should have been used. Sorry.

marcphilipp:
@alb-i986 Good catch!

@varming Do you have time to provide a fix?

varming:
@marcphilipp I am working on it now.

kcooney:
Fixed via #1304

PR #1310, opened 2016-05-27
Fix regression with TemporaryFolder introduced by #1304

varming:
Looks like #1304 caused a regression where the location of the temporary folder created no longer ends up in the standard temporary directory. See issue #1309.

I added a test that checks the location of the created folder relative to the system property "java.io.tmpdir".


kcooney:
Merged. Thanks for the fix and the test!

PR #1312, opened 2016-05-31
Refactoring: remove duplication by adding Throwables#getStacktrace

alb-i986:


src/main/java/org/junit/internal/Throwables.java
```diff
@@ -39,4 +42,14 @@ public static Exception rethrowAsException(Throwable e) throws Exception {
     private static <T extends Throwable> void rethrow(Throwable e) throws T {
         throw (T) e;
     }
+
+    /**
+     * @return a String with the stacktrace of the given throwable
```
kcooney: Style-nit: remove the `@return` and instead give the method a description. The first sentence of a method's description shows up in the method summary section of the class Javadoc.

kcooney:
Nice. Thanks!

src/main/java/org/junit/internal/Throwables.java
```diff
@@ -39,4 +42,14 @@ public static Exception rethrowAsException(Throwable e) throws Exception {
...
+     */
+    public static String getStacktrace(Throwable exception) {
```
kcooney: Please add "@since 4.13" to the Javadoc

src/main/java/org/junit/internal/Throwables.java
alb-i986: True, thanks!

marcphilipp:
@alb-i986 Thanks!

PR #1313, opened 2016-05-31
RuleChain#around should not allow null args

alb-i986:
By throwing IllegalArgEx from around(),
we allow for better feedback to the final user,
as the stacktrace will point to the exact line where the null rule is declared.

Previously the stacktrace would have been something like this:

```
	at org.junit.rules.RunRules.applyAll(RunRules.java:26)
	at org.junit.rules.RunRules.<init>(RunRules.java:15)
	at org.junit.rules.RuleChain.apply(RuleChain.java:96)
	at org.junit.rules.RunRules.applyAll(RunRules.java:26)
	at org.junit.rules.RunRules.<init>(RunRules.java:15)
	at org.junit.runners.BlockJUnit4ClassRunner.withTestRules(BlockJUnit4ClassRunner.java:424)
	at org.junit.runners.BlockJUnit4ClassRunner.withRules(BlockJUnit4ClassRunner.java:378)
	at org.junit.runners.BlockJUnit4ClassRunner.methodBlock(BlockJUnit4ClassRunner.java:299)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:83)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:58)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
```

As you can see, it was pretty hard to figure out that the null pointer was one of the rules in the chain.

Whereas with this change it is:

```
java.lang.IllegalArgumentException: The enclosed rule should not be null

	at org.junit.rules.RuleChain.around(RuleChain.java:84)
	at org.junit.rules.RuleChainTest$RuleChainWithNullRules.<init>(RuleChainTest.java:81)  <-----
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at org.junit.runners.BlockJUnit4ClassRunner.createTest(BlockJUnit4ClassRunner.java:228)
	at org.junit.runners.BlockJUnit4ClassRunner.createTest(BlockJUnit4ClassRunner.java:238)
```

where `at org.junit.rules.RuleChainTest$RuleChainWithNullRules.<init>(RuleChainTest.java:81)` points at the line where the null rule is appended.

Follows the sample test class used above:

    public class RuleChainWithNullRules {
        @Rule
        public final RuleChain chain = outerRule(new LoggingRule("outer rule"))
                .around(null);

        @Test
        public void example() {}
    }

### DIFF
--- a/src/main/java/org/junit/rules/RuleChain.java
+++ b/src/main/java/org/junit/rules/RuleChain.java
@@ -77,8 +77,12 @@ public class RuleChain implements TestRule {
      *
      * @param enclosedRule the rule to enclose.
      * @return a new {@code RuleChain}.
+     * @throws IllegalArgumentException if {@code enclosedRule} is null
      */
     public RuleChain around(TestRule enclosedRule) {
+        if (enclosedRule == null) {
+            throw new IllegalArgumentException("The enclosed rule should not be null");
+        }
         List<TestRule> rulesOfNewChain = new ArrayList<TestRule>();
         rulesOfNewChain.add(enclosedRule);
         rulesOfNewChain.addAll(rulesStartingWithInnerMost);
--- a/src/main/java/org/junit/rules/RuleChain.java
+++ b/src/main/java/org/junit/rules/RuleChain.java
@@ -77,11 +77,11 @@ public class RuleChain implements TestRule {
      *
      * @param enclosedRule the rule to enclose.
      * @return a new {@code RuleChain}.
-     * @throws IllegalArgumentException if {@code enclosedRule} is null
+     * @throws NullPointerException if the argument {@code enclosedRule} is null
      */
     public RuleChain around(TestRule enclosedRule) {
         if (enclosedRule == null) {
-            throw new IllegalArgumentException("The enclosed rule should not be null");
+            throw new NullPointerException("The enclosed rule should not be null");
         }
         List<TestRule> rulesOfNewChain = new ArrayList<TestRule>();
         rulesOfNewChain.add(enclosedRule);
--- a/src/main/java/org/junit/rules/RuleChain.java
+++ b/src/main/java/org/junit/rules/RuleChain.java
@@ -75,13 +75,13 @@ public class RuleChain implements TestRule {
      * Create a new {@code RuleChain}, which encloses the given {@link TestRule} with
      * the rules of the current {@code RuleChain}.
      *
-     * @param enclosedRule the rule to enclose.
+     * @param enclosedRule the rule to enclose; must not be {@code null}.
      * @return a new {@code RuleChain}.
-     * @throws NullPointerException if the argument {@code enclosedRule} is null
+     * @throws NullPointerException if the argument {@code enclosedRule} is {@code null}
      */
     public RuleChain around(TestRule enclosedRule) {
         if (enclosedRule == null) {
-            throw new NullPointerException("The enclosed rule should not be null");
+            throw new NullPointerException("The enclosed rule must not be null");
         }
         List<TestRule> rulesOfNewChain = new ArrayList<TestRule>();
         rulesOfNewChain.add(enclosedRule);
--- a/src/test/java/org/junit/rules/RuleChainTest.java
+++ b/src/test/java/org/junit/rules/RuleChainTest.java
@@ -17,6 +17,7 @@ import java.util.List;
 
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.internal.Throwables;
 import org.junit.runner.Description;
 import org.junit.runner.JUnitCore;
 import org.junit.runner.Result;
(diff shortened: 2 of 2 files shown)


src/main/java/org/junit/rules/RuleChain.java
```diff
@@ -77,8 +77,12 @@ private RuleChain(List<TestRule> rules) {
      *
      * @param enclosedRule the rule to enclose.
      * @return a new {@code RuleChain}.
+     * @throws IllegalArgumentException if {@code enclosedRule} is null
      */
     public RuleChain around(TestRule enclosedRule) {
+        if (enclosedRule == null) {
+            throw new IllegalArgumentException("The enclosed rule should not be null");
```
kcooney: We generally throw `NullPointerException` if someone passes in a `null` for a parameter that shouldn't be `null` (this is also what the Java Collections library does and what Effective Java recommends).

alb-i986: I see. Done, although I really prefer IllegalArgumentEx as I find it more meaningful than a NPE: NPEs can be thrown by the JVM, whereas IllegalArgEx cannot.
For example, take the unit test: since it's now considering a NPE as expected, I had to add an assertion on the message to make sure that the NPE is really what we are expecting, and not some other null reference (a programming error).

src/test/java/org/junit/rules/RuleChainTest.java
```diff
@@ -55,4 +63,41 @@ public void executeRulesInCorrectOrder() throws Exception {
...
+    public void whenRuleChainHasNullRuleTheStacktraceShouldPointToIt() {
+        Result result = JUnitCore.runClasses(RuleChainWithNullRules.class);
+
+        assertThat(result.getFailures().size(), equalTo(1));
+        String stacktrace = stacktraceToString(result.getFailures().get(0).getException());
+        assertThat(stacktrace, containsString("\tat org.junit.rules.RuleChainTest$RuleChainWithNullRules.<init>(RuleChainTest.java:"));
+    }
+
+    // TODO delete as soon as #1312 is merged
+    private static String stacktraceToString(Throwable e) {
```
alb-i986: In case you are going to merge #1312 soon, please merge that before this, so that I can remove this method.

alb-i986:
Ready

src/main/java/org/junit/rules/RuleChain.java
```diff
@@ -77,8 +77,12 @@ private RuleChain(List<TestRule> rules) {
      *
      * @param enclosedRule the rule to enclose.
```
marcphilipp: Please change this line to `@param enclosedRule the rule to enclose; must not be {@code null}.`

src/main/java/org/junit/rules/RuleChain.java
```diff
@@ -77,8 +77,12 @@ private RuleChain(List<TestRule> rules) {
...
      * @return a new {@code RuleChain}.
+     * @throws NullPointerException if the argument {@code enclosedRule} is null
```
marcphilipp: Please change this line to `@throws NullPointerException if the argument {@code enclosedRule} is {@code null}

src/main/java/org/junit/rules/RuleChain.java
```diff
@@ -77,8 +77,12 @@ private RuleChain(List<TestRule> rules) {
...
      */
     public RuleChain around(TestRule enclosedRule) {
+        if (enclosedRule == null) {
+            throw new NullPointerException("The enclosed rule should not be null");
```
marcphilipp: Please replace "should" with "must".

alb-i986:
Any other concerns on this one?

marcphilipp:
LGTM, thanks!

@junit-team/junit-committers Please take a look!

stefanbirkner:
LGTM

marcphilipp:
Thanks!

@alb-i986 Please add a description of this improvement to the [draft of the release notes](https://github.com/junit-team/junit4/wiki/4.13-Release-Notes).

alb-i986:
> Please add a description of this improvement to the draft of the release notes.

Done:
https://github.com/junit-team/junit4/wiki/4.13-Release-Notes#pull-request-1313-rulechainaround-rejects-null-arguments

marcphilipp:
Thanks! 👍

Issue #1178, opened 2015-07-07
Null Pointer exception assertArrayEquals with no message.


EmteeW:
When assertArrayEquals (at least for Boolean[]) fails the assert, it throws a null pointer exception looking for a message.  This is, of course, when using the prototype that has no message.  You'll still get the line number for the failure, you'll still probably just debug focused test method, and be on your merry way, but just technically it's not quite the best/correct behavior.

Thanks!

Update: this is for 4.12

marcphilipp:
Can you please provide a test case to illustrate your problem?

EmteeW:
@Test
    public void testJunit() {
        assertArrayEquals(new Boolean[]{true}, new Boolean[]{false});
    }

marcphilipp:
```java
package sandbox;

import static org.junit.Assert.assertArrayEquals;

import org.junit.Test;
import org.junit.runner.JUnitCore;
import org.junit.runner.Result;

public class Snippet {
	@Test
	public void testJunit() {
		assertArrayEquals(new Boolean[] { true }, new Boolean[] { false });
	}
	
	public static void main(String[] args) {
		Result result = JUnitCore.runClasses(Snippet.class);
		result.getFailures().get(0).getException().printStackTrace(System.out);
	}
}
```

Running this as a Java application prints 

```
arrays first differed at element [0]; expected:<true> but was:<false>
	at org.junit.internal.ComparisonCriteria.arrayEquals(ComparisonCriteria.java:55)
	at org.junit.Assert.internalArrayEquals(Assert.java:532)
	at org.junit.Assert.assertArrayEquals(Assert.java:283)
	at org.junit.Assert.assertArrayEquals(Assert.java:298)
	at sandbox.Snippet.testJunit(Snippet.java:12)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
[...]

Issue #976, opened 2014-08-13
ComparisonFailure reports all failures as "expected: null<null> but was: null<null>"


dsaff:
Still debugging locally, but since my local gradle build picked up 4.12-beta-1, all string comparison failures are reporting

org.junit.ComparisonFailure: expected: null<null> but was: null<null>
        at org.junit.Assert.assertEquals(Assert.java:115)
        at org.junit.Assert.assertEquals(Assert.java:144)

This is on comparison of two obviously non-null strings, for example assertEquals("a", "b").

This appears to be affecting our whole local team, but still may be something odd about our setup.  Continuing to investigate.

navels:
FYI, you missed ArrayComparisonFailure.fCause, which disappeared in 4.12.  This caused much hair-pulling for us as it caused a gradle test task to fail with no apparent cause.  We had upgraded to gradle 2.4 which bundles junit 4.12 whereas our application classpath pointed to 4.11.

PR #1315, opened 2016-05-31
ArrayComparisonFailure serialization incompatibility fix:

aishahalim:
Add back field fCause, initialize and use in the constructor (via initCause()) to avoid npe and unused field warnings, respectively
Issue #1178, #976


marcphilipp:
Thanks for the pull request. On second thought I think we need to do a little more.

As you have verified in https://github.com/junit-team/junit4/issues/1178#issuecomment-222830227 this will ensure that pre 4.12 versions can consume post 4.12 versions' `ArrayComparisonFailures`. However, this does not hold the other way around. We should override `getCause()` similar to this:

```java
public synchronized Throwable getCause() {
    return super.getCause() == null ? fCause : super.getCause();
}
```

I agree that this is hard to test. However, I think the following should be ~~fairly easy~~ doable (@kcooney has done the same for `Result` in f6149a9236d9a2778bf2fd6db927fd9f498a79c8):

1. Create a `ArrayComparisonFailure` using 4.11 and serialize it into a file.
2. Put the file into `src/test/resources` and add a test that reads it, deserializes it and checks the resulting `ArrayComparisonFailure`

Do the same for 4.12.

@aishahalim Please let me know if you're okay with doing that.

aishahalim:
> However, this does not hold the other way around.

Yikes, yes I see it now.

Thanks for pointing out how to construct that test! On it!

aishahalim:
Hey Marc!

I've update this pullrequest with another commit. Your test strategy description, when implemented in the test, very apparently caught the forward incompatibility problem with getCause() not falling back to fCause.

I threw in a description in the test to explain what it's interacting with since super wordy names don't fill in all the serializing to a file setup, and why.

What do you think about it so far?

I didn't mull over it long enough, but I don't see how I could test the other way, where I have a current version of an exception deserializing with an older version of the class. Is this implemented in https://github.com/junit-team/junit4/commit/f6149a9236d9a2778bf2fd6db927fd9f498a79c8? I threw a breakpoint in when running ResultTest and still didn't see anything that tried to serialize to a special version of a class, unless you were referring to the fixed runtime sub class -- i.e. subclassing ArrayComparisonFailure to behave like the previous versions', then testing compatibility with the current version's instance of the failure?

Apologies for all these questions.

marcphilipp:
Looks good to me, you did exactly what I had in mind! 👍 

@junit-team/junit-committers What do you think, is this good to merge?

PeterWippermann:
Any updates on this?

aishahalim:
^ It might be a bit of a hefty changeset for a not-at-all-critical bug, so, understandably, second review might take longer.

Let me know if I should split out any functional bits in separate commits for reviewing succinctness or whatnot.

marcphilipp:
@junit-team/junit-committers I'd like a second opinion on this. Can one of you guys please have a look?

kcooney:
LGTM. @marcphilipp was there something specific you were concerned about?

marcphilipp:
> was there something specific you were concerned about?

@kcooney Basically I just wanted to have another pair of eyes look at it to decrease the probability that we've overlooked something. :-)

marcphilipp:
@aishahalim Thanks again! 👍 Please add a description of this fix to the [draft of the release notes](https://github.com/junit-team/junit4/wiki/4.13-Release-Notes).

aishahalim:
Thanks @marcphilipp :D

Added the description here: https://github.com/junit-team/junit4/wiki/4.13-Release-Notes#pull-request-1315-assertarrayequals-shouldnt-throw-an-npe-when-test-suites-are-compiledrun-across-versions-of-junit

Thanks a bunch all!

PeterWippermann:
Well done, @aishahalim! :+1:

PR #1319, opened 2016-06-01
fix #1295

### LINKED ISSUE #1295: Javadoc for RuleChain contains errors
* The class modifier `static` is not necessary.
* The method calls are missing closing parentheses.

In addition, the documentation should make clear that the `RuleChain` cannot be used to define the order of existing rules, e.g. the following does not work as expected:

```
@Rule
public LoggingRule middle = new LoggingRule("middle rule");
@Rule
public RuleChain chain = RuleChain
    .outerRule(new LoggingRule("outer rule"))
    .around(middle)
    .around(new LoggingRule("inner rule"));
```

In this code, `middle` is executed twice.

Harrisnick:
Removed error from `RuleChain` Javadoc and clarified how it works with existing rules:

- Removed `static` modifier
- Added missing closing parenthesis of method calls
- Added clarification

### DIFF
--- a/src/main/java/org/junit/rules/RuleChain.java
+++ b/src/main/java/org/junit/rules/RuleChain.java
@@ -13,12 +13,12 @@ import org.junit.runners.model.Statement;
  * {@link #around(TestRule)}:
  *
  * <pre>
- * public static class UseRuleChain {
+ * public class UseRuleChain {
  * 	&#064;Rule
  * 	public RuleChain chain= RuleChain
- * 	                       .outerRule(new LoggingRule("outer rule")
- * 	                       .around(new LoggingRule("middle rule")
- * 	                       .around(new LoggingRule("inner rule");
+ * 	                       .outerRule(new LoggingRule("outer rule"))
+ * 	                       .around(new LoggingRule("middle rule"))
+ * 	                       .around(new LoggingRule("inner rule"));
  *
  * 	&#064;Test
  * 	public void example() {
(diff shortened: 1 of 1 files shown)


src/main/java/org/junit/rules/RuleChain.java
```diff
@@ -37,6 +37,22 @@
  * finished middle rule
  * finished outer rule
  * </pre>
+ * 
+ * {@code RuleChain} cannot be used to define the order of existing rules.</br>
```
marcphilipp: Please remove this `</br>` tag.

Harrisnick:
</br> tag removed

src/main/java/org/junit/rules/RuleChain.java
Harrisnick: </br> tag removed

src/main/java/org/junit/rules/RuleChain.java
```diff
@@ -37,6 +37,22 @@
...
+ * &#064;Rule
+ * public LoggingRule middle = new LoggingRule("middle rule");
+ * 
+ * &#064;Rule
+ * public RuleChain chain = RuleChain
+ *                          .outerRule(new LoggingRule("outer rule"))
+ *                          .around(middle)
+ *                          .around(new LoggingRule("inner rule"));
+ * </pre>
+ * <p>
```
marcphilipp: This `<p>` tag should be at the beginning of the paragraph not its end, shouldn't it?

Harrisnick: I doubt that the `<p>` tag is required at all. It doesn't add anything to the layout really. I will remove it.

marcphilipp: It's usually a good idea to take a look at the generated Javadoc locally. You can generate it using Maven. Let me know if you need more detailed instructions.

Harrisnick: I assumed that the <p> tag was required to add some spacing at the end - this clearly wasn't the case. I used the Javadoc view in eclipse to check my changes. Now I also checked the Javadocs generated by maven and it looks alright.

marcphilipp:
@Harrisnick Thanks! Please add a description of this improvement to the [draft of the release notes](https://github.com/junit-team/junit4/wiki/4.13-Release-Notes).

Issue #1323, opened 2016-06-03
Theories and Assumptions doesn't work as expected


SeriyBg:
Since Theories doc says:
> If any of the assumptions fail, the data point is silently ignored

And Assume doc says:
> A good example of using assumptions is in Theories where they are needed to exclude certain datapoints that aren't suitable or allowed for a certain test case.

I'm expecting @Theory to be ignored if assumption doesn't pass the condition. But instead it fails the test.

This test fails, while I'm expecting it to be ignored.
```
@RunWith(Theories.class)
public class SnippetTest {

   @Theory
   public void validateIndices(){
	   assumeTrue(false);
	   assertTrue(false);
   }
}
```

I'm using JUnit version 4.12.

SeriyBg:
Thax to comment on SO I found what is the issue. Test if failed only if all assertion fail and is ignored if at least one asserion passed.
So this test fails
```
@RunWith(Theories.class)
public class SnippetTest {

    @DataPoints
    public static boolean[] data(){
        return new boolean[]{false, false};
    }

   @Theory
   public void validateIndices(boolean data){
       assumeTrue(data);
       assertTrue(true);
   }
}
```
But this one pass
```
@RunWith(Theories.class)
public class SnippetTest {

    @DataPoints
    public static boolean[] data(){
        return new boolean[]{false, true};
    }

   @Theory
   public void validateIndices(boolean data){
       assumeTrue(data);
       assertTrue(true);
   }
}
```

Looks like right behavior for me, but it's not obvious from documentation. It would be great to explaint this behaviour in docs.

marcphilipp:
Documentation will be improved in #1324.

PR #1324, opened 2016-06-03
Add javadoc explanation for case when all assumption doesn't pass the…

SeriyBg:
It was not clear for me why test fails when all the assumptions fails. I created an issue #1323 since I thought it's a bug. Lately I find out why It's failed, so I add the explanation to javadoc to make it clear.


marcphilipp:
Thanks for the pull request!

I would phrase it more like the error message, 

> If no parameters can be found that satisfy all assumptions, the test fails.

Do you mind changing it accordingly?

SeriyBg:
Thank you for your suggestion. Totally agree.

marcphilipp:
Thanks!

PR #1331, opened 2016-06-22
Add throws annotation

phoenix384:
The protected methods of TestWatcher should have a 'throws Throwable' annotation since they all are called inside a try-catch catching Throwable. With this throws annotation checked exceptions can be handled also.

### DIFF
--- a/src/main/java/org/junit/rules/TestWatcher.java
+++ b/src/main/java/org/junit/rules/TestWatcher.java
@@ -122,19 +122,19 @@ public abstract class TestWatcher implements TestRule {
     /**
      * Invoked when a test succeeds
      */
-    protected void succeeded(Description description) {
+    protected void succeeded(Description description) throws Throwable {
     }
 
     /**
      * Invoked when a test fails
      */
-    protected void failed(Throwable e, Description description) {
+    protected void failed(Throwable e, Description description) throws Throwable {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


kcooney:
The Javadoc of the class says: "TestWatcher is a base class for Rules that take note of the testing action, without modifying it." So I don't think the intention of the class was to provide a mechanism to make the test fail. The exception handling is there to report badly behaving implementations.

In any case, this is a potentially breaking change since subclasses could , for example, override `succeeded()` and have their overridden method call `super.succeeded()`

marcphilipp:
I'm closing this pull request since this change would break backwards compatibility.

phoenix384:
The intention is not to have a mechanism making the test fail. Let's say you want to write something to a file that depends on the test result. Although it will probably never happen, this could throw an IOException. So what to do it in order to not let in go unnoticed? Wrap it into a RuntimeException? Ugly.
So perhaps it is the best way to report it via a test failure.

> In any case, this is a potentially breaking change since subclasses could , for example, override succeeded() and have their overridden method call super.succeeded()

It's not. Why should anyone call super.succeeded()? This method does nothing.

kcooney:
Having to wrap checked exceptions with runtime exceptions here is actually a good thing. It's a reminder that your rule could throw exceptions that affect the outcome of the test.

PR #1335, opened 2016-06-28
Fix ExternalResource: the test failure was lost..

### LINKED ISSUE #1334: ExternalResource: the behavior is different than using multiple After methods
Multiple issues here:
1. When the test fails _and_ closing the resource fails, the test failure is lost: only the exception coming from the after() method is propagated.
2. In case there are multiple (ExternalResource) rules in place, and (closing) one fails, then all of the subsequent resources will _not_ be closed (without even trying to!).

This was not the case when encapsulating the tear down logic in multiple `@After` methods: RunAfters runs all of the After methods regardless of the result of each, and throws a MultipleFailureException in case more than one thing (test method and/or after logic) goes wrong.

I'm working on a fix for 1 for now, which is an easy one.

alb-i986:
.. when the test failed *and* closing the resource failed:
only the exception coming from the after() method was propagated, as per semantics of the try-finally
(see http://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.20.2).

Finally, the new behavior is compatible with `@After` method semantics (see class RunAfters).

Fixes #1334, point 1.

### DIFF
--- a/src/main/java/org/junit/rules/ExternalResource.java
+++ b/src/main/java/org/junit/rules/ExternalResource.java
@@ -1,7 +1,6 @@
 package org.junit.rules;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 
 import org.junit.runner.Description;
--- a/src/main/java/org/junit/rules/ExternalResource.java
+++ b/src/main/java/org/junit/rules/ExternalResource.java
@@ -1,6 +1,11 @@
 package org.junit.rules;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
 import org.junit.runner.Description;
+import org.junit.runners.model.MultipleFailureException;
 import org.junit.runners.model.Statement;
 
 /**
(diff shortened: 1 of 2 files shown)


marcphilipp:
I don't like the duplication with `RunAfters` but besides that it looks good to me.

@junit-team/junit-committers Please take a look!

src/main/java/org/junit/rules/ExternalResource.java
```diff
@@ -44,11 +48,20 @@ private Statement statement(final Statement base) {
             @Override
             public void evaluate() throws Throwable {
                 before();
+
+                List<Throwable> errors = new ArrayList<Throwable>();
                 try {
                     base.evaluate();
+                } catch(Throwable t) {
```
kcooney: Style-nit : please put a space after "catch" (we use Google Java style which requires spaces around keywords like "if" and "catch"

kcooney:
LGTM other than very minor style issues

src/main/java/org/junit/rules/ExternalResource.java
alb-i986: Fixed, thanks.

marcphilipp: Did you forget to push?

src/main/java/org/junit/rules/ExternalResource.java
```diff
@@ -44,11 +48,20 @@ private Statement statement(final Statement base) {
...
+                    errors.add(t);
                 } finally {
-                    after();
+                    try {
+                        after();
+                    } catch(Throwable t) {
```
kcooney: same issue here, too

kcooney:
I fixed the style issues and merged it. Thanks!

alb-i986:
@kcooney would you please update the release notes?

Done. Sorry for the delay.

PR #1336, opened 2016-07-04
Add TestFixture

kcooney:
This is a prototype of test fixtures, an extension point designed to replaced rules. The core class is a new `TestFixture` interface that can be implemented in users' tests or put in a common location to be shared  by multiple tests.

The `TestFixture` API was designed so that it could be supported by JUnit Vintage tests as well as JUnit Jupiter tests. In the first commit I show integration with Vintage tests. In the second commit, as a proof of concept, I updated `TestCase` to support `TestFixture` (I figured if JUnit Legacy tests could support them, anything could).

Although designed as a possible replacement for rules, test fixtures are in some ways less powerful, but in other ways more powerful. A fixture can cause code to be executed 1) before a test (or suite), 2) after a test (or suite) and/or 3) between the successful completion of the test (or suite) and any tear down calls (`@After`,etc) have been called. Unlike rules, however, they cannot be used to cause a failing test to pass, or to run a test multiple times.

Test fixtures are also easy to compose. Here is an example of a test fixture using another one:

```
public class TestWithWebDriver implements TestFixture {
  private final ServerStarter server = new ServerStarter(); // another fixture
  private final WebDriver webDriver;

  public void initialize(FixtureContext context) throws Exception {
    server.initialize(context); // start the server
    webDriver = new FirefoxDriver();
    driver.get("http://localhost+" + server.getPort()); // open the home page
  }
}
```

Here's an example how one might might run two fixtures in order:

```
private final TestFixture firstFixture = ...
private final TestFixture secondFixture = ...

@Fixture
public final TestFixture initializeInOrder = new TestFixture() {

  public void initialize(FixtureContext context) throws Exception {
    firstFixture.initialize(context);
    secondFixture.initialize(context);
  }
}
```

See the `TemporaryDirectory` class for an example fixture.

As it happens, test fixtures are also easier to test than rules are. Note that `TemporaryDirectory` does not have any methods that were added only for testing purposes, but the tests in `TemporaryDirectoryTest` have the same coverage.

Note that this change is missing many tests, and would likely require some framework. I just wanted to send this as a concrete proposal for a simple but powerful extension point that can be supported by multiple styles of JUnit tests.

### DIFF
--- a/src/main/java/junit/framework/TestCase.java
+++ b/src/main/java/junit/framework/TestCase.java
@@ -3,6 +3,14 @@ package junit.framework;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.fixtures.FixtureContext;
+import org.junit.fixtures.FixtureManager;
+import org.junit.fixtures.TearDown;
+import org.junit.fixtures.TestFixture;
+import org.junit.runners.model.MultipleFailureException;
 
 /**
  * A test case defines the fixture to run multiple tests. To define a test case<br/>
@@ -80,6 +88,10 @@ public abstract class TestCase extends Assert implements Test {
      */
     private String fName;
 
+    private FixtureManager fixtureManager = new FixtureManager();
+    private boolean testStarted = false;
+    private final List<TestFixture> testFixtures = new ArrayList<TestFixture>();
+
     /**
      * No-arg constructor to enable serialization. This method
      * is not intended to be used by mere mortals without calling setName().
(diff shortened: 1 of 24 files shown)


kcooney:
@marcphilipp FYI I added a commit with improved Javadoc. Also, now the fixtures have access to the class and method annotations.

kcooney:
@marcphilipp any more thoughts on this proposal?

As I said in the JUnit5 issue, I could do this outside of JUnit but I would need a way to install global rules. In the past @dsaff wasn't thrilled with global rules (in part because would be nothing in the test indicating that the test has additional behavior applied to it). Maybe you feel differently?

marcphilipp:
I've now (finally) responded to your comment in https://github.com/junit-team/junit5/issues/169#issuecomment-246431360. Sorry for the delay!

kcooney:
Created pull request to add the core classes to opentest4j: https://github.com/ota4j-team/opentest4j/pull/26

kcooney:
Moved to opentest4j

Issue #396, opened 2012-03-08
If @Before throws an exception, then @After of the other subclasses should not be executed


viggyprabhu:
Hi,

I have 3 classes such that A extends from B which extends from C. Each one of them have a @Before and @After block. If @Before block in C actually throws an exception, then @After blocks from A and B should not be executed and only @After block from C should be executed. 
In my particular case, I was creating a folder in @Before and in @After writing a file in the folder in class B. Now since @Before at C threw an exception, @Before of B did not get executed but it executed @After of B which caused another exception as the folder it expected did not exists.
What do you suggest?

Regards,

Vignesh

PR #1340, opened 2016-07-13
Introduce class TestRuleAdapter

saschpe:
Allows to wrap method rules in test rules so that they can be used
inside rule chains. Not every third party project ported their method
rules yet (or even wants to).

Inspired by Johan Haleby's solution to PowerMock issue #396

### DIFF
--- a/src/main/java/org/junit/rules/TestRuleAdapter.java
+++ b/src/main/java/org/junit/rules/TestRuleAdapter.java
@@ -1,0 +1,73 @@
+package org.junit.rules;
+
+import org.junit.runner.Description;
+import org.junit.runners.model.FrameworkMethod;
+import org.junit.runners.model.Statement;
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/junit/rules/TestRuleAdapter.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    private Class<?> getTestClass(Description description) {
+        return description.getTestClass();
+    }
+
+    private Object getTestObject(Description description) {
+        try {
+            return getTestClass(description).newInstance();
```
marcphilipp: IMHO creating a new instance of the test class is no solution. What if the method rule wants to interact with the actual test class instance?

kcooney: Agreed.

PR #1343, opened 2016-07-16
Replace uses of synchronized(something) with ReentrantLock.

kcooney:
In almost all cases, this is more efficent, especially if there is any
contention on the locks.

### DIFF
--- a/src/main/java/org/junit/internal/requests/ClassRequest.java
+++ b/src/main/java/org/junit/internal/requests/ClassRequest.java
@@ -1,11 +1,14 @@
 package org.junit.internal.requests;
 
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
 import org.junit.internal.builders.AllDefaultPossibilitiesBuilder;
 import org.junit.runner.Request;
 import org.junit.runner.Runner;
 
 public class ClassRequest extends Request {
-    private final Object runnerLock = new Object();
+    private final Lock runnerLock = new ReentrantLock();
 
     /*
      * We have to use the f prefix, because IntelliJ's JUnit4IdeaTestRunner uses
@@ -28,10 +31,13 @@ public class ClassRequest extends Request {
     @Override
     public Runner getRunner() {
         if (runner == null) {
-            synchronized (runnerLock) {
+            runnerLock.lock();
+            try {
                 if (runner == null) {
                     runner = new AllDefaultPossibilitiesBuilder(canUseSuiteMethod).safeRunnerForClass(fTestClass);
                 }
+            } finally {
+                runnerLock.unlock();
             }
         }
         return runner;
--- a/src/main/java/org/junit/runners/ParentRunner.java
+++ b/src/main/java/org/junit/runners/ParentRunner.java
@@ -12,6 +12,8 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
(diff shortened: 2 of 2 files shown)


marcphilipp:
Are you sure this is an improvement? It doesn't really use any of the advantages (API-wise) of ReentrantLock, does it? Is it really more efficient? It sure requires a couple of lines of additional code.

kcooney:
It is more efficient. For `synchronized(object)` if there is any contention in any synchronized block that uses an object of the same class, the synchronized block will get slower, even if that particular block is uncontested.

kcooney:
The alternative, to get slightly better performance while having a simpler locking API is to change this:

    private final Object runnerLock = new Object();

to this:

    private final Object runnerLock = new Object() {};

That way, each class that has a lock uses a different class for the monitor object, so contention in the locks in one class won't affect the locks in the other class

marcphilipp:
SGTM!

kcooney:
@marcphilipp what sounds good? Merging this pull or my comment starting with "The alternative" ?

marcphilipp:
I was referring to the alternative.

panchenko:
I am kind of curious on what's the source of "It is more efficient" information. And if contention in this places is high enough.

panchenko:
Also on the alternative: why "new Object() {}" is supposed to be faster? Is this defined by java spec or if such behavior was experienced by some java implementation on the particular system?

kcooney:
@panchenko I provided a link to some general performance results. I also have other sources but unfortunately I cannot forward them.

`ReentrantLock` is really the preferred way to do this. I could change the type of the variables use the `Lock` interface to alleviate the concern that we aren't using all the methods on `ReentrantLock`

smallyard:
See: http://www.ibm.com/developerworks/java/library/j-jtp10264/?S_TACT=105AGX52&S_CMP=cn-a-j

kcooney:
"new Object() {}" can be faster because the JVM can use the class of the object to determine if the lock has ever been contested. This kind of makes sense. If you use `synchronized` methods in your class, and the JVM sees contention in the intrinsic lock of the object, it's likely to see contention on other locks for that class. So if you synchronize on "new Object()" and there is contention anywhere in your application on a lock that uses "new Object()" then the JVM could assume that your lock could have contention, and avoid thin locks or biased locking.

panchenko:
I've tried tests from the 2nd mechanical-sympathy post http://mechanical-sympathy.blogspot.com/2011/11/biased-locking-osr-and-benchmarking-fun.html in java 8:
* if synchronized block contains only "++counter" then with ReentrantLock it is 10..25% faster
* specifically with 2 threads ReentrantLock is slower (as in the post)

If doing a bit more work instead of ++counter then difference stays at ~10%. It seems like if doing significantly more work (including object allocation, etc) with less contention then it would be quite hard to notice any difference.

kcooney:
@panchenko That's not a great benchmark. It's entirely possible that the JVM will optimize those locks away after warm-up since they are only protecting a local variable and the result is never used. Also, the measured time includes starting threads (!) and waiting for the joins.

Writing good micro-benchmarks is hard.

Going back to this change, it's highly unlikely that the changes to `ClassRequest` would be noticeable, since most users call `Request.aClass()` at most once per run. The `ParentRunner` change could be more noticeable. Since lock contention tracking is done by type, I would like to either use `ReentrantLock` (as recommended in the links in this thread) or `new Object() {}`. I suspect most readers of the code would be puzzled by the use of `new Object() {}`

kcooney:
I'm merging this. The alternative (using new `Object() {}`) is non-obvious, and the performance of `ReentrantLock` is better in almost all benchmarks listed in this thread.

PR #1346, opened 2016-07-18
Optimize string format

smallyard:
In format strings, it is generally preferable better to use %n, which will produce the platform-specific line separator.


marcphilipp:
I don't have a strong opinion about this one.

@junit-team/junit-lambda What do you think?

marcphilipp:
Oops, I mistook this issue for an opentest4j one (now there is one over there, too: ota4j-team/opentest4j#24).

@junit-team/junit-committers What do you think about this one?

kcooney:
The commit message is a bit weird (this isn't an optimizationn) but we can fix that when we merge the commit if we use `hub`. The change is fine

PR #1349, opened 2016-07-22
Introduce class TestRuleAdapter / RuleChain support for MethodRule

saschpe:
Allows to wrap method rules in test rules so that they can be used
inside rule chains. Not every third party project ported their method
rules yet (or even wants to).

Add respective overrides for RuleChain.outerRule and RuleChain.around
to simplify usage.

Inspired by Johan Haleby's solution to PowerMock issue #396

### DIFF
--- a/src/main/java/org/junit/rules/RuleChain.java
+++ b/src/main/java/org/junit/rules/RuleChain.java
@@ -1,9 +1,11 @@
 package org.junit.rules;
 
+import java.beans.MethodDescriptor;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
+import org.junit.Rule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 
@@ -82,6 +84,19 @@ public class RuleChain implements TestRule {
         return emptyRuleChain().around(outerRule);
     }
 
+    /**
+     * Returns a {@code RuleChain} with a single {@link MethodRule}. This method
+     * is the usual starting point of a {@code RuleChain}.
+     *
+     * @param outerRule the outer rule of the {@code RuleChain}.
+     * @param testObject instance of the current test class.
+     * @return a {@code RuleChain} with a single {@link TestRule}.
+     * @since 4.12
+     */
+    public static RuleChain outerRule(MethodRule outerRule, Object testObject) {
+        return emptyRuleChain().around(outerRule, testObject);
+    }
+
     private RuleChain(List<TestRule> rules) {
         this.rulesStartingWithInnerMost = rules;
     }
@@ -100,6 +115,26 @@ public class RuleChain implements TestRule {
         }
         List<TestRule> rulesOfNewChain = new ArrayList<TestRule>();
         rulesOfNewChain.add(enclosedRule);
+        rulesOfNewChain.addAll(rulesStartingWithInnerMost);
+        return new RuleChain(rulesOfNewChain);
+    }
+
+    /**
+     * Create a new {@code RuleChain}, which encloses the given {@link MethodRule} with
+     * the rules of the current {@code RuleChain}.
+     *
+     * @param enclosedRule the rule to enclose; must not be {@code null}.
+     * @param testObject instance of the current test class.
+     * @return a new {@code RuleChain}.
+     * @throws NullPointerException if the argument {@code enclosedRule} is {@code null}
+     * @since 4.12
+     */
+    public RuleChain around(MethodRule enclosedRule, Object testObject) {
+        if (enclosedRule == null) {
+            throw new NullPointerException("The enclosed rule must not be null");
+        }
+        List<TestRule> rulesOfNewChain = new ArrayList<TestRule>();
+        rulesOfNewChain.add(new TestRuleAdapter(enclosedRule, testObject));
         rulesOfNewChain.addAll(rulesStartingWithInnerMost);
         return new RuleChain(rulesOfNewChain);
     }
--- a/src/main/java/org/junit/rules/TestRuleAdapter.java
+++ b/src/main/java/org/junit/rules/TestRuleAdapter.java
@@ -1,0 +1,75 @@
+package org.junit.rules;
+
+import org.junit.runner.Description;
+import org.junit.runners.model.FrameworkMethod;
+import org.junit.runners.model.Statement;
+
+import java.lang.reflect.Method;
+
+/**
+ * A TestRuleAdapter allows to wrap {@link MethodRule} classes in
+ * {@link TestRule} classes in order to use them inside {@link RuleChain}.
+ * Assuming we have an existing method rule:
+ *
+ * <pre>
+ * class MyMethodRule implements MethodRule {
+ *     public Statement apply(final Statement base, FrameworkMethod method,
+ *                            Object target) {
+ *         return new Statement() {
+ *             &#064;Override
+ *             public void evaluate() throws Throwable {
+ *                 base.evaluate();
+ *             }
+ *         };
+ *     }
+ * }
+ * </pre>
+ *
+ * It can then be used in a {@link RuleChain} like this:
+ * <pre>
+ * class MyTest {
+ *     &#064;Rule
+ *     public final RuleChain chain = RuleChain
+ *             .outerRule(new TestRuleAdapter(new MyMethodRule(), this))
+ *             .around(new TestRuleAdapter(new MyMethodRule(), this));
+ *
+ *     ...
+ * }
+ * </pre>
+ *
+ * @since 4.12
+ */
+class TestRuleAdapter implements TestRule {
+    private final MethodRule rule;
+    private final Object testObject;
+
+    /**
+     * Creates a new instance wrapping the provided {@link MethodRule}.
+     *
+     * Needs an instance of the test class using this TestRuleAdapter or
+     * the surrounding {@link RuleChain} to be allow the wrapped
+     * {@link MethodRule} to interact with it.
+     *
+     * @param rule the MethodRule to wrap
+     * @param testObject instance of the test class
+     */
+    public TestRuleAdapter(MethodRule rule, Object testObject) {
+        this.rule = rule;
+        this.testObject = testObject;
+    }
+
+    public Statement apply(Statement base, Description description) {
+        try {
+            String methodName = description.getMethodName();
+            Method method = testObject.getClass().getDeclaredMethod(methodName);
+            return rule.apply(base, new FrameworkMethod(method), testObject);
+        } catch (final NoSuchMethodException e) {
+            return new Statement() {
+                @Override
+                public void evaluate() throws Throwable {
+                    throw e;
+                }
+            };
+        }
+    }
+}
--- a/src/test/java/org/junit/rules/RuleChainTest.java
+++ b/src/test/java/org/junit/rules/RuleChainTest.java
@@ -10,20 +10,42 @@ import static org.junit.Assert.fail;
 import static org.junit.experimental.results.PrintableResult.testResult;
 import static org.junit.rules.RuleChain.outerRule;
 
-import java.io.PrintWriter;
-import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.internal.Throwables;
 import org.junit.runner.Description;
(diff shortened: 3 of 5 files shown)


saschpe:
Unfortunately, https://github.com/junit-team/junit4/pull/1340 was closed before I was able to reply or update the pull. Let's give it another try.

I simplified TestRuleAdapter based on your comments and added overrides to RuleChain. I'm not exactly sure if the later works for you, so let's discuss.

src/main/java/org/junit/rules/TestRuleAdapter.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+public class TestRuleAdapter implements TestRule {
+    private final MethodRule rule;
+
+    public TestRuleAdapter(MethodRule rule) {
+        this.rule = rule;
+    }
+
+    public Statement apply(Statement base, Description description) {
+        return rule.apply(base, createFrameworkMethod(description), description.getTestClass());
```
kcooney: The object passed as the last parameter of `MethodRule.apply()` needs to be the test instance, not the test class. The only way to get it is to have the caller pass in the test instance into the `TestRuleAdapter` constructor.

saschpe:
Ok, updated accordingly. I wonder whether it should contain a convenience constructor only taking the methodrule nonetheless. It would simplify test code for people that don't need to interact with the test instance.

src/main/java/org/junit/rules/TestRuleAdapter.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private FrameworkMethod createFrameworkMethod(Description description) {
+        try {
+            String methodName = description.getMethodName();
+            Class<?> c = description.getTestClass();
+            Method m = c.getDeclaredMethod(methodName);
+            return new FrameworkMethod(m);
+        } catch (Exception e) {
+            throw new IllegalStateException(e);
```
kcooney: This seems like an odd exception type to use. Let's use `RuntimeException` and pass a useful message to the `RuntimeException`  constructor.

src/test/java/org/junit/rules/RuleChainTest.java
```diff
@@ -57,6 +78,7 @@ public void example() {
 
     @Test
     public void executeRulesInCorrectOrder() throws Exception {
+        LOG.clear();
```
kcooney: Let's put this call in a `@Before` method named `clearLog()`

src/test/java/org/junit/rules/RuleChainTest.java
```diff
@@ -79,7 +101,7 @@ public void aroundShouldNotAllowNullRules() {
     public static class RuleChainWithNullRules {
         @Rule
         public final RuleChain chain = outerRule(new LoggingRule("outer rule"))
-                .around(null);
+                .around((TestRule) null);
```
kcooney: This cast isn't needed, is it?

src/test/java/org/junit/rules/RuleChainTest.java
```diff
@@ -93,4 +115,26 @@ public void whenRuleChainHasNullRuleTheStacktraceShouldPointToIt() {
         String stacktrace = Throwables.getStacktrace(result.getFailures().get(0).getException());
         assertThat(stacktrace, containsString("\tat org.junit.rules.RuleChainTest$RuleChainWithNullRules.<init>(RuleChainTest.java:"));
     }
+
+    public static class RuleChainWithMethodRules {
+        @Rule
+        public final RuleChain chain = RuleChain
+                .outerRule(new LoggingMethodRule("outer method rule"), RuleChainWithMethodRules.this)
```
kcooney: Please replace `RuleChainWithMethodRules.this` with simply `this` (here and in the example in the Javadoc)

src/test/java/org/junit/rules/TestRuleAdapterTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runners.model.FrameworkMethod;
+import org.junit.runners.model.Statement;
+
+public class TestRuleAdapterTest {
```
kcooney: Thia test needs to be added to the appropriate suite or it won't run as part of the Maven build

src/test/java/org/junit/rules/TestRuleAdapterTest.java
```diff
@@ -0,0 +1,59 @@
...
+                    base.evaluate();
+                };
+            };
+        }
+    };
+
+    public static class UseRuleChain {
+        @Rule
+        public final RuleChain chain = RuleChain
+                .outerRule(new LoggingRule("outer rule"), UseRuleChain.this)
```
kcooney: You can just use `this` here too

src/main/java/org/junit/rules/RuleChain.java
```diff
@@ -105,6 +119,25 @@ public RuleChain around(TestRule enclosedRule) {
...
     /**
+     * Create a new {@code RuleChain}, which encloses the given {@link MethodRule} with
+     * the rules of the current {@code RuleChain}.
+     *
+     * @param enclosedRule the rule to enclose; must not be {@code null}.
+     * @param testObject instance of the current test class.
+     * @return a new {@code RuleChain}.
+     * @throws NullPointerException if the argument {@code enclosedRule} is {@code null}
+     */
+    public RuleChain around(MethodRule enclosedRule, Object testObject) {
```
kcooney: Need `@since` in the Javadoc for the new methods

src/main/java/org/junit/rules/TestRuleAdapter.java
```diff
@@ -0,0 +1,75 @@
...
+ *             .outerRule(new TestRuleAdapter(new MyMethodRule(), MyTest.this))
+ *             .around(new TestRuleAdapter(new MyMethodRule(), MyTest.this));
+ *
+ *     ...
+ * }
+ * </pre>
+ *
+ * @since 4.12
+ */
+public class TestRuleAdapter implements TestRule {
```
kcooney: I am wondering if this should be package scope. What is the use case for code outside of `RuleChain` using this?

kcooney:
To answer your question, I don't think we should create versions of these APIs that do not take in a test instance. Users of `RuleChain` might not know whether a given `MethodRule` needs a test instance (or would need one in a future version)

src/test/java/org/junit/rules/RuleChainTest.java
saschpe: It is because there's now two overrides of around(Something)

src/main/java/org/junit/rules/TestRuleAdapter.java
saschpe: Makes sense for me too with the convenience methods on RuleChain

saschpe:
ping

src/main/java/org/junit/rules/TestRuleAdapter.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public Statement apply(Statement base, Description description) {
+        return rule.apply(base, createFrameworkMethod(description), testObject);
+    }
+
+    private FrameworkMethod createFrameworkMethod(Description description) {
+        try {
+            String methodName = description.getMethodName();
+            Class<?> c = description.getTestClass();
```
kcooney: I'm wondering if we should use `testObject` to get the test class instead of the description. On the one hand, if there are multiple class loaders then the `Class` instance obtained from `Description` might not be the correct class. On the other hand, what if `testObject` isn't the test instance?

src/main/java/org/junit/rules/TestRuleAdapter.java
```diff
@@ -0,0 +1,75 @@
...
+            Method m = c.getDeclaredMethod(methodName);
+            return new FrameworkMethod(m);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
kcooney: If we get here, `apply()` will throw an exception. Instead, I think `apply()` should return a `Statement` that, when executed, rethrows the exception.

Also, if we need to wrap the description in another exception, we should probably provide a message

kcooney:
@saschpe This problem came up in part of our code base, and upon taking a closer look, I found some potential problems. Do you have time to comment on them?

src/main/java/org/junit/rules/TestRuleAdapter.java
saschpe: Yep, this can certainly be done...

kcooney: Let's please have `apply()` return a Statement that, when executed, rethrows the exception.

Also, if we need to wrap the description in another exception, provide a message

src/main/java/org/junit/rules/TestRuleAdapter.java
kcooney: Let's use `testObject` to get the test class instead of the description.

saschpe:
@kcooney Ping, any update?

saschpe:
I'm not allowed to merge :-)

kcooney:
@saschpe I know. Lately we have been waiting for multiple maintainers to approve each pull (or one approves, another merges)

src/main/java/org/junit/rules/RuleChain.java
```diff
@@ -82,6 +84,19 @@ public static RuleChain outerRule(TestRule outerRule) {
...
     }
 
+    /**
+     * Returns a {@code RuleChain} with a single {@link MethodRule}. This method
+     * is the usual starting point of a {@code RuleChain}.
+     *
+     * @param outerRule the outer rule of the {@code RuleChain}.
+     * @param testObject instance of the current test class.
+     * @return a {@code RuleChain} with a single {@link TestRule}.
+     * @since 4.12
```
marcphilipp: 4.13

src/main/java/org/junit/rules/RuleChain.java
```diff
@@ -105,6 +120,26 @@ public RuleChain around(TestRule enclosedRule) {
...
 
     /**
+     * Create a new {@code RuleChain}, which encloses the given {@link MethodRule} with
+     * the rules of the current {@code RuleChain}.
+     *
+     * @param enclosedRule the rule to enclose; must not be {@code null}.
+     * @param testObject instance of the current test class.
+     * @return a new {@code RuleChain}.
+     * @throws NullPointerException if the argument {@code enclosedRule} is {@code null}
+     * @since 4.12
```
marcphilipp: 4.13

src/main/java/org/junit/rules/TestRuleAdapter.java
```diff
@@ -0,0 +1,75 @@
...
+ *     &#064;Rule
+ *     public final RuleChain chain = RuleChain
+ *             .outerRule(new TestRuleAdapter(new MyMethodRule(), this))
+ *             .around(new TestRuleAdapter(new MyMethodRule(), this));
+ *
+ *     ...
+ * }
+ * </pre>
+ *
+ * @since 4.12
```
marcphilipp: 4.13

src/main/java/org/junit/rules/TestRuleAdapter.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public TestRuleAdapter(MethodRule rule, Object testObject) {
+        this.rule = rule;
+        this.testObject = testObject;
+    }
+
+    public Statement apply(Statement base, Description description) {
+        try {
+            String methodName = description.getMethodName();
+            Method method = testObject.getClass().getDeclaredMethod(methodName);
```
marcphilipp: This doesn't work for methods with parameters, e.g. Theories, does it?

kcooney: Marc is correct. Our options appear to be:

1. Update the Javadoc to mention this limitation, but otherwise merge this
2. Add a method to `Description` to get the `Method`, then merge this
3. Do not merge this

For option 2, I would prefer that we not add more static methods for creating `Description` instances. We could explore the work that was done a while back on creating builders for creating immutable `Description` instances (see https://github.com/junit-team/junit4/tree/immutable-description). I recently cleaned up the code in that branch, but it's a rather large set of changes, and the changes could break people.

I'm removing the "4.13" tag, since I don't want to block the next JUnit 4.x release on resolving this issue.

kcooney:
I'm closing this in favor of #1445

PR #1355, opened 2016-08-27
A new suite that allows running the specified classes a specified number of times

lolocohen:
Example usage:

``` java
@RunWith(RepeatingSuite.class)
@RepeatingSuite.RepeatingSuiteClasses(repeat = 10,
        shuffleClasses = false, shuffleMethods = false,
        classes = { MyClass1.class, MyClass2.class, MyClass3.class })
public class MySuite {
}

// with only the classes parameter, the behavior is equivalent to that of Suite
@RunWith(RepeatingSuite.class)
@RepeatingSuite.RepeatingSuiteClasses(
    classes = { MyClass1.class, MyClass2.class, MyClass3.class })
public class MySuite2 {
}
```

Where:
- `repeat` is the number of times to repeat the classes in the suite (defaults to 1)
- `shuffleClasses` specifies that the order of the classes should be randomized at each repetition (defaults to false)
- `shuffleMethods` specifies that the order of the test methods in each class should be randomized at each repetition (defaults to false)
- `classes` is the set of classes to run (same semantics as for the `Suite` class)

### DIFF
--- a/src/main/java/org/junit/runners/RepeatingSuite.java
+++ b/src/main/java/org/junit/runners/RepeatingSuite.java
@@ -1,0 +1,142 @@
+package org.junit.runners;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import org.junit.runner.Runner;
+import org.junit.runners.model.FrameworkMethod;
+import org.junit.runners.model.InitializationError;
+
+/**
+ * A suite that allows running the specified classes a specified number of
+ * times, and optionally allows a random shuffling of the test classes and
+ * the test methods at each repetition.
+ * <p>
+ * Example usage:
+ * 
+ * <pre>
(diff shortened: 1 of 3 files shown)


kcooney:
Thanks for the contribution.

What is the benefit of running tge tests multiple times. If any of the runs fail, the test run fails. Moreover, tests should be deterministic. I could somewhat understand the need if the tests were flaky and the runner would only report a failure if a test method failed multiple times, but that is what is going on here.

If the intent is to find flaky tests, then I suggest  doing periodic runs to root out for flaky tests (run all tests multiple times). With the current solution, you need a special runner (so you can't combine with other runners) and every developer has to wait for all of the reruns every time they run the tests even if they aren't flaky.

PeterWippermann:
The identification of flaky tests is a use case I can well understand and which I also stumbled upon.
In that case you also want to report test cases that only fail - let's say - 1 out of 10 times.

Once you identified your flaky test cases (and were not able to stabilise them) your need for reporting would be quite the contrary: Now you maybe only want to report test cases that fail all the time - just as you said @kcooney .

Another use case would be measuring the mean execution time / memory usage of your tests. That way JUnit is (mis?)used as a performance test tool.

kcooney:
No response so closing this pull. Anyone who wants this behavior can implement it themselves, since it requires no changes to the core JUnit code.

Issue #1238, opened 2016-02-04
MultipleFailureException should override printStackTrace()


drewhannay:
Currently `MultipleFailureException` overrides the `getMessage()` method, but not this `printStackTrace()` method. This means that anyone calling `printStackTrace()` will only see the last exception in the `MultipleFailureException`, which is often the least useful.

Ideally `MultipleFailureException#printStackTrace()` should include the stack traces of all the wrapped exceptions.

jlink:
Since JUnit 5 is based on the exception hierarchy in opentest4j you might also want to bring up your point there: https://github.com/ota4j-team/opentest4j
It's called `MultipleFailuresError` there.

kcooney:
This was fixed via 35c8fd37

PR #1359, opened 2016-09-14
Fix StackTrace printing when multiple exceptions occur (#1238)

jacobrh91:
I fixed the way MultipleFailureException prints the stacktraces. Now, when multiple exceptions are found, the stacktrace for each exception is printed

### DIFF
--- a/src/main/java/org/junit/runners/model/MultipleFailureException.java
+++ b/src/main/java/org/junit/runners/model/MultipleFailureException.java
@@ -1,5 +1,7 @@
 package org.junit.runners.model;
 
+import java.io.PrintStream;
+import java.io.PrintWriter;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


marcphilipp:
Who calls printStackTrace directly and why?

kcooney:
@marcphilipp apparently, we do: https://github.com/junit-team/junit4/search?utf8=%E2%9C%93&q=printStackTrace

The only call I see that could take in a `MultipleFailureException` is in `Throwables

marcphilipp:
Thanks for the pointer!

@kcooney Do you foresee this causing any issues?

kcooney:
@marcphilipp if users store these logs, then the log files could obviously get longer. The output might be hard to read, but if there are multiple failures then we should print *something* about each of them.

We might want to revisit trimming the stack traces in the future. Each of these traces would likely contain duplicate frames that include the guts of JUnit

marcphilipp:
I am in favor of merging this change.

@junit-team/junit-committers What do you guys think?

kcooney:
@jacobrh91 could you please update the release notes?

Issue #1363, opened 2016-09-21
ErrorCollector reports AssumptionViolatedException as failure when assumption is not meet multiple times.


rpost:
Sample test:

``` java
import org.junit.Assume;
import org.junit.Rule;
import org.junit.Test;
import org.junit.rules.ErrorCollector;

public class SampleTest {

    @Rule
    public ErrorCollector errorCollector = new ErrorCollector();

    @Test
    public void test() throws Exception {
        errorCollector.checkSucceeds(() -> {
            doSth();
            return null;
        });
        errorCollector.checkSucceeds(() -> {
            doSth();
            return null;
        });
    }

    private void doSth() {
        Assume.assumeTrue(false);
    }

}
```

Expected result: test ignored.
Actual: test failed.

Working as expected (ie. test results as ignored) when:

``` java
    @Test
    public void test() throws Exception {
        errorCollector.checkSucceeds(() -> {
            doSth();
            return null;
        });
    }
```

kcooney:
I am tempted to argue that error collector shouldn't be used for assumption failures. ErrorCollector is intended to allow you to make an assertion but have the test continue. Assumption failures are meant to stop processing and move on to the next test.

Perhaps if the delegate passed to `checkSucceeds()` throws an assumption failure  `checkSucceeds()` should immediately rethrow the exception

rpost:
> Perhaps if the delegate passed to checkSucceeds() throws an assumption failure checkSucceeds() should immediately rethrow the exception

That's what I'd expect

kcooney:
@rpost Could you send us a pull with that change and a test? The JavaDoc should document this of course

rpost:
I took a look into code and now I think we should make all `ErrorCollector` AssumptionViolationException-proof, which implies changes in `org.junit.Assert#expectThrows` so that assumption violation is rethrown if expected exception is not `AssumptionViolationException`. Do you agree?

kcooney:
I don't think `expectThrows()` should special case `AssumptionViolationException`. The contract of `expectThrows()` is very explicit; if the expected exception isn't thrown by the lambda you get an assertion error.

Taking a step back, what problem are you trying to solve? The expressions passed into these methods should be production code, not test code, and `AssumptionViolationException` should generally be thrown by test code.

rpost:
> Taking a step back, what problem are you trying to solve?

[...]

PR #1367, opened 2016-10-04
Not met assumptions fails fast when used with ErrorCollector #1363

rpost:
PR for #1363

### DIFF
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -4,6 +4,7 @@ import org.hamcrest.Matcher;
 import org.hamcrest.MatcherAssert;
 import org.junit.function.ThrowingRunnable;
 import org.junit.internal.ArrayComparisonFailure;
+import org.junit.internal.AssumptionNotSupportedException;
 import org.junit.internal.ExactComparisonCriteria;
 import org.junit.internal.InexactComparisonCriteria;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/junit/Assert.java
```diff
@@ -991,6 +996,8 @@ public static void assertThrows(Class<? extends Throwable> expectedThrowable, Th
     public static <T extends Throwable> T expectThrows(Class<T> expectedThrowable, ThrowingRunnable runnable) {
         try {
             runnable.run();
+        } catch (AssumptionViolatedException e) {
+            throw new AssumptionNotSupportedException("Assumptions are not supported in expectThrows/assertThrows", e);
```
kcooney: This would break tests that are explicitly asserting that a given lambda throws `AssumptionViolatedException`. This is a reasonable thing to test if you have a common testing library that you want to test:

    expectThrows(AssumptionViolatedException.class, () -> tester.verify());

I don't think these methods should special-case `AssumptionViolatedException`. The contracts for these methods are clear and simple, and the behavior you are adding would be unexpected to many users (even if documented). It also would make these methods differ from the JUnit5 versions.

Instead, I think `MultipleFailureException.assertEmpty()` should special case lists that contain one element of type `AssumptionViolatedException`.

kcooney:
Thanks for the pull. I'm not convinced this is the right direction to take (see my comment)

src/main/java/org/junit/Assert.java
kcooney: also, I don't see how this helps, since befoer your change if `AssumptionViolatedException` was not expected and was thrown, it would be wrapped with an `AssertionError`

kcooney:
Proposing #1395

Issue #1320, opened 2016-06-02
Request.classes(newTestClass).getRunner().getDescription()


Reissner:
With a quite straightforward testclass annotated 
@RunWith(Suite.class)
@SuiteClasses({
    XMSectionTest.Basic.class
})
public class XMSectionTest {
...

I obtain for Request.classes(testClass).getRunner().getDescription() the value null, 
whereas for Request.aClass(testClass).getRunner().getDescription() 
I obtain eu.simuline.arithmetics.left2right.XMSectionTest 
Hm,.... the first seems a bug to me, 
whereas the second one.. seems to me at least uncomfortable, 
because I think the description shall reflect the structure of the request and not its origin.

marcphilipp:
For the sake of a complete example:

```java
@RunWith(Suite.class)
@SuiteClasses({
        XMSectionTest.Basic.class
})
public class XMSectionTest {
    public static class Basic {
        @Test
        public void test() {
            Assert.fail("foo");
        }
    }

    public static void main(String... args) {
        Class<?> testClass = XMSectionTest.class;

        System.out.println("Request.classes(testClass):");
        Runner a = Request.classes(testClass).getRunner();
        printTree(a.getDescription());
        System.out.println();

        System.out.println("Request.aClass(testClass):");
        printTree(Request.aClass(testClass).getRunner().getDescription());
        System.out.println();
    }

    private static void printTree(Description description) {
        printTree(description, "- ");
    }
    private static void printTree(Description description, String indentation) {
        System.out.print(indentation);
        System.out.println(description + " (" + System.identityHashCode(description) + ")");
        for (Description child : description.getChildren()) {
            printTree(child, indentation + "  ");
        }
    }
}
```

Executing the main method prints:

```
Request.classes(testClass):
- null (1325547227)
-   org.junit.tests.XMSectionTest (980546781)
-     org.junit.tests.XMSectionTest$Basic (2061475679)
-       test(org.junit.tests.XMSectionTest$Basic) (140435067)

Request.aClass(testClass):
- org.junit.tests.XMSectionTest (1450495309)
-   org.junit.tests.XMSectionTest$Basic (1670782018)
-     test(org.junit.tests.XMSectionTest$Basic) (1706377736)
```

So, there is an intended difference: When using `Request.classes` there's an additional `Description` at the root of the hierarchy. Its display name is `"null"` because it doesn't have a test class associated with it. So the descriptor is not `null` but its `toString()` method returns the String `"null"`.

[...]

PR #1377, opened 2016-10-25
Test for #1320 - Description produced by Request.classes() shouldn't be null

PeterWippermann:


src/test/java/org/junit/runner/RequestTest.java
```diff
@@ -0,0 +1,21 @@
+package org.junit.runner;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import org.junit.Test;
+
+public class RequestTest {
```
kcooney: You need to add RequestTest.class to src/test/java/org/junit/runner/AllRunnerTests.java in order for the Maven build to run your test.

Please verify the test fails when the changes you made to Computer are removed

PeterWippermann: done :-)

kcooney:
Thanks! Could you please update the release notes?

PeterWippermann:
Added release notes.

Issue #1350, opened 2016-07-24
Code base doesn't comply with JUnit's own coding style


PeterWippermann:
I'd like to suggest to reformat all code under `org.junit.*` once to meet the defined [formatting guidelines](https://github.com/junit-team/junit4/blob/master/CODING_STYLE.txt).

At the moment not all the files are formatted accordingly to that standard. This causes problems for contributors - especially in pull request. If you format the code you get a lot of extraneous diffs - just because of the reformatting - in addition to your actual changes.

I think it's not too late to do this formatting.
Regarding existing PRs that tried to preserve the outdated styling: They would get extraneous diffs than of course, but if they apply the new format as well everything would be fine again. In addition, I think many of PRs are dead anyway, since many of them haven't seen any updates for months or even years.

As a result, contributing to JUnit 4 and reviewing the changes by the core team should become easier in then.

kcooney:
If you can find a way to guarantee that new code will use the new formatting regardless of what editor they use (for instance, with [Spotless](https://github.com/diffplug/spotless/blob/master/README.md)) then I'd be fine doing a reformat.

PeterWippermann:
All I can do is: import Google's code style in Eclipse. Change 2 space indentantion to 4 spaces. And run the formatting over all affected files.

Stephan202:
A non-Eclipse solution could be to use [google-java-format](https://github.com/google/google-java-format/). There's also a [googleformatter-maven-plugin](https://github.com/talios/googleformatter-maven-plugin).

kcooney:
Thanks, @Stephan202 

I'm going on vacation for a few days, but hopefully I can look into the Maven plugin next weekend.

tomwhoiscontrary:
JUnit uses the Google style with the exception of different indentation.

The google-java-format tool doesn't support overriding the indentation.

If you want to enforce the style after reformatting, the tool of choice would be [Checkstyle](http://checkstyle.sourceforge.net/), via its [Maven plugin](https://maven.apache.org/plugins-archives/maven-checkstyle-plugin-2.16/). This supports [using the Google style as a built-in custom configuration](https://maven.apache.org/plugins-archives/maven-checkstyle-plugin-2.16/examples/custom-checker-config.html). Again, there's no way to override the indentation. You could use a modified copy of the [configuration file](https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml), but that's under LGPL, so you immediately run into a license nightmare.

[...]

PR #1379, opened 2016-10-30
Make some effort to comply with style guide

tomwhoiscontrary:
This relates to #1350, "Code base doesn't comply with JUnit's own coding style". It adds the Google Formatter to reformat the code, and Checkstyle to enforce compliance. It doesn't actually include a mass reformatting of code.

I'm not sure this is something you'd want to merge, but it might help understand the situation better.

Some observations:
- JUnit differs from Google in that it uses four spaces from indentation. The Google Formatter can't do that. However, we can roughly restore four-space indentation with a fairly simple sed command (as given in the POM!). Also, we can suppress indentation-related violations in Checkstyle, so we won't fail the build because of our formatting.
- By default, Checkstyle will only fail the build for 'errors', of which the codebase has none. However, the codebase has 1289 'warnings' (ignoring indentation), which still seem serious enough to want to fix.
- Google Formatter actually doesn't fix all the problems Checkstyle finds. For example, it doesn't reorder imports. I get the impression that there is reformatting that Google Formatter can do on the command line that it can't do through its API, which is what the Maven plugin uses. After running the formatter, the codebase is down to 1113 warnings. However, after restoring indentation, it's back up to 1235 - even with indentation warnings ignored!

So, whilst you might well want Checkstyle as part of the build, i'm skeptical that the Google Formatter is actually useful.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -396,6 +396,15 @@
                     <violationIgnore>indentation</violationIgnore>
                 </configuration>
             </plugin>
+            <plugin>
+                <!--
+                To apply: mvn googleformatter:format
+                To repair indentation (BSD/Mac sed): git status -s | sed -En 's/^ M (src.*\.java)$/\1/p' | xargs sed -Ei '' 's/^( *)/\1\1/'
+                -->
+                <groupId>com.theoryinpractise</groupId>
+                <artifactId>googleformatter-maven-plugin</artifactId>
+                <version>1.0.6</version>
+            </plugin>
         </plugins>
     </build>
 
--- a/pom.xml
+++ b/pom.xml
@@ -107,6 +107,7 @@
         <project.build.sourceEncoding>ISO-8859-1</project.build.sourceEncoding>
         <arguments />
         <gpg.keyname>67893CC4</gpg.keyname>
+        <checkstyle.config.location>google_checks.xml</checkstyle.config.location>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -379,6 +380,31 @@
                 <artifactId>maven-resources-plugin</artifactId>
                 <version>2.7</version>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+                <version>2.17</version>
```
PeterWippermann: One of the builds on Travis CI for jdk6 is failing. As far as I can see, this version of Checkstyle requires Java 7. If you could go on a lower version of Checkstyle, that might fix this build.
The actual problem is, if it should be possible to build Junit 4 on JDK 6? And if we don't build on JDK 6, does this imply that JUnit can no longer be used on JRE 6? And is this really still necessary, @kcooney and @marcphilipp ?

marcphilipp: Yes, it should be possible to build it on JDK6. But to make sure we don't use any JDK > 5 APIs we should build using JDK 5.

Why shouldn't it be possible to use JUnit on JDK 6 if it wasn't build on JDK 6?

tomwhoiscontrary: If it was possible to disable Checkstyle in the JDK 6 build, would that be okay? It's not as if the source code is any different between the JDK 6 and JDK 7/8 builds. Using a JDK 6-compatible version of something in 2016 seems a bit ludicrous.

In fact, i would argue that what you really want to do is build all the classes under whatever JDK you are going to use to build the releases, and then run the tests on the various target JDKs, to more accurately simulate what happens to users. I have no idea how to get Maven and Travis to do that, though!

pom.xml
```diff
@@ -379,6 +380,31 @@
...
+                </configuration>
+            </plugin>
+            <plugin>
+                <!--
+                To apply: mvn googleformatter:format
+                To repair indentation (BSD/Mac sed): git status -s | sed -En 's/^ M (src.*\.java)$/\1/p' | xargs sed -Ei '' 's/^( *)/\1\1/'
+                -->
+                <groupId>com.theoryinpractise</groupId>
+                <artifactId>googleformatter-maven-plugin</artifactId>
+                <version>1.0.6</version>
```
Stephan202: The plugin [allows](https://github.com/talios/googleformatter-maven-plugin/blob/master/src/main/java/com/theoryinpractise/googleformatter/GoogleFormatterMojo.java#L74) one to configure [AOSP](https://github.com/google/google-java-format/blob/master/core/src/main/java/com/google/googlejavaformat/java/JavaFormatterOptions.java#L40) style, which uses four spaces.

PR #1380, opened 2016-10-31
Throw AssumptionViolatedException when single null is passed to Assum…

xendo:
…e.assumeNotNull(Object... objects) method

Previous implementation threw NullPointerException when single null argument was passed:
at java.util.Objects.requireNonNull(Objects.java:203)
at java.util.Arrays$ArrayList.<init>(Arrays.java:3813)
at java.util.Arrays.asList(Arrays.java:3800)

### DIFF
--- a/src/main/java/org/junit/Assume.java
+++ b/src/main/java/org/junit/Assume.java
@@ -79,6 +79,7 @@ public class Assume {
      * If called with one or more null elements in <code>objects</code>, the test will halt and be ignored.
      */
     public static void assumeNotNull(Object... objects) {
+        assumeThat(objects, notNullValue());
         assumeThat(asList(objects), everyItem(notNullValue()));
     }
 
--- a/src/main/java/org/junit/Assume.java
+++ b/src/main/java/org/junit/Assume.java
@@ -76,7 +76,8 @@ public class Assume {
     }
 
     /**
-     * If called with one or more null elements in <code>objects</code>, the test will halt and be ignored.
+     * If called with a {@code null} array or one or more {@code null} elements in {@code objects},
+     * the test will halt and be ignored.
      */
     public static void assumeNotNull(Object... objects) {
         assumeThat(objects, notNullValue());
--- a/src/test/java/org/junit/tests/experimental/AssumptionTest.java
+++ b/src/test/java/org/junit/tests/experimental/AssumptionTest.java
@@ -111,6 +111,16 @@ public class AssumptionTest {
     }
 
     @Test
+    public void assumeNotNullSingleNullThrowsException() {
+        try {
+            assumeNotNull(null);
+            fail("should throw AssumptionViolatedException");
+        } catch (AssumptionViolatedException e) {
+            // expected
+        }
+    }
+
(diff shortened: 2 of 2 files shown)


kcooney:
BTW, are you absolutely sure the test would fail without the code changes? I would think you would need to pass in a `null` array to trigger the new code path

xendo:
Kevin,

Out of those two:

`    @Test
    public void arraysNull() throws Exception
    {
        Arrays.asList(null);
    }

    @Test
    public void arraysArrayNull() throws Exception
    {
        Object[] objects = {null};
        Arrays.asList(objects);
    }`

only the first one fails with:

> java.lang.NullPointerException
	at java.util.Objects.requireNonNull(Objects.java:203)
	at java.util.Arrays$ArrayList.<init>(Arrays.java:3813)
	at java.util.Arrays.asList(Arrays.java:3800)
	at com.amazon.personid.service.test.integration.ArraysTest.arraysNull
>

xendo:
Also it's fair to say that this invocation produces warning:

> Warning:(14, 23) java: non-varargs call of varargs method with inexact argument type for last parameter;
  cast to java.lang.Object for a varargs call
  cast to java.lang.Object[] for a non-varargs call and to suppress this warning

More details here: http://stackoverflow.com/a/4028086

kcooney:
To extrapolate, the following is the only way to get a NPE out of this method:

    assumeNotNull(null);

and

    Object[] values = null;
    assumeNotNull(values);

The first isn't clearly a valid use of this API. The second should be exceedingly rare.

xendo:
Kevin,

You are correct. This is much more subtle than I originally thought. If I was using Java I would have never caught this. Reproducing this in Groovy is, however, much simpler:

```
    def "assumeNotNull should throw AssumptionViolatedException when passed single null value" () {
        given:
        String string = null
        when:
        assumeNotNull(string)
        then:
        thrown AssumptionViolatedException
    }
```

FYI in Java invocation on any type of Array will throw e.g.

```
String[] values = null;
assumeNotNull(values);
```

src/test/java/org/junit/tests/experimental/AssumptionTest.java
```diff
@@ -111,6 +111,16 @@ public void assumeNotNullThrowsException() {
     }
 
     @Test
+    public void assumeNotNullSingleNullThrowsException() {
+        try {
+            assumeNotNull(null);
```
marcphilipp: Please add a cast to `Object[]` here.

src/main/java/org/junit/Assume.java
```diff
@@ -79,6 +79,7 @@ public static void assumeFalse(String message, boolean b) {
      * If called with one or more null elements in <code>objects</code>, the test will halt and be ignored.
```
marcphilipp: Please update the Javadoc:

    If called with a {@code null} array or one or more {@code null} elements in {@code objects},
    the test will halt and be ignored.

stefanbirkner:
Thanks! Could you please update the release notes?

Issue #45, opened 2009-11-16
Add Before and After Parameterized Run annotations


bigmikef:
Currently there is no simple way for a method to be invoked before or after a Parameterized instance is ran.

Said another way, along with @BeforeClass and @AfterClass there should be @BeforeParameterRun @AfterParameterRun.  Each fires before/after a Parameterized instance is ran.

bileshg:
@coreyjv Is this functionality in the pipeline because we are running Parameterized Test Cases where we need such annotations. Both @BeforeParam and @AfterParam are required.

coreyjv:
@bileshg --

Per @dsaff this request was 'mothballed' a few years ago now. I defer to him if he wants to resurface it again. Additionally, I'm not sure if this is something that would be more appropriate for JUnit 5 now (I haven't kept up with the progress there).

bileshg:
@coreyjv Ok. Thanks for responding. Actually, I needed it for something else - Getting failed test case count as per parameter. Take a look at this [question](http://stackoverflow.com/questions/37697014/get-failed-test-case-count-as-per-parameter).

dstaraster:
I see that "@BeforeParam/@AfterParam for Parameterized runner #1435" got merged and added to the 4.13 release notes. Does that mean there's going to be a 4.13 release? When? I'm in dire need of this functionality! I didn't see it in JUnit 5, so I'm wondering when I can expect @BeforeParam/@AfteParam to be available.

kcooney:
@dstaraster currently most of the JUnit team is focusing on releasing JUnit 5.0. I hope that soon after that we can start on the 4.13 release.

raubel:
@kcooney I'm also very interested in these @BeforeParam and @AfterParam annotations (and unfortunately not yet ready to migrate my thousands of tests to JUnit5).
Is the 4.13 release still expected?

kcooney:
@raubel I opened #1496 to discuss a 4.13 release.

JUnit5 can run JUnit4-style tests, so you wouldn't need to migrate all of your tests to use JUnit5 features.

PR #1381, opened 2016-11-01
Parameterized: added new @StaticParameter annotation

meden:
The new annotation allows to inject parameterized values in static test
members.

This is useful when one has to run a parameterized test case against
parametric expensive resources.

### DIFF
--- a/src/main/java/org/junit/runners/Parameterized.java
+++ b/src/main/java/org/junit/runners/Parameterized.java
@@ -218,6 +218,27 @@ public class Parameterized extends Suite {
     }
 
     /**
+     * Annotation for fields of the test class which will be initialized by the
+     * method annotated by <code>StaticParameter</code>.
+     * It can be used to initialize static fields.
+     * Index range must start at 0.
+     * Default value is 0.
+     */
+    @Retention(RetentionPolicy.RUNTIME)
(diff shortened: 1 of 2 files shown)


marcphilipp:
Can you please provide an example where this is useful?

meden:
I made this to do an integration test of a mail processor. The processors uses a MessageBatch abstraction which insists on different mail stores (i.e. protocols). As you can imagine, the interface is the same, but the behavior is not.
I set up only one parametric test case, where different MessageBatch implementations get tested. Each of them, though, requires to connect to a different store, which is a slow task to do at each test method.

An alternative (which I implemented) without static parameter injection is to use a `@ClassRule` whith static fixtures, from which pull needed mail folder and such. IMHO, this solution is less straightforward than a static injection.

BTW, I can understand this is a bit a corner case.

kcooney:
I am not sure I completely understand, but it sounds like your use case is a bit of a stretch for Parameterized. You aren't testing one thing with multiple values, you are testing multiple implementations. 

Can you provide an example of what your test looks like with static injection and what it looks like with `@ClassRule`?

stefanbirkner:
I think that something like `@BeforeParameterRun` in #45 would be a solution for your problem, too. Am I right?

PR #1382, opened 2016-11-07
Remove unnecessary modifiers "public" for TestRunListener interface

AxelMonroyX:
Remove unnecessary modifiers "public" for TestRunListener interface because all abstract, default, and static methods in an interface are implicitly public, so you can omit the public modifier.

Here is specified: [Java Doc About Interface Definition](http://docs.oracle.com/javase/tutorial/java/IandI/interfaceDef.html)


stefanbirkner:
Thanks, @AxelMonroyX.

PR #1384, opened 2016-11-16
Refactored some access modifiers, mainly the redundant ones on interfaces

Mathreyu:
Noted some warning on the IDE, mainly because of access modifiers being narrowed down to smaller scopes. in the TestListener interface at the framework module there was a redundancy on making methods public.

### DIFF
--- a/src/main/java/junit/extensions/TestDecorator.java
+++ b/src/main/java/junit/extensions/TestDecorator.java
@@ -11,7 +11,7 @@ import junit.framework.TestResult;
  */
 @SuppressWarnings("deprecation")
 public class TestDecorator extends Assert implements Test {
-    protected Test fTest;
+    private Test fTest;
 
     public TestDecorator(Test test) {
         fTest = test;
--- a/src/main/java/junit/extensions/TestDecorator.java
+++ b/src/main/java/junit/extensions/TestDecorator.java
@@ -11,7 +11,7 @@ import junit.framework.TestResult;
  */
 @SuppressWarnings("deprecation")
 public class TestDecorator extends Assert implements Test {
-    private Test fTest;
+    protected Test fTest;
 
     public TestDecorator(Test test) {
         fTest = test;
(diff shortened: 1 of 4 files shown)


src/main/java/junit/extensions/TestDecorator.java
```diff
@@ -11,7 +11,7 @@
  */
 @SuppressWarnings("deprecation")
 public class TestDecorator extends Assert implements Test {
-    protected Test fTest;
+    private Test fTest;
```
kcooney: This change could break users' code (for example, subclasses that reference this field)

src/main/java/junit/framework/JUnit4TestAdapterCache.java
```diff
@@ -29,7 +29,7 @@ public Test asTest(Description description) {
         }
     }
 
-    Test createTest(Description description) {
+    private Test createTest(Description description) {
```
kcooney: This change *might* break people. I think it's worth the risk

src/main/java/junit/framework/TestFailure.java
```diff
@@ -10,8 +10,8 @@
  * @see TestResult
  */
 public class TestFailure {
-    protected Test fFailedTest;
-    protected Throwable fThrownException;
+    private Test fFailedTest;
```
kcooney: This change could break users' code (for example, subclasses that reference this field)

src/main/java/junit/framework/JUnit4TestAdapterCache.java
Mathreyu: So, should I leave it this way? or should I roll it back?

src/main/java/junit/framework/TestFailure.java
Mathreyu: Got it, will work in another issue then from the JUnit4 tree :) thanks

src/main/java/junit/framework/JUnit4TestAdapterCache.java
kcooney: Sorry, typo. I think it is **not** worth the risk. We should  not change the modifiers of this field.

I don't think we should ever reduce the visibility of a field or method in a public class (unless we have not released a version of JUnit where that member existed)

PR #1385, opened 2016-11-16
Remove unnecessary boxing on ListTest

AxelMonroyX:
Remove unnecessary boxing on ListTest because Java creates an Integer object from `i` and adds to the object `fFull`

Reference:
[`Autoboxing and Unboxing`](https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html)


kcooney:
Thanks!

PR #1386, opened 2016-11-17
Outboxing issues corrected

Mathreyu:
Some outboxing warnings were corrected as signaled in the IDE.

### DIFF
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -198,7 +198,7 @@ public class Assert {
      */
     public static void assertNotEquals(String message, long unexpected, long actual) {
         if (unexpected == actual) {
-            failEquals(message, Long.valueOf(actual));
+            failEquals(message, actual);
         }
     }
 
@@ -231,7 +231,7 @@ public class Assert {
     public static void assertNotEquals(String message, double unexpected,
             double actual, double delta) {
         if (!doubleIsDifferent(unexpected, actual, delta)) {
-            failEquals(message, Double.valueOf(actual));
+            failEquals(message, actual);
         }
     }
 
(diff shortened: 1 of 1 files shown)


kcooney:
Thanks for the contribution.

I actually prefer the explicit call to the `valueOf()` methods in these cases, so i vote to not merge this.

Mathreyu:
@kcooney I found a previous PR on which another outboxing warning was corrected https://github.com/junit-team/junit4/pull/1385

Is there a reason for this to be correct, but mine is not?

Thanks :+1:

kcooney:
There are a few differences:

1. The previous code there did an explicit "new" on a primitive type, which
is usually not the right thing to do (Integer.valueOf() has a cache). The
code you are changing uses `valueOf()` which is what the compiler would do
when autoboxing a primative type.

2. ListTest is sample code, so having it do something that was clearly not
canonical Java is a bit embarrassing

3. The Assert class has some overloaded methods, so changes like these
might have unintended consequences.

Finally (and this is just my personal opinion) I don't like changes that
only update code from one code style to another. Generally the code in
JUnit was approved by at least one person, so two people agreed that the
style was reasonable.

if you want to make a larger impact, we have many issues labeled
"lowhanging" that you could work on. See
https://github.com/junit-team/junit4/issues?q=is%3Aissue+is%3Aopen+label%3Alowhanging

Mathreyu:
Thanks for the clarification, I will try to contribute on the issues you recommended :)

marcphilipp:
I'm with @kcooney on this one.

kcooney:
FYI, I updated https://github.com/junit-team/junit4/blob/master/CONTRIBUTING.md to document this policy

PR #1390, opened 2016-11-21
Add a StringBuilder to warning messages on TestSuite

AxelMonroyX:
##### Description of change
- Create a StringBuilder in the createTest method because we are not reusing the Strings in the warning messages, it is much faster and consumes less memory

##### Checklist
- [x] Run `mvn verify`
- [x] All tests are passing

##### References:
[The StringBuilder Class](http://docs.oracle.com/javase/tutorial/java/data/buffers.html)

### DIFF
--- a/src/main/java/junit/framework/TestSuite.java
+++ b/src/main/java/junit/framework/TestSuite.java
@@ -47,11 +47,14 @@ public class TestSuite implements Test {
      * mountains, our intrepid adventurers type...
      */
     static public Test createTest(Class<?> theClass, String name) {
+        StringBuilder warningMessage = new StringBuilder();
         Constructor<?> constructor;
         try {
             constructor = getTestConstructor(theClass);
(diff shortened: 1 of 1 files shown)


kcooney:
Thanks for the changes.

While I'm fine with improvements to the sample code (since they are designed for humans to read), we are not accepting other changes to the code under `src/main/java/junit', `test/java/junit/tests/framework` or `test/java/junit/tests/extensions`

(incidentally, the compiler converts string concatenation to calls to `StringBuilder`, so most of the changes in your pull have no performance impact)

Please see https://github.com/junit-team/junit4/issues?q=is%3Aissue+is%3Aopen+label%3Alowhanging to find other places where you can have an impact.

kcooney:
FYI, I updated https://github.com/junit-team/junit4/blob/master/CONTRIBUTING.md to document this policy

PR #1392, opened 2016-11-22
Change javadoc reference of ThrowingRunnable on ExpectedException

AxelMonroyX:
##### Description of change
- Change reference of javadoc from `org.junit.Assert.ThrowingRunnable` to `org.junit.function.ThrowingRunnable` on `ExpectedException`

##### Checklist
- [x] Run `mvn verify`
- [x] All tests are passing
- [x] Generated JavaDoc with working links

### DIFF
--- a/src/main/java/org/junit/rules/ExpectedException.java
+++ b/src/main/java/org/junit/rules/ExpectedException.java
@@ -15,12 +15,12 @@ import org.junit.runners.model.Statement;
 /**
  * The {@code ExpectedException} rule allows you to verify that your code
  * throws a specific exception. Note that, starting with Java 8,
- * {@link org.junit.Assert#assertThrows(java.lang.Class, org.junit.Assert.ThrowingRunnable)
+ * {@link org.junit.Assert#assertThrows(java.lang.Class, org.junit.function.ThrowingRunnable)
  * Assert.assertThrows}
  * is often a better choice since it allows you to express exactly where you
  * expect the exception to be thrown. Use
  * {@link org.junit.Assert#expectThrows(java.lang.Class,
- * org.junit.Assert.ThrowingRunnable) expectThrows}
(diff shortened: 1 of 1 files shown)


marcphilipp:
I think it would actually be more readable if we removed the parameter list from the `@link` references.

AxelMonroyX:
@marcphilipp should I add the qualifier `org.junit.function.ThrowingRunnable` ? 
 ```java
import org.junit.function.ThrowingRunnable;
```
To keep it like this?
```java
/*
 * {@link org.junit.Assert#assertThrows(Class, ThrowingRunnable)
*/
```

kcooney:
I think the parameters are required for @link to produce a working link. Either way, please review the generated JavaDoc

AxelMonroyX:
@kcooney 

##### Checklist
- [x] Generated JavaDoc with working links

Before
```html
<div class="block">The <code>ExpectedException</code> rule allows you to verify that your code
 throws a specific exception. Note that, starting with Java 8,
 <code>Assert.assertThrows</code>
 is often a better choice since it allows you to express exactly where you
 expect the exception to be thrown. Use
 <code>expectThrows</code>
 if you need to assert something about the thrown exception.

```
After
```html
<div class="block">The <code>ExpectedException</code> rule allows you to verify that your code
 throws a specific exception. Note that, starting with Java 8,
 <a href="../../../org/junit/Assert.html#assertThrows-java.lang.Class-org.junit.function.ThrowingRunnable-"><code>Assert.assertThrows</code></a>
 is often a better choice since it allows you to express exactly where you
 expect the exception to be thrown. Use
 <a href="../../../org/junit/Assert.html#expectThrows-java.lang.Class-org.junit.function.ThrowingRunnable-"><code>expectThrows</code></a>
 if you need to assert something about the thrown exception.
```

kcooney:
Thanks!

PR #1393, opened 2016-11-22
Simplify if statements and change StringBuilder to append as chain

AxelMonroyX:
##### Description of change
Simplify if statements on TemporaryFolder
Use StringBuilder properly on InvalidTestClassError

##### Checklist
- [x] Run `mvn verify`
- [x] All tests are passing

### DIFF
--- a/src/main/java/org/junit/rules/TemporaryFolder.java
+++ b/src/main/java/org/junit/rules/TemporaryFolder.java
@@ -273,19 +273,13 @@ public class TemporaryFolder extends ExternalResource {
      *         {@code false} otherwise.
      */
     protected boolean tryDelete() {
-        if (folder == null) {
-            return true;
-        }
-        
-        return recursiveDelete(folder);
+        return folder == null ? true : recursiveDelete(folder);
     }
 
     private boolean recursiveDelete(File file) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/junit/rules/TemporaryFolder.java
```diff
@@ -273,19 +273,13 @@ public void delete() {
      *         {@code false} otherwise.
      */
     protected boolean tryDelete() {
-        if (folder == null) {
-            return true;
-        }
-        
-        return recursiveDelete(folder);
+        return folder == null ? true : recursiveDelete(folder);
```
kcooney: I prefer the was the code was

src/main/java/org/junit/rules/TemporaryFolder.java
```diff
@@ -273,19 +273,13 @@ public void delete() {
...
     }
 
     private boolean recursiveDelete(File file) {
         // Try deleting file before assuming file is a directory
         // to prevent following symbolic links.
-        if (file.delete()) {
-            return true;
-        }
+        if (file.delete()) return true;
```
kcooney: Our current style is to always use braces with `if` statements.

PR #1394, opened 2016-11-27
Use canonical class name for failure message in expectThrows()

kcooney:


kcooney:
Note there are two commits, and I want to keep them separate. I'll be happy to merge once someone approves.

marcphilipp:
There's the edge case of classes that don't have a canonical name. Can we use `getName()` as a fallback for such cases?

kcooney:
@marcphilipp done. PTAL

kcooney:
@marcphilipp done. https://github.com/junit-team/junit5/issues/575

PR #1397, opened 2016-11-30
Change generics on ExpectedException.expectCause().

### LINKED ISSUE #1073: Backwards wildcard bound on ExpectedException.expectCause()
The signature of that method is `expectCause(Matcher<? extends Throwable> expectedCause)`.  This accepts `Matcher<RuntimeException>`, but possibly passes it other types such as `IOException`, resulting in heap pollution.

On the other hand, it forbids `Matcher<Object>`, even though that would be completely type-safe.

The type should be `Matcher<? super Throwable>` instead.

kcooney:
The previous generics (Matcher<? extends Throwable>) would not allow
you do use matchers on Object, like notNullValue().

Fixes #1073


stefanbirkner:
Thanks. Could you add this to the release notes, please.

kcooney:
I updated the release notes.

PR #1398, opened 2016-12-01
Putting a For Loop to reduce lines of code

jesusmtzarvizu:
3 lines of code were changed for a For loop

### DIFF
--- a/src/test/java/org/junit/samples/ListTest.java
+++ b/src/test/java/org/junit/samples/ListTest.java
@@ -35,9 +35,9 @@ public class ListTest {
     public void setUp() {
         fEmpty = new ArrayList<Integer>();
         fFull = new ArrayList<Integer>();
-        fFull.add(1);
-        fFull.add(2);
-        fFull.add(3);
+        for (int i = 1; i < 4; i++) {
+            fFull.add(i);
+        }
     }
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/junit/samples/ListTest.java
```diff
@@ -35,9 +35,9 @@ public static void setUpOnce() {
     public void setUp() {
         fEmpty = new ArrayList<Integer>();
         fFull = new ArrayList<Integer>();
-        fFull.add(1);
-        fFull.add(2);
-        fFull.add(3);
+        for (int i = 1; i < 4; i++) {
```
kcooney: I personally prefer to avoid loops in tests. Tests, especially unit tests, should be simple and obviously correct.

In this particular case, the `for` loop doesn't follow the standard idiom for looping `n` times. It does so for a good reason, but that makes the code slightly more confusing.

jesusmtzarvizu: You are right, I get it , the code is more confusing that way actually

stefanbirkner:
We are not happy with both changes of the pull request and nothing happened on this pull request for more than a month. Therefore I would like to close this pull request. Any objections?

PR #1400, opened 2016-12-11
Runner to detect default test methods on interfaces (mixin support)

ollierob:
I'd like to be able to mixin tests defined as default methods on interfaces. 

This PR adds a new Mixin runner that supports default methods on Java 8+, and operates the same as BlockJUnit4ClassRunner on on Java 7-.

My feeling is that this behaviour should be moved into BlockJUnit4ClassRunner at some point, but for now it's kept as a separate experimental runner.

### DIFF
--- a/src/main/java/org/junit/experimental/runners/Mixin.java
+++ b/src/main/java/org/junit/experimental/runners/Mixin.java
@@ -2,7 +2,6 @@ package org.junit.experimental.runners;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


marcphilipp:
I don't think we should add support for interface default methods to JUnit4. The JUnit Jupiter engine already supports it, see http://junit.org/junit5/docs/current/user-guide/#writing-tests-interface-default-methods.

kcooney:
I agree with Marc.

ollierob:
In the absence of a JUnit 5 release why not add it now? It's a tiny change to add some major functionality, and if anything helps us write JUnit-5-ey tests right now.

kcooney:
It's possible that JUnit 5 will be released before JUnit 4.13, so adding this class to JUnit 4 might not make it available to you any time sooner. JUnit 5 also has other APIs that make better use of Java 8 features, so we are targeting Java 8 improvements to JUnit 5.

As your pull demonstrates, if you want to define tests via default methods you can write a custom runner.

marcphilipp:
I'm closing this pull request since the feature will be addressed by JUnit 5.

Issue #946, opened 2014-06-30
TemporaryFolder.newFolder(String/String...) gives a poor error message when passed paths that contain slashes


paulduffin:
If you call TemporaryFolder.newFolder("foo/bar") in JUnit 4.10 the method returns a File object for the new folder but actually fails to create it. That is contrary to the expected behaviour of the method which is to actually create the folder. In JUnit 4.11 the same call throws an exception. Nowhere in the documentation does it explain that the String(s) passed to that method can only be single path components.

PR #1402, opened 2017-01-01
Update TemporaryFolder.newFolder(String) to support passing in paths

kcooney:
that include path separator characters. The overload of newFolder() that
supports passing in multiple strings still does not allow path separators.

### DIFF
--- a/src/main/java/org/junit/rules/TemporaryFolder.java
+++ b/src/main/java/org/junit/rules/TemporaryFolder.java
@@ -167,16 +167,27 @@ public class TemporaryFolder extends ExternalResource {
     }
 
     /**
-     * Returns a new fresh folder with the given name under the temporary
+     * Returns a new fresh folder with the given path under the temporary
      * folder.
      */
     public File newFolder(String folder) throws IOException {
-        return newFolder(new String[]{folder});
+        if (new File(folder).isAbsolute()) {
+            throw new IOException("folder name must be a relative path");
+        }
+        File file = new File(getRoot(), folder);
+        if (!file.mkdirs()) {
+            throw new IOException(
+                    "a folder with the name \'" + folder + "\' already exists");
+        }
+        return file;
     }
 
     /**
      * Returns a new fresh folder with the given name(s) under the temporary
-     * folder.
+     * folder. For example, if you pass in the strings {@code "parent"} and {@code "child"}
+     * then a directory named {@code "parent"} will be created under the temporary folder
+     * and a directory named {@code "child"} will be created under the newly-created
+     * {@code "parent"} directory.
      */
     public File newFolder(String... folderNames) throws IOException {
         File file = getRoot();
--- a/src/test/java/org/junit/rules/TemporaryFolderUsageTest.java
+++ b/src/test/java/org/junit/rules/TemporaryFolderUsageTest.java
@@ -84,12 +84,32 @@ public class TemporaryFolderUsageTest {
     }
     
     @Test
-    public void newFolderWithGivenFolderThrowsIOExceptionIfFolderNameConsistsOfMultiplePathComponents()
+    public void newFolderWithPathStartingWithFileSeparatorThrowsIOException()
             throws IOException {
         tempFolder.create();
         thrown.expect(IOException.class);
-        thrown.expectMessage("name cannot consist of multiple path components");
+        thrown.expectMessage("folder name must be a relative path");
+        tempFolder.newFolder(File.separator + "temp1");
+    }
+    
+    @Test
+    public void newFolderWithPathContainingFileSeparaterCreatesDirectories()
+            throws IOException {
+        tempFolder.create();
+        tempFolder.newFolder("temp1" + File.separator + "temp2");
+        File temp1 = new File(tempFolder.getRoot(), "temp1");
+        assertFileIsDirectory(temp1);
+        assertFileIsDirectory(new File(temp1, "temp2"));
+    }
+ 
+    @Test
+    public void newFolderWithPathContainingForwardSlashCreatesDirectories()
+            throws IOException {
+        tempFolder.create();
         tempFolder.newFolder("temp1/temp2");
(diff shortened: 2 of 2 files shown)


src/main/java/org/junit/rules/TemporaryFolder.java
```diff
@@ -167,16 +169,22 @@ public File newFile() throws IOException {
...
-     * Returns a new fresh folder with the given name under the temporary
+     * Returns a new fresh folder with the given path under the temporary
      * folder.
      */
     public File newFolder(String folder) throws IOException {
-        return newFolder(new String[]{folder});
+        if (folder.startsWith(File.separator)) {
+            throw new IOException("Folder name cannot start with a file separator");
+        }
+        return newFolder(folder.split(FILE_SEPARATOR_REGEX));
```
marcphilipp: Instead of the regex, can we use `new File(folder)` and call `getParent()` until it returns `null`? This way a test could use `newFolder("foo/bar")` and be able to run on Windows and Unix.

src/test/java/org/junit/rules/TemporaryFolderUsageTest.java
```diff
@@ -84,12 +84,22 @@ public void newFolderWithGivenFolderThrowsIllegalArgumentExceptionIfFolderExists
...
+    }
+    
+    @Test
+    public void newFolderWithPathContainingFileSeparaterCreatesDirectories()
+            throws IOException {
+        tempFolder.create();
         tempFolder.newFolder("temp1/temp2");
+        File temp1 = new File(tempFolder.getRoot(), "temp1");
+        assertFileIsDirectory(temp1);
+        assertFileIsDirectory(new File(temp1, "temp2"));
```
marcphilipp: This test won't run on Windows, right?

src/main/java/org/junit/rules/TemporaryFolder.java
kcooney: I didn't realize that File supported / as a path separator for Windows.

Instead of your solution, how about I rewite this method to use mkdirs()?

marcphilipp: How would you report already existing intermediate folders (see `newFolder(String[])`) that way?

kcooney: `newFolder(String...)` does not report already existing intermediate directories

marcphilipp: From the code it looks like it does… 😯

kcooney:
BTW, the behavior changes in 4.12 broke some of our tests. When I am back in the office tomorrow I will verify that this change resolves the problems.

src/main/java/org/junit/rules/TemporaryFolder.java
marcphilipp: Never mind...

src/test/java/org/junit/rules/TemporaryFolderUsageTest.java
kcooney: Fixed

src/main/java/org/junit/rules/TemporaryFolder.java
kcooney: Changed to use mkdirs() and added a test for paths that contain a forward slash and for paths that contain the file separator.

kcooney:
PTAL. Thanks, @marcphilipp !

kcooney:
I verified that this is a regression from the behavior in JUnit 4.11 and changed the commit comment accordingly.

I also removed the use of Pattern (I forgot to remove that when I changed the code to use `mkdirs`).

Any objections to merging this?

Note I'm also considering changing newFolder(String...) to support paths with slashes, since the inconsistency is weird and supporting slashes would be consistent with `java.nio.file.Paths.get(String, String...)`

src/main/java/org/junit/rules/TemporaryFolder.java
```diff
@@ -167,16 +167,27 @@ public File newFile() throws IOException {
     }
 
     /**
-     * Returns a new fresh folder with the given name under the temporary
+     * Returns a new fresh folder with the given path under the temporary
      * folder.
      */
     public File newFolder(String folder) throws IOException {
```
marcphilipp: We should probably rename `folder` to `path`, right?

src/main/java/org/junit/rules/TemporaryFolder.java
```diff
@@ -167,16 +167,27 @@ public File newFile() throws IOException {
...
-        return newFolder(new String[]{folder});
+        if (new File(folder).isAbsolute()) {
+            throw new IOException("folder name must be a relative path");
+        }
+        File file = new File(getRoot(), folder);
+        if (!file.mkdirs()) {
+            throw new IOException(
+                    "a folder with the name \'" + folder + "\' already exists");
```
marcphilipp: Is this error message really accurate? I know it's the same as in the other method but I'm sure there are other reasons why a folder could not be created.

marcphilipp:
> I verified that this is a regression from the behavior in JUnit 4.11 and changed the commit comment accordingly.
> 
> I also removed the use of `Pattern` (I forgot to remove that when I changed the code to use `mkdirs`).

👍

> Any objections to merging this?

None at all.

> Note I'm also considering changing newFolder(String...) to support paths with slashes, since the inconsistency is weird and supporting slashes would be consistent with `java.nio.file.Paths.get(String, String...)`

I think supporting slashes is great! I've tried to find out why it was implemented that way. From #946 it seems 4.10 used only `mkdir`, 4.11 introduced `newFolder(String...)` but kept using `mkdir`. Supporting slashes should be okay because it would only be more tolerant than before and no existing call should break.

kcooney:
@marcphilipp I updated the release notes.

Sorry, I somehow didn't see some of your comments. I'll address them soon.

marcphilipp:
Thanks! And no worries about those comments. 🙂

PR #503, opened 2012-09-12
Configurable Categories

Tibor17:
Categories in suite should be multiple in annotations Categories.IncludeCategory and Categories.ExcludeCategory, and configurable by system properties as well.
This pull request referst to discussions in #142 and my old pull request #354 closed by my mistake #461.

PR #1403, opened 2017-01-02
Add public CategoryFilter constructor that was removed in JUnit 4.12

kcooney:
Note that this constructor was removed in #503

### DIFF
--- a/src/main/java/org/junit/experimental/categories/Categories.java
+++ b/src/main/java/org/junit/experimental/categories/Categories.java
@@ -5,6 +5,7 @@ import java.lang.annotation.RetentionPolicy;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Set;
 
 import org.junit.runner.Description;
@@ -295,7 +296,7 @@ public class Categories extends Suite {
         }
 
         private static Set<Class<?>> copyAndRefine(Set<Class<?>> classes) {
-            Set<Class<?>> c= new HashSet<Class<?>>();
+            Set<Class<?>> c= new LinkedHashSet<Class<?>>();
             if (classes != null) {
                 c.addAll(classes);
             }
@@ -361,7 +362,7 @@ public class Categories extends Suite {
 
         return classes.length == 1
             ? Collections.<Class<?>>singleton(classes[0])
-            : new HashSet<Class<?>>(Arrays.asList(classes));
+            : new LinkedHashSet<Class<?>>(Arrays.asList(classes));
     }
 
     private static Set<Class<?>> nullableClassToSet(Class<?> nullableClass) {
--- a/src/main/java/org/junit/experimental/categories/Categories.java
+++ b/src/main/java/org/junit/experimental/categories/Categories.java
@@ -2,6 +2,7 @@ package org.junit.experimental.categories;
 
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


src/main/java/org/junit/experimental/categories/Categories.java
```diff
@@ -346,11 +347,28 @@ private static boolean hasAssignableTo(Set<Class<?>> assigns, Class<?> to) {
         return false;
     }
 
-    private static Set<Class<?>> createSet(Class<?>... t) {
-        final Set<Class<?>> set= new HashSet<Class<?>>();
-        if (t != null) {
-            Collections.addAll(set, t);
+    private static Set<Class<?>> createSet(Class<?>[] t) {
```
marcphilipp: While you're at it, can you please rename `t` to something with a little more meaning?

src/main/java/org/junit/experimental/categories/Categories.java
```diff
@@ -346,11 +347,28 @@ private static boolean hasAssignableTo(Set<Class<?>> assigns, Class<?> to) {
...
+        for (Class<?> category : t) {
+            if (category == null) {
+                throw new NullPointerException("has null category");
+            }
         }
-        return set;
+
+        return t.length == 1
+            ? Collections.<Class<?>>singleton(t[0])
+            : new HashSet<Class<?>>(Arrays.asList(t));
```
marcphilipp: How about a `LinkedHashSet` so the iteration order stays intact?

src/main/java/org/junit/experimental/categories/Categories.java
```diff
@@ -151,14 +146,30 @@ public static CategoryFilter categoryFilter(boolean matchAnyInclusions, Set<Clas
...
         }
 
+        private CategoryFilter(boolean matchAnyIncludes, Class<?>[] inclusions,
+                               boolean matchAnyExcludes, Class<?>[] exclusions) {
+            includedAny = matchAnyIncludes; 
+            excludedAny = matchAnyExcludes;
+            included = createSet(inclusions);
+            excluded = createSet(exclusions);
+        }
+
```
marcphilipp: Is it possible to remove the duplication in these constructors by having only a single private one that does the assignments?

kcooney: I think doing that would sacrifice readability

src/main/java/org/junit/experimental/categories/Categories.java
kcooney: The previous code used HashSet for this case. I don't think ordering matters, since these are inclusion/exclusion filters. Am I missing something?

marcphilipp: You're right: The ordering doesn't really matter. In general, I like to preserve it whenever possible so it's easier to debug/test and to not confuse users with a different order in error messages. However, in this case using `HashSet` should be fine.

src/main/java/org/junit/experimental/categories/Categories.java
marcphilipp: Ok

src/main/java/org/junit/experimental/categories/Categories.java
kcooney: Done

src/main/java/org/junit/experimental/categories/Categories.java
kcooney: Good point about debugging. The `toString()` and `describe()` methods do print out the categories. Printing them in the same order makes sense. Changed to use `LinkedHashSet` in a separate commit

src/main/java/org/junit/experimental/categories/Categories.java
kcooney: Done

src/main/java/org/junit/experimental/categories/Categories.java
marcphilipp: 👍

PR #1404, opened 2017-01-02
Update Classes.getClass() to handle a null context class loader.

kcooney:


marcphilipp:
What's the use case for this change?

kcooney:
I have see cases where I changed code from using Class.forName() to using the context class loader, and that resulted in a NPE for some callers because the context class loader is null.

marcphilipp:
Well, in this case, when the context class loader is null `Class.forName(String, boolean, ClassLoader)` will use the bootstrap `ClassLoader` and there won't be a NPE.

> If the parameter `loader` is null, the class is loaded through the bootstrap class loader.

kcooney:
@marcphilipp good point, but the bootstrap classloader is never the right classloader to use unless you are loading JDK classes.

kcooney:
BTW, Happy New Year @marcphilipp 📣🎆

marcphilipp:
Thanks, I wish you a very happy new year, too! 😊

> good point, but the bootstrap classloader is never the right classloader to use unless you are loading JDK classes.

Sure, but how do you know, the `ClassLoader` of the `CategoryFilterFactory` class is any better?

kcooney:
So it's my understanding that Class.forName() uses the calling class's classloader. If we used that in Classss.java then it would use the same classloader as JUnit, which is better than the bootstrap classloader. The reason I special-cased `CategoryFilterFactory` is because someone could in theory subclass it, and if they did their subclass could in theory have a different class loader (with the classloader that loaded the JUnit classes as it's (grand)parent)

kcooney:
I renamed the commit to make things more clear, and added a comment in `CategoryFilterFactory` to explain what's going on.

PR #1406, opened 2017-01-06
TemporaryFolder.newFolder(String...) supports path separator

kcooney:


src/test/java/org/junit/rules/TemporaryFolderUsageTest.java
```diff
@@ -113,22 +124,46 @@ public void newFolderWithPathContainingForwardSlashCreatesDirectories()
...
+        tempFolder.newFolder("temp1", "temp2", "temp3" + File.separator + "temp4");
+
+        File directory = new File(tempFolder.getRoot(), "temp1");
+        assertFileIsDirectory(directory);
+        directory = new File(directory, "temp2");
+        assertFileIsDirectory(directory);
+        directory = new File(directory, "temp3");
+        assertFileIsDirectory(directory);
+        directory = new File(directory, "temp4");
+        assertFileIsDirectory(directory);
```
marcphilipp: Wouldn't it be sufficient have a single assertion: `assertFileIsDirectory(new File(tempFolder.getRoot(), "temp1/temp2/temp3/temp4")`?

src/main/java/org/junit/rules/TemporaryFolder.java
```diff
@@ -170,56 +170,45 @@ public File newFile() throws IOException {
...
+            if (new File(path).isAbsolute()) {
+                throw new IOException("folder path \'" + path + "\' is not a relative path");
+            }
+        }
+
+        File relativePath = null;
+        File file = root;
+        boolean lastMkdirsCallSuccessful = true;
+        for (int i = 0; i < paths.length; i++) {
+            relativePath = relativePath == null ? new File(paths[i]) : new File(relativePath, paths[i]);
```
marcphilipp: According to the Javadoc of `File` this can be simplified to `relativePath = new File(relativePath, paths[i])`.

src/main/java/org/junit/rules/TemporaryFolder.java
```diff
@@ -170,56 +170,45 @@ public File newFile() throws IOException {
...
-            if (!file.mkdir() && isLastElementInArray(i, folderNames)) {
+    public File newFolder(String... paths) throws IOException {
+        // Before checking the paths, check if create() was ever called, and if it wasn't, throw
+        File root = getRoot();
+
+        for (String path : paths) {
+            if (new File(path).isAbsolute()) {
+                throw new IOException("folder path \'" + path + "\' is not a relative path");
+            }
+        }
```
marcphilipp: I think we should also add a check for `paths.length > 0`. Because in that case we currently return `getRoot()` which isn't a "new fresh folder".

src/test/java/org/junit/rules/TemporaryFolderUsageTest.java
kcooney: I wasn't 100% certain that would work on all platforms

src/main/java/org/junit/rules/TemporaryFolder.java
kcooney: Thanks! Done.

src/main/java/org/junit/rules/TemporaryFolder.java
kcooney: Yes, I noticed that was the current behavior, too. I wasn't sure you wanted that in the same pull. I added the fix in a new commit. I also added a commit fixing some incorrect test method names.

src/test/java/org/junit/rules/TemporaryFolderUsageTest.java
marcphilipp: I'm pretty sure it should. Can we assume it does until proven otherwise?

src/test/java/org/junit/rules/TemporaryFolderUsageTest.java
```diff
@@ -63,7 +63,7 @@ public void newFileWithGivenFilenameThrowsIllegalArgumentExceptionIfFileExists()
     }
 
     @Test(expected = IllegalStateException.class)
-    public void newFolderThrowsIllegalStateExceptionIfCreateWasNotInvoked()
+    public void newFolderThrowsIOExceptionIfCreateWasNotInvoked()
```
marcphilipp: This one actually does throw an `IllegalStateException`.

kcooney: Woops! Fixed

src/test/java/org/junit/rules/TemporaryFolderUsageTest.java
kcooney: Sure. I have no way to verify on Windows though.

src/main/java/org/junit/rules/TemporaryFolder.java
```diff
@@ -170,56 +170,51 @@ public File newFile() throws IOException {
...
-            if (!file.mkdir() && isLastElementInArray(i, folderNames)) {
+    public File newFolder(String... paths) throws IOException {
+        if (paths.length == 0) {
+            throw new IllegalArgumentException("must pass at least one path");
+        }
+
+        /*
+         * Before checking if the paths are absolute paths, check if create() was ever called,
+         * and if it wasn't, throw IllegalStateException.
+         */
```
stefanbirkner: I think the comment is wrong because I could not find the `IllegalStateException`.

kcooney: `getRoot()` throws `IllegalStateException` if `before()` wasn't (yet?) called.

stefanbirkner:
That's a nice improvement. Thanks.

PR #1408, opened 2017-01-09
Classes annotated with @RunWith(Suite.class) do not need to be public.

kcooney:
This fixes a regression in JUnit 4.12.


kcooney:
I updated the comment message. @stefanbirkner if the new commit message looks fine, please merge.

PR #1409, opened 2017-01-09
Fixed failing tests when running build on Windows.

bangnab:
Note that File.separator is not sufficient to build an absolute path on windows.
String.format with "%n" (used in MultipleFailureException.getMessage()) produces a platform dependent line separator, i.e. not necessarily "/n";

### DIFF
--- a/src/test/java/org/junit/rules/TemporaryFolderUsageTest.java
+++ b/src/test/java/org/junit/rules/TemporaryFolderUsageTest.java
@@ -20,7 +20,6 @@ import org.junit.Test;
  */
 public class TemporaryFolderUsageTest {
 
-    public static final String ROOT_PATH = File.listRoots()[0].getAbsolutePath();
     private TemporaryFolder tempFolder;
 
     @Rule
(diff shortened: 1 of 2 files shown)


src/test/java/org/junit/rules/TemporaryFolderUsageTest.java
```diff
@@ -20,6 +20,7 @@
  */
 public class TemporaryFolderUsageTest {
 
+    public static final String ROOT_PATH = File.listRoots()[0].getAbsolutePath();
```
kcooney: Please make the a local variable in `newFolderWithPathStartingWithFileSeparatorThrowsIOException()`

kcooney: I meant "Please make _this_ a local..."

bangnab:
Maybe I was a little too quick and I should have changed the local variable name to camel case...

kcooney:
@bangnab I updated the code in the branch to use camel case and added some sanity checks. I'll squash the commits when we commit (assuming it passes)

kcooney:
Merged. Thanks!

Issue #1411, opened 2017-01-11
Enforce order of @Before/@After annotated methods even when override is also annotated


paulduffin:
I recently had a problem where a fix to a compiler that copied annotations from a method onto a synthetic bridge method created by the compiler changed the behavior of a test.

I had two classes, lets call them Base and Derived. Each of them had an @Before annotated method, lets call them baseBefore() and derivedBefore() respectively. The order in which they were executed was critical, Base.baseBefore() must be invoked prior to Derived.derivedBefore(). While the order of invocation of the methods within a class is not defined, @Before annotated methods on a base class are always invoked before such methods on a derived class so everything worked fine.

The Base class is package protected so a synthetic bridge method is created in Derived that overrides the Base.baseBefore() method. Previously, that method did not have any annotations so was ignored by JUnit. However, the fix to copy annotations onto the bridge method now causes JUnit to see that synthetic bridge method and that causes it to ignore the method from Base because it was shadowed by the synthetic method. That also changes the order in which the methods are invoked causing the tests to fail. 

A simple solution is to exclude synthetic methods from the ones returned by JUnit but it seems as though the existing logic is flawed. Basically, as implemented overriding an @Before annotated method will not change the method order but adding an @Before annotation to it could change the order.

It seems to me that even when overridden and annotated with @Before that an @Before method declared on the base class should be run before any methods on the derived class.

kcooney:
I don't think I agree with the last sentence. If you override a `@Before` method you shouldn't annotate the method on the subclass with `@Before`.  If you do, then I am personally fine with the call order being undefined.

Personally, I recommend always making Before and After methods final. If subclasses need to change or replace the base class behavior, extract a non-final protected method and have your. Before/After method call the extracted method. My personal preference is to make the extracted method abstract or, if non-abstract, have it do nothing.

I can't speak to the compiler issue. Are there reasons outside of JUnit to copy the annotations?

paulduffin:
The background for the compiler issue can be found at:
https://bugs.openjdk.java.net/browse/JDK-6695379
https://bugs.eclipse.org/bugs/show_bug.cgi?id=495396

The compiler change was not made with JUnit in mind, and they're certainly not going to revert it for JUnit.

I agree that making Before/After methods final is better but it's not possible in the specific case that I mentioned above because JUnit cannot invoke an @Before method on a package protected class. Marking the Base.baseBefore() method final would prevent the creation of the bridge method and so prevent JUnit from being able to access the method.

[...]

PR #1412, opened 2017-01-13
Fix for issue 1411

kcooney:


kcooney:
I fixed a spelling error, then added commits to fix the code so static methods are never considered shadowed and fixed the code so fields are never considered shadowed.

kcooney:
This is one possible fix for #1411
The other possible fix is #1413

kcooney:
Closed in favor of #1413

PR #1413, opened 2017-01-17
Ignore bridge methods when scanning for annotated methods.

kcooney:
This is one possible fix for #1411
The other possible fix is #1412

### DIFF
--- a/src/main/java/org/junit/runners/model/FrameworkMethod.java
+++ b/src/main/java/org/junit/runners/model/FrameworkMethod.java
@@ -149,6 +149,11 @@ public class FrameworkMethod extends FrameworkMember<FrameworkMethod> {
     }
 
     @Override
+    boolean isBridgeMethod() {
+        return method.isBridge();
+    }
+
+    @Override
     public boolean equals(Object obj) {
         if (!FrameworkMethod.class.isInstance(obj)) {
             return false;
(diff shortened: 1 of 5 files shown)


src/main/java/org/junit/runners/model/FrameworkMethod.java
```diff
@@ -149,6 +149,11 @@ public boolean isShadowedBy(FrameworkMethod other) {
     }
 
     @Override
+    boolean isBridgeMethod() {
```
sormuras: @kcooney -- i found an use-case where "isBridge()" is not enough. I think, it should be widened to "isSynthetic()". See https://github.com/junit-team/junit5/compare/ignore_synthetic_methods with a test case included.

kcooney: JUnit4 doesn't support using `@Test`, `@Before` etc on anything but "normal" methods.

sormuras: ...and JUnit 4 does not provide [support](https://github.com/junit-team/junit5/tree/master/junit-platform-commons/src/main/java/org/junit/platform/commons/support) tools like JUnit 5 now does.

sormuras: @kcooney Is the assumption correct, that every bridge method is also synthetic? Couldn't find a definitive answer to that question in the language spec or via googling, yet.

kcooney: @sormuras synetetic methods are any methods introduced by the compiler that do not correspond to something in the source (with the exception of default constructors and static initializers). Therefore any bridge method is a synthetic method.

PR #1414, opened 2017-01-23
Only non static methods can shadow

kcooney:


stefanbirkner:
Shadowing for static fields has been restored by ed6813d3bbe2da4d67892f73e641fa80ab4d126e.

Issue #1221, opened 2015-11-21
RuleChain annotation equivalent


rmannibucau:
Idea is to be able to chain rules but keeping track of each instance

One quick and simple but efficient solution is to add to @Rule an index/order method:

```
@Rule(index = 0)
private MyRule superRule = new MyRule(...);

@Rule(index = 1)
private MyRule2 superRule2 = new MyRule2(...);
```

 it allows to have the chain well formed automatically and the @Test to still use the rule instance to do things/access rule data.

Tibor17:
This would be an elegant solution (default index=0) for ordering problem for produce methods/fields of rules and new proposal @rules on test method https://github.com/junit-team/junit/issues/1074#issuecomment-158158565.
WDYT?

Tibor17:
btw. Nice to have the field private as in Roman's example.
I do not see any reason why JUnit requires public modifier on this field, maybe because they do not use setAccessible(true) which anyway would not harm the runtime of the test.

kcooney:
We require methods and fields that are annotated with `@Test`, `@Rule` etc to be public because `setAccessible(true)` might not work if the tests run in an environment with a security manager installed. It would be unfortunate if tests passed in one environment (an IDE, for example) but not in another environment (your build tool).

One way to think about this: test methods and rules are functionality that are exposed by your test to JUnit, so they are an external API, so should be public.

Tibor17:
@kcooney you'r right with security manager, thx.

ctrimble:
It would be very cool if rules allowed you to specify the dependencies by name and let JUnit work out the correct order, sorting by name for unrelated rules.  For instance:

```
@Rule
MyRule1 rule1 = new MyRule1();

@Rule(before=["rule1"])
MyRule2 rule2 = new MyRule2();

@Rule(after=["rule2"])
MyRule3 rule3 = new MyRule3();
```

With the resulting execution order being:

```
rule2
rule1
rule3
```

ctrimble:
I have this class in an old framework that can do the sorting https://github.com/ctrimble/xchain/blob/master/core/src/main/java/org/xchain/framework/util/DependencySorter.java .  That code is under the Apache License, but that could be worked out.

stefanbirkner:
Fixed by #1445.

Issue #1045, opened 2014-12-13
Add @Rule(around="otherRule") to specify order


shicks:
By naming another @Rule field/method in an annotation, this would allow topologically sorting the rules to provide a deterministic order. Should probably apply to @ClassRule, too.

FeuRenard:
PR #1418 was rejected because you don't like the feature. As a consequence this issue should be closed.

stefanbirkner:
You're right. Thanks.

kcooney:
Note for other people that may be following along: we are considering other ways to specify the order of rules via the `@Rule` annotation. See #1417

PR #1417, opened 2017-02-07
allow RuleChain construction by annotation: @Rule(priority=42)

SchmittSimon:
This PR addresses #1221. 
This Feature is very similar to #1418 for Issue #1045

I extended the `@Rule` interface with an optional parameter for autoconstructing a `RuleChain` out of the annotated `TestRules`.
The parameter represents a priority and the order of the `RuleChain` will be accordingly.
Highest priority becomes most outer `TestRule` in `RuleChain`.
Negative priorities will be ignored and not put into the `RuleChain`.

For the implementation I introduced a new class `SortRules` in the rule package.
This class takes care of the actual sorting and `RuleChain`-construction.
Also a minor refactoring was necessary for the implementation.

In addition I added Tests to check the functionality of the implementation.

Please let me know what you think about this Implementation and if it is any useful for the project.

This Issue was part of a project in university. Two other classmates worked on JUnit as well.
You can check their PR at #1418 and #xxxx.

### DIFF
--- a/src/main/java/org/junit/rules/SortRules.java
+++ b/src/main/java/org/junit/rules/SortRules.java
@@ -1,0 +1,86 @@
+package org.junit.rules;
+
+import static org.junit.rules.RuleChain.emptyRuleChain;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+
+/*
+ * Stores a list of TestRule-priorities-pairs
+ * and can return a lists of all TestRules
+ * as well as a list containing the autoconstructed RuleChain.
+ */
+public class SortRules {
+    private final List<RulePriorityPair> rules = new ArrayList<RulePriorityPair>();
+
+    /*
+     * Adds a TestRulePriorityPair into the List.
+     * @param rule: the TestRule to be added to SortRules
+     * @param priority: the annotated priority of the TestRule
+     */
+    public void add(TestRule rule, int priority) {
+        rules.add(new RulePriorityPair(rule,priority));
+    }
+
+    /*
+     * Returns the a list of all the stored TestRules in insertion order.
+     * @return A list of TestRules in the order of their insertion.
+     */
+    public List<TestRule> listWithAllRules() {
+        List<TestRule> result = new ArrayList<TestRule>(rules.size());
+        for (RulePriorityPair rule : rules) {
+            result.add(rule.testRule);
+        }
+        return result;
+    }
+
+    /*
+     * Construct a list of all TestRules with negative priorities
+     * plus a autoconstructed RuleChain containing all TestRules with non-negative priorities.
+     * The TestRules of the RuleChain are sorted in such a way
+     * that the TestRule with the highest priority becomes the most outer TestRule.
+     */
+    public List<TestRule> listWithAutoconstructedRuleChain() {
+        List<TestRule> sortedRules = new ArrayList<TestRule>();
+        Integer[] index = new Integer[rules.size()];
+        for( int i = 0 ; i < index.length; i++ ) {
+            index[i] = i;
+        }
+        Arrays.sort(index, new Comparator<Integer>() {
+            public int compare(Integer i1, Integer i2) {
+                return rules.get(i1).compareTo(rules.get(i2));
+            }
+        });
+        RuleChain chain = emptyRuleChain();
+        for( int i: index) {
+            if (rules.get(i).priority >= 0) {
+                chain = chain.around(rules.get(i).testRule);
+            } else {
+                sortedRules.add(rules.get(i).testRule);
+            }
+        }
+        if (!chain.equals(emptyRuleChain())) {
+            sortedRules.add(chain);
+        }
+        return sortedRules;
+    }
+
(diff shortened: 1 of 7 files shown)


src/main/java/org/junit/rules/SortRules.java
```diff
@@ -0,0 +1,86 @@
...
+        return result;
+    }
+
+    /*
+     * Construct a list of all TestRules with negative priorities
+     * plus a autoconstructed RuleChain containing all TestRules with non-negative priorities.
+     * The TestRules of the RuleChain are sorted in such a way
+     * that the TestRule with the highest priority becomes the most outer TestRule.
+     */
+    public List<TestRule> listWithAutoconstructedRuleChain() {
```
kcooney: This is a bit of an awkward name. How about we make this method package-scope, and add a public method named `toStatement()` with class this method and then does the work that is currently done in `withTestRules(FrameworkMethod, List<TestRule>, Statement)`?

src/test/java/org/junit/rules/SortRulesTest.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+      }
+
+    @Test
+    public final void testSortListWithoutPriorities() {
+        SortRules testObject = new SortRules();
+        testObject.add(new IdentityRule(), -1);
+        List<TestRule> result = testObject.listWithAutoconstructedRuleChain();
+        //only the IdentityRule is inserted and not an empty RuleChain.
+        assertEquals(result.size(),1);
```
panchenko: expected should be the first parameter, here and below as well.

panchenko:
I would suggest also discussing if highest number is outer (like implemented currently) or inner (might be together with renaming the annotation parameter to `index`). The rationale for me is that if rules are annotated with 1 and 2 then execution order is:

* 1 before
* 2 before
* test
* 2 after
* 1 after

The thing is there is no common approach how priorities are sorted :-)

FeuRenard:
> My primary concern with this approach is that it will only work with rules that implement `TestRule`, and not rules that implement `MethodRule`. While this problem also is the case with `RuleChain`, if you try to use `RuleChain` with a rule that doesn't implement `TestRule` you will get a compile error.
> 
> To address that concern, we should probably do a validation check to ensure no `MethodRule` fields/methods are annotated with a Rule annotation with a priority. See `RuleMemberValidator` for the current rule validation checks.

Why not build upon #1349 and include `MethodRule`s?

kcooney:
@FeuRenard #1349 doesn't work with methods with test parameters, so I am not sure we will merge it.

One possibility is to combine the rules as a series of `Statement` instances instead of a series of `TestRule` instances.

In any case, the tricky part would be to make these changes without breaking external subclasses of our runners. Allowing users to extend JUnit by subclassing concrete classes has really limited our availability to add features. That is one of the reasons why JUnit 5 was a complete rewrite.

panchenko:
It should be easier to support ordering of MethodRule this way, than via RuleChain. 
The runner is currently capable of calling both and this change is internal to runner.

src/main/java/org/junit/rules/SortRules.java
```diff
@@ -0,0 +1,86 @@
...
+        private RulePriorityPair(TestRule testRule, int priority) {
+            this.testRule = testRule;
+            this.priority = priority;
+        }
+
+        public int compareTo(RulePriorityPair o) {
+            if (this.priority < 0 && o.priority < 0) {
+                return 0;
+            }
+            return o.priority - this.priority;
```
FeuRenard: `Integer.compare(n,n)` was introduced with Java 7 but we have to comply with Java 5.
https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#compare-int-int-

kcooney: Could point. The code here is prone to overflow, though.

kcooney:
Before we go to far along to path of trying to figure out how to implement this change without breaking existing code, I want to make sure we are generally in favor of using the `@Rule` annotation to order rules.

@junit-team/junit-committers any concerns with this?

marcphilipp:
I think this looks good! However, I would call it `order` instead of `priority` and order them the other way around (like suggested above in https://github.com/junit-team/junit4/pull/1417#issuecomment-278554911).

stefanbirkner:
I like the idea of a parameter for the `@Rule` annotation, too. I would it neither call `order` nor `priority`, but something like `level` with the value `1` or `0` for the innermost rule. The naming of the parameter should help people to understand that rules are nested and not something that is executed sequentially.

kcooney:
@SchmittSimon any thoughts on resolving the issues raised on this pull? It would be great to get it in.

I personally prefer "priority" over "level" or "order". I haven't heard "level" used outside of logging frameworks (which have predefined levels). Using "order" can cause confusion (is it the apply order or the order of the final chain?). The name "priority" makes it clear that the numerical value is used to order the rules, but the reader will naturally wonder if low numbers are first or last, so they will hopefully read the JavaDoc to see how we interpret the values.

panchenko:
@SchmittSimon if you don't have time currently, would it be OK if I continue this effort?

kcooney:
Closing in favor of #1445

PR #1418, opened 2017-02-07
Allow RuleChain building by @Rule(around="rule")

### LINKED ISSUE #1221: RuleChain annotation equivalent

FeuRenard:
This PR addresses #1045 and is similar to PR #1417, which attempts to resolve #1221.

I extended the `@Rule` interface with an around parameter as suggested. Then I added tests for the behaviour I expect from this feature. I did not implement and/or test the interaction with a superclass (this.\*, super.\*). The actual implementation introduces a new class `SortRules` to the rule package. It handles all `TestRules` of a test class and sorts them when the according method is called. A topological sorting algorithm is implemented, whose performance could possibly be further enhanced.

As I did not discuss the implementation with you before I started working on it, I could totally comprehend a rejection of this PR. Otherwise I am willing to adapt the PR to your wishes :)
I developed this feature implementation in the context of a university lecture on software evolution in collaboration with the lecturers and two classmates. The two classmates also worked on their own issues, you can find their PRs here: #1417, #xxxx (tba)

### DIFF
--- a/src/main/java/org/junit/Rule.java
+++ b/src/main/java/org/junit/Rule.java
@@ -69,4 +69,10 @@ import java.lang.annotation.Target;
 @Target({ElementType.FIELD, ElementType.METHOD})
 public @interface Rule {
 
+    /**
+     * Optionally specify <code>around</code>, the name of a TestRule,
+     * to chain the annotated TestRule around the specified TestRule.
+     * See also {@link org.junit.rules.RuleChain RuleChains} for more details.
+     */
+    String around() default "";
 }
--- a/src/main/java/org/junit/rules/SortRules.java
+++ b/src/main/java/org/junit/rules/SortRules.java
@@ -1,0 +1,112 @@
+package org.junit.rules;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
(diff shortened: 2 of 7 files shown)


kcooney:
Thanks for the contribution!

My main concern with this approach is that renaming fields will result in dramatically different test behavior (and, most likely, broken tests). I think that would be quite unexpected from most user's expectations. In contrast to JUnit3-style tests, test and field names in JUnit4-style tests don't matter; only the annotations do. 

Because of these concerns, I unfortunately recommend against merging these changes.

@junit-team/junit-committers any thoughts?

panchenko:
The other approach with the numeric priorities feels easier to use.

marcphilipp:
I also think #1417 is easier to use and more refactoring-friendly.

Thus, I'm closing this pull request.

@FeuRenard Thanks for the proposal!

PR #1421, opened 2017-02-19
ExternalResource: declare after() to throw Exception

alb-i986:
So that clients don't have to try-catch-and-rethrow explicitly checked exceptions.

You may not be happy with changing the signature, though :(

### DIFF
--- a/src/main/java/org/junit/rules/ExternalResource.java
+++ b/src/main/java/org/junit/rules/ExternalResource.java
@@ -78,9 +78,9 @@ public abstract class ExternalResource implements TestRule {
     /**
      * Override to tear down your specific external resource.
      *
-     * @throws Exception if teardown fails
+     * @throws Throwable if teardown fails
      */
-    protected void after() throws Exception {
+    protected void after() throws Throwable {
         // do nothing
     }
 }
--- a/src/main/java/org/junit/rules/ExternalResource.java
+++ b/src/main/java/org/junit/rules/ExternalResource.java
@@ -77,8 +77,10 @@ public abstract class ExternalResource implements TestRule {
 
     /**
      * Override to tear down your specific external resource.
+     *
+     * @throws Exception if teardown fails
      */
-    protected void after() {
+    protected void after() throws Exception {
         // do nothing
     }
 }


src/main/java/org/junit/rules/ExternalResource.java
```diff
@@ -77,8 +77,10 @@ protected void before() throws Throwable {
 
     /**
      * Override to tear down your specific external resource.
+     *
+     * @throws Exception if teardown fails
      */
-    protected void after() {
+    protected void after() throws Exception {
```
kcooney: Why `Exception` instead of `Throwable` like we did for `before()`?

alb-i986: Just because AFAIK rarely methods throw Throwable, it's more likely they throw Exception, e.g. `AutoCloseable#close`.
But ya, we should probably make it more generic and throw Throwable.
Do you want me to?

kcooney: While it's rare for methods to throw `Throwable`, when you call a method using reflection you often handle the `InvocationTargetException` by throwing the cause (a `Throwable`). That's why ` Statement.evaluate()` is declared to throw `Throwable`.

I don't think it harms anything to have the method declared to throw `Throwable`. The calling code catches `Throwable` and subclasses of `ExternalResource` can override the method to throw a different exception.

alb-i986: I don't think that closing a resource usually involves doing reflection or anything that throws Throwable, but anyway yeah, it's harmless, so I'll do.

kcooney:
Merged. @alb-i986 could you please update the release notes?

alb-i986:
Release notes updated.

Cheers

kcooney:
@marcphilipp This breaks code that extends ExternalResource, overrides `after()`, and calls `super.after()` in `after()`. I think we should roll it back.

marcphilipp:
Agreed!

@kcooney Do you have time to do so?

kcooney:
@marcphilipp I thought I might have time but apparently not 🙁

Issue #1219, opened 2015-11-17
Global JUnit rules revisited


tburny:
A word in advance: I never have written such an extensive request in such a major project before, so please bear with me. :) My goal was to provide a comprehensive background as well as concrete suggestions for a solution. I hope that does not seem impolite :)

Due to the topic dating back as far as 2009, I tried to sum up previous discussions, so this issue has become a bit lengthy (sorry :/).

TLDR:
-   some users in the JUnit community would love to have global rules
-   I coin these global rules _Guards_, to differentiate from normal `TestRule`s, `RunListener`s and `Runner`s which mostly are scoped to test class level only
-   Guards should be deactivated by default, with a global opt-in so they don't have to be declared inside every test class.
-   Guards should not cause too much friction with existing code or tools (e.g. Maven Surefire)
-   I would like to contribute time and code to solve the problem (as a prototype at first, then as a real solution)
-   Your feedback is very much appreciated :)
# Problem setting

At our company we want to run a set of global checks against each test to ensure it complies with company standards. If not, the according tests should fail.

Some actual use cases are:
-   nothing is printed to standard error
-   our dependency injection container has been shut down properly in `@After(Class)#tearDown()`).

There are more cases I can name on demand, but I wanted to keep the list short. All these cases have in common that they are some sort of global `TestRule`. In some cases it's not that obvious that a rule has been violated or someone simply forgot a line of code(nobody is perfect).
# History

"Global rules" have been [initially requested in 2009 on the JUnit mailing list](https://groups.yahoo.com/neo/groups/junit/conversations/topics/22133), which resulted in [#69 Global rules](https://github.com/junit-team/junit/issues/69). This discussion then has been continued in [#444 JUnit should have an exhaustive listener framework](https://github.com/junit-team/junit/issues/444) in 2012. I will pick up some suggestions from these issues far further below. Besides the discussion of other use cases, @BenRomberg has been making some important arguments for having such a feature.

Some of the main arguments **against** global rules have been
-   integration with other tools that use their custom runners, like Maven or other IDEs
-   Tests should not fail due to some "obscure" magic (package level annotations, junit.properties).
-   The same could be achieved with: `TestListener`, `Runner`, ...

However it seems a part of the community is asking for a long time. There has been a lot of discussion and I guess that JUnit also evolved much more in the past 3 years (hence "revisited").
# Tried approaches

[...]

Issue #670, opened 2013-05-02
JUnit should process annoations in package-info.java


brianegge:
It would be useful to be able to add package level annotations to package-info.java. For example adding @org.junit.Ignore to the package should ignore all tests in that package.

adrianosimoes:
Is this still a valid issue?

kcooney:
@adrianosimoes I don't think the maintainers are thrilled with the idea of processing package-level annotations. I think we should leave it open in case someone has a use case other than `@Ignore` where it would be useful to process package-level annotations. If `@Ignore` is the only supported feature, then it doesn't seem very compelling IMHO.

kcooney:
One use case just came up (`@RunWith`). I could imagine wanting to specify `@Category` at a package level as well. For `@RunWith` it is clear what happens when the package has the annotation and some classes in the package are also annotated. For `@Category` the desired behavior is less clear.

brianegge:
Supporting package level annotations would be straight forward to do. However, the number of people actually using them would likely be between zero and one. It appears there is little demand for this feature, so it's probably best to close this request.

kcooney:
There is discussion about possibly using them in #1423

cesar1000:
@brianegge I disagree. A package-level @Ignore annotation as you suggested in your original request can be really useful, as there's no convenient way to disable a set of test un current JUnit. So is a @RunWith annotation for defining a default runner for a set of tests.

As to whether this would see adoption: maybe not if annotations need to be added manually, but IDEs could use this feature and expose it through a high level interface.

kcooney:
From a discussion in #1423 

> it appears that Java will only initialize `Package`
> objects when first loading a class from the package.
> This is ok if annotating the package containing the
> tests, but won't work with parent packages (unless
> adding placeholder classes to these packages and
> loading them explicitly, but that's a bad hack).

PR #1423, opened 2017-02-27
Support the selection of a default Runner in ClassRequest

cesar1000:
This patch updates `Request` and `ClassRequest` to enable the configuration of a default JUnit4 Runner other than `BlockJUnit4ClassRunner`. The only visible change is a new static method in Request, `classWithDefaultRunner`, which configures the runner class to use if not configured in the test.

The context of this change is the following: when managing a large test suite that gets contributions from a large development team, it's typical to enforce certain logic across all tests for global state initialization and cleanup. In the particular case of Twitter for Android, all our tests set up global application graphs before test execution and reset static state afterwards. It is critical for correctness that every single test follows this pattern. This seems to be a particular example of the general need for a mechanism that configures the global state of the test environment.

This is typically achieved by forcing all tests to use a single shared runner or to extend a base class. However, these approaches are error prone, since they require dilligence for setting up the test correctly, or an external validation (eg. a checkstyle detector). A default test runner is a low-tech solution to this problem: a test runner extending `BlockJUnit4ClassRunner` implements state management logic and is configured to be used with all tests that don't explicitly say otherwise.

#1219 proposes an alternative solution to a similar problem. I feel that this approach covers the general need in a simple, non-intrusive way.

The change is incomplete and doesn't implement tests, and it's meant a small proof of concept for feedback.

### DIFF
--- a/src/main/java/org/junit/internal/builders/AllDefaultPossibilitiesBuilder.java
+++ b/src/main/java/org/junit/internal/builders/AllDefaultPossibilitiesBuilder.java
@@ -31,6 +31,7 @@ public class AllDefaultPossibilitiesBuilder extends RunnerBuilder {
                 annotatedBuilder(),
                 suiteMethodBuilder(),
                 junit3Builder(),
+                defaultBuilder(),
                 junit4Builder());
 
         for (RunnerBuilder each : builders) {
@@ -56,6 +57,10 @@ public class AllDefaultPossibilitiesBuilder extends RunnerBuilder {
 
     protected IgnoredBuilder ignoredBuilder() {
         return new IgnoredBuilder();
+    }
+
+    protected RunnerBuilder defaultBuilder() {
+        return new NullBuilder();
     }
 
     protected RunnerBuilder suiteMethodBuilder() {
--- a/src/main/java/org/junit/internal/builders/AnnotatedBuilder.java
+++ b/src/main/java/org/junit/internal/builders/AnnotatedBuilder.java
@@ -2,7 +2,6 @@ package org.junit.internal.builders;
 
 import org.junit.runner.RunWith;
 import org.junit.runner.Runner;
-import org.junit.runners.model.InitializationError;
 import org.junit.runners.model.RunnerBuilder;
 
 import java.lang.reflect.Modifier;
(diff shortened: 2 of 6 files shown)


kcooney:
Thanks for the contribution.

My concern with this approach is that if you run tests in a way that won't call this new method (ex: from an IDE) you would get different results. This could lead to confusion. Having your tests explicitly specify a runner avoids this problem.

cesar1000:
Thanks, Kevin!

Yeah, that's a good point. Ideally, the information should be available in code. Would you consider something like looking up a @RunWith annotation applied to a package? In the case that the test is not annotated, the annotation runner would look up the chain of parent packages for the annotation. Implementers can then either add the package info file to the codebase or generate dynamically at compilation time - in either case, the configuration would be picked up by all test execution methods.

kcooney:
@cesar1000 see #670  for a related discussion.

One open question is what to do if your test class extends a test class in a different package and neither test class has `@RunWith`? My gut tells be you should use the ((great) grand) parent's package to find the runner because you are extending the base class, so should inherit the base class's runner (because the runner is part of the behavior of the base clas)s.

cesar1000:
Ah, thanks for the pointer. Yeah, I feel that the meaning of `@RunWith` or `@Ignore` when applied to a package would be well defined. I can see the point that developers may not expect annotations in packages, but I think there are strong practical reasons why teams may decide to use them, since JUnit does not offer alternative tools for applying a certain configuration to a set of tests.

In regard to selecting a runner when extending a test class: I think we should always use the test's package. On the one hand, this is a simple and consistent rule that makes it easy to find the runner. On the other hand, I think it's the responsibility of the implementer of a test to ensure that the test runner is consistent with its parent's (as you would normally do if you were to annotate your test to use a different runner). Do you have any particular case in mind?

If this sounds good, I can throw a prototype together to see how that looks and we can discuss.

cesar1000:
I hit a snag with using package annotations: it appears that Java will only initialize `Package` objects when first loading a class from the package. This is ok if annotating the package containing the tests, but won't work with parent packages (unless adding placeholder classes to these packages and loading them explicitly, but that's a bad hack). Annotating packages containing tests can be useful in itself, but it feels incomplete and error prone, since people are likely to expect this feature to work on any package. Alternatively, we could make the annotation apply only to the package itself and not subpackages, but this limits its usefulness.

An alternative may be to define a named class (JUnitPackageConfiguration or similar) and load that by name. The class would define annotations or other configuration parameters. Would this be an acceptable solution? Do you have any other ideas?

kcooney:
@cesar1000 I think a named package would be even less discoverable to a layperson than using package-level annotations.

cesar1000:
@kcooney yeah, I agree. I was just hoping for a more powerful solution, but I understand that discoverability is an important goal. It's a shame that Java support for package annotations is so inconsistent. Let me give it a bit more thought.

PR #1425, opened 2017-03-01
Refactor BlockJUnit4ClassRunner.possiblyExpectingExceptions

alb-i986:
Extract to local variable the result of getExpectedException(), thus saving one extra call.


marcphilipp:
Thanks again! 👍

PR #1435, opened 2017-04-01
@BeforeParam/@AfterParam for Parameterized runner

panchenko:
This PR addresses issue #45.
Some `@ParamRule` could also be introduced.

### DIFF
--- a/src/main/java/org/junit/internal/runners/statements/RunAfters.java
+++ b/src/main/java/org/junit/internal/runners/statements/RunAfters.java
@@ -39,6 +39,9 @@ public class RunAfters extends Statement {
         MultipleFailureException.assertEmpty(errors);
     }
 
+    /**
+     * @since 4.13
+     */
     protected void invokeMethod(FrameworkMethod method) throws Throwable {
         method.invokeExplosively(target);
     }
--- a/src/main/java/org/junit/internal/runners/statements/RunAfters.java
+++ b/src/main/java/org/junit/internal/runners/statements/RunAfters.java
@@ -30,12 +30,16 @@ public class RunAfters extends Statement {
         } finally {
             for (FrameworkMethod each : afters) {
                 try {
-                    each.invokeExplosively(target);
+                    invokeMethod(each);
                 } catch (Throwable e) {
                     errors.add(e);
                 }
             }
         }
         MultipleFailureException.assertEmpty(errors);
+    }
+
+    protected void invokeMethod(FrameworkMethod method) throws Throwable {
+        method.invokeExplosively(target);
     }
 }
--- a/src/main/java/org/junit/internal/runners/statements/RunBefores.java
+++ b/src/main/java/org/junit/internal/runners/statements/RunBefores.java
@@ -26,6 +26,9 @@ public class RunBefores extends Statement {
         next.evaluate();
     }
 
+    /**
+     * @since 4.13
+     */
     protected void invokeMethod(FrameworkMethod method) throws Throwable {
         method.invokeExplosively(target);
     }
--- a/src/main/java/org/junit/internal/runners/statements/RunBefores.java
+++ b/src/main/java/org/junit/internal/runners/statements/RunBefores.java
@@ -21,8 +21,12 @@ public class RunBefores extends Statement {
     @Override
     public void evaluate() throws Throwable {
         for (FrameworkMethod before : befores) {
-            before.invokeExplosively(target);
+            invokeMethod(before);
         }
         next.evaluate();
+    }
+
+    protected void invokeMethod(FrameworkMethod before) throws Throwable {
+        before.invokeExplosively(target);
     }
 }
--- a/src/main/java/org/junit/internal/runners/statements/RunBefores.java
+++ b/src/main/java/org/junit/internal/runners/statements/RunBefores.java
@@ -26,7 +26,7 @@ public class RunBefores extends Statement {
         next.evaluate();
     }
 
-    protected void invokeMethod(FrameworkMethod before) throws Throwable {
-        before.invokeExplosively(target);
+    protected void invokeMethod(FrameworkMethod method) throws Throwable {
+        method.invokeExplosively(target);
     }
 }
--- a/src/main/java/org/junit/runners/Parameterized.java
+++ b/src/main/java/org/junit/runners/Parameterized.java
@@ -334,7 +334,7 @@ public class Parameterized extends Suite {
 
         private TestWithParameters createTestWithNotNormalizedParameters(
                 String pattern, int index, Object parametersOrSingleParameter) {
-            final Object[] parameters = (parametersOrSingleParameter instanceof Object[]) ? (Object[]) parametersOrSingleParameter
+            Object[] parameters = (parametersOrSingleParameter instanceof Object[]) ? (Object[]) parametersOrSingleParameter
                     : new Object[] { parametersOrSingleParameter };
             if (parameterCount == null) {
                 parameterCount = parameters.length;
--- a/src/main/java/org/junit/runners/Parameterized.java
+++ b/src/main/java/org/junit/runners/Parameterized.java
@@ -12,6 +12,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+
 import org.junit.runner.Runner;
 import org.junit.runners.model.FrameworkMethod;
 import org.junit.runners.model.InvalidTestClassError;
--- a/src/main/java/org/junit/runners/Parameterized.java
+++ b/src/main/java/org/junit/runners/Parameterized.java
@@ -137,7 +137,7 @@ import org.junit.runners.parameterized.TestWithParameters;
  * }
  * </pre>
  *
- * <h3>Executing code before/after executing tests for a specific parameter</h3>
+ * <h3>Executing code before/after executing tests for specific parameters</h3>
  * <p>
  * If your test needs to perform some preparation or cleanup based on the
  * parameters, this can be done by adding public static methods annotated with
--- a/src/main/java/org/junit/runners/Parameterized.java
+++ b/src/main/java/org/junit/runners/Parameterized.java
@@ -137,6 +137,19 @@ import org.junit.runners.parameterized.TestWithParameters;
  * }
  * </pre>
  *
+ * <h3>Executing code before/after executing tests for a specific parameter</h3>
+ * <p>
+ * If your test needs to perform some preparation or cleanup based on the parameters this can be
+ * done by adding public static methods annotated with &#064;BeforeParam/&#064;AfterParam,
+ * such methods should either have no parameters or have same parameters as the test.
+ * <pre>
+ * &#064;BeforeParam
+ * public static void beforeTestsForParameter(String onlyParameter) {
+ *     System.out.println("Testing " + onlyParameter);
+ * }
+ * </pre>
+ * </p>
+ *
  * <h3>Create different runners</h3>
  * <p>
  * By default the {@code Parameterized} runner creates a slightly modified
--- a/src/main/java/org/junit/runners/Parameterized.java
+++ b/src/main/java/org/junit/runners/Parameterized.java
@@ -139,16 +139,16 @@ import org.junit.runners.parameterized.TestWithParameters;
  *
  * <h3>Executing code before/after executing tests for a specific parameter</h3>
  * <p>
- * If your test needs to perform some preparation or cleanup based on the parameters this can be
- * done by adding public static methods annotated with &#064;BeforeParam/&#064;AfterParam,
- * such methods should either have no parameters or have same parameters as the test.
+ * If your test needs to perform some preparation or cleanup based on the
+ * parameters, this can be done by adding public static methods annotated with
+ * {@code @BeforeParam}/{@code @AfterParam}. Such methods should either have no
+ * parameters or the same parameters as the test.
  * <pre>
  * &#064;BeforeParam
  * public static void beforeTestsForParameter(String onlyParameter) {
  *     System.out.println("Testing " + onlyParameter);
  * }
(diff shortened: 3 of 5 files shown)


src/test/java/org/junit/tests/running/classes/ParameterizedTestTest.java
```diff
@@ -260,6 +262,92 @@ public void beforeAndAfterClassAreRun() {
...
+        }
+
+        @Test
+        public void second() {
+            fLog += "second(" + x + ") ";
+        }
+    }
+
+    @Test
+    public void beforeParamAndAfterParamAreRun() {
```
kcooney: It would be great to have tests for

1. Multiple `@BeforeParam` methods
1. Multiple `@AfterParam` methods
1. `@AfterParam` methods with parameters
1. `@BeforeParam` methods without parameters
1. Parameterized tests with more than one parameter

panchenko:
@kcooney Sounds good, I will add more tests. 
Could you please clarify one thing on the style changes: the project uses google code style, the line length is 100 there, seems like lines were smaller already, however you still wrapped some, to make them even shorter.

kcooney:
@panchenko yes we allow lines up to 100 characters. Sorry about that, I was reviewing the GitHub diff view when I did the cleanup. The only minor issue is that we do not use `final` for fields or parameters unless forced to by the compiler (so `final` on a field is a hint it is used in an anonymous class)

src/test/java/org/junit/tests/running/classes/ParameterizedTestTest.java
panchenko: @kcooney I've added some more tests and validation for number of parameters of the BeforeParam/AfterParam methods.
Do you think it's OK already?

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -235,34 +238,104 @@
...
-                throws Throwable {
-            return new RunnersFactory(klass).createRunners();
-        }
+        private Iterable<Object> allParameters;
 
         private RunnersFactory(Class<?> klass) {
             testClass = new TestClass(klass);
         }
 
+        private void evaluateParameters() throws Throwable {
```
kcooney: This isn't thread-safe, which can be a problem if tests are run in parallel. Also, having this method have a side-effect of assigning `allParameters` is a bit odd.

If we really need to cache the result of calling `allParameters()` I would prefer that to happen inside of `allParameters()` itself. You could rename `allParameters()` to `callParametersMethod()` and then add this:

    private volatile Iterable<Object> allParameters;

    private Iterable<Object> allParameters() throws Throwable {
      if (allParameters == null) { // volatile read
        allParameters = callParametersMethod();
      }
      return allParameters;
    }

(and then change all the places where you reference `allParameters` to call `allParmeters()` like was done before) 

This won't guarantee that the the parameters method is only called once, but the only way to guarantee that would be 
to have a lock that is held while we call the parameters method.

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -235,34 +238,104 @@
...
+
+    private Parameterized(Class<?> klass, RunnersFactory runnersFactory) throws Throwable {
+        super(klass, runnersFactory.createRunners());
+        validateBeforeParamAndAfterParamMethods(runnersFactory.getParameterCount());
+    }
+
+    private void validateBeforeParamAndAfterParamMethods(Integer parameterCount)
+            throws InvalidTestClassError {
+        List<Throwable> errors = new ArrayList<Throwable>();
+        validatePublicStaticVoidMethods(Parameterized.BeforeParam.class, parameterCount, errors);
```
kcooney: Could we delete `validateBeforeParamAndAfterParamMethods()` and instead override `collectInitializationErrors(List<Throwable>)` and call `validatePublicStaticVoidMethods` there?

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -235,34 +238,104 @@
...
                     Parameters.class);
             return Collections.unmodifiableList(createRunnersForParameters(
-                    allParameters(), parameters.name(),
+                    allParameters, parameters.name(),
                     getParametersRunnerFactory()));
         }
 
+        private Integer getParameterCount() throws Throwable {
+            evaluateParameters();
+            Iterator<Object> iterator = allParameters.iterator();
```
kcooney: I don't think there's any guarantee that the `Iterable` can be iterated more than once. The current code only iterates over the collection once, so this could possibly break people.

I think we might need to wait until we are about to call a `BeforeParam` or `AfterParam` method before we check if the method has the right number of parameters.

src/main/java/org/junit/runners/Parameterized.java
panchenko: I tried that before, `collectInitializationErrors()` is called from the super class constructor and the number of method parameters can't be easily validated (only with some tricks, e.g. using `ThreadLocal`).

src/main/java/org/junit/runners/Parameterized.java
panchenko: Thread safety is not an issue here, as an instance of this inner class exists only during the `Parameterized` constructor invocation.

src/main/java/org/junit/runners/Parameterized.java
panchenko: Good point on iterating it multiple times. I would like to validate these methods once, otherwise an error would in each child runner (they are created for each parameter).

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -276,10 +330,15 @@ private ParametersRunnerFactory getParametersRunnerFactory()
...
 
+        private Integer parameterCount;
+
         private TestWithParameters createTestWithNotNormalizedParameters(
                 String pattern, int index, Object parametersOrSingleParameter) {
             Object[] parameters = (parametersOrSingleParameter instanceof Object[]) ? (Object[]) parametersOrSingleParameter
                     : new Object[] { parametersOrSingleParameter };
+            if (parameterCount == null) {
+                parameterCount = parameters.length;
+            }
```
panchenko: As a side effect we could have a check that number of parameters is always the same.

kcooney: If the number of parameters were different for some parameter sets, you would get a failure when the child runner is run.

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -276,10 +330,15 @@ private ParametersRunnerFactory getParametersRunnerFactory()
             }
         }
 
+        private Integer parameterCount;
```
kcooney: Having this set as a side-effect of calling `createRunners()` is fragile. It would be better to make this final and assign this at construction time. I added a commit that does ths.

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -278,25 +340,37 @@ private ParametersRunnerFactory getParametersRunnerFactory()
...
-            Object parameters = getParametersMethod().invokeExplosively(null);
-            if (parameters instanceof Iterable) {
-                return (Iterable<Object>) parameters;
+        private static List<Object> allParameters(
+                TestClass testClass, FrameworkMethod parametersMethod) throws Throwable {
+            Object parameters = parametersMethod.invokeExplosively(null);
+            if (parameters instanceof List) {
+                return (List<Object>) parameters;
+            } else if (parameters instanceof Collection) {
+                return new ArrayList<Object>((Collection<Object>) parameters);
```
panchenko: @kcooney IMHO there is no need to handle `Collection` specially, it can be treated as `Iterable`.

src/main/java/org/junit/runners/parameterized/BlockJUnit4ClassRunnerWithParameters.java
```diff
@@ -135,7 +138,73 @@ protected void validateFields(List<Throwable> errors) {
...
+        @Override
+        public void evaluate() throws Throwable {
+            for (FrameworkMethod before : befores) {
+                int paramCount = before.getMethod().getParameterTypes().length;
+                before.invokeExplosively(
+                        null, paramCount == 0 ? (Object[]) null : parameters);
+            }
+            next.evaluate();
+        }
+    }
```
marcphilipp: Can we reuse `org.junit.internal.runners.statements.RunBefores` and pass `parameters` to it?

src/main/java/org/junit/runners/parameterized/BlockJUnit4ClassRunnerWithParameters.java
```diff
@@ -135,7 +138,73 @@ protected void validateFields(List<Throwable> errors) {
...
+                        int paramCount = each.getMethod().getParameterTypes().length;
+                        each.invokeExplosively(
+                                null, paramCount == 0 ? (Object[]) null : parameters);
+                    } catch (Throwable e) {
+                        errors.add(e);
+                    }
+                }
+            }
+            MultipleFailureException.assertEmpty(errors);
+        }
```
marcphilipp: Can we reuse `org.junit.internal.runners.statements.RunAfters` and pass `parameters` to it?

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -137,6 +137,19 @@
  * }
  * </pre>
  *
+ * <h3>Executing code before/after executing tests for a specific parameter</h3>
+ * <p>
+ * If your test needs to perform some preparation or cleanup based on the parameters this can be
+ * done by adding public static methods annotated with &#064;BeforeParam/&#064;AfterParam,
+ * such methods should either have no parameters or have same parameters as the test.
```
marcphilipp: Please replace this paragraph with:

```
 * If your test needs to perform some preparation or cleanup based on the
 * parameters, this can be done by adding public static methods annotated with
 * {@code @BeforeParam}/{@code @AfterParam}. Such methods should either have no
 * parameters or the same parameters as the test.
```

src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -137,6 +137,19 @@
...
+ * <pre>
+ * &#064;BeforeParam
+ * public static void beforeTestsForParameter(String onlyParameter) {
+ *     System.out.println("Testing " + onlyParameter);
+ * }
+ * </pre>
+ * </p>
```
marcphilipp: Please remove the `</p>` line.

src/main/java/org/junit/internal/runners/statements/RunAfters.java
```diff
@@ -30,12 +30,16 @@ public void evaluate() throws Throwable {
...
+                    invokeMethod(each);
                 } catch (Throwable e) {
                     errors.add(e);
                 }
             }
         }
         MultipleFailureException.assertEmpty(errors);
     }
+
+    protected void invokeMethod(FrameworkMethod method) throws Throwable {
```
kcooney: Please add `@since 4.13`

src/main/java/org/junit/internal/runners/statements/RunBefores.java
```diff
@@ -21,8 +21,12 @@ public RunBefores(Statement next, List<FrameworkMethod> befores, Object target)
     @Override
     public void evaluate() throws Throwable {
         for (FrameworkMethod before : befores) {
-            before.invokeExplosively(target);
+            invokeMethod(before);
         }
         next.evaluate();
     }
+
+    protected void invokeMethod(FrameworkMethod method) throws Throwable {
```
kcooney: Please add `@since 4.13`

kcooney:
Thanks, @panchenko ! Could you please update the release notes at https://github.com/junit-team/junit4/wiki/4.13-Release-Notes ?

raffig:
I have one doubt: Currently @BeforeParam is called BEFORE parameters are known (before test instance constructor) thus annotated metod cannot do nothing that depends on parameter value. Of course one can do this in constructor, but I'm not sure if that is by the book. @AfterParam seems to work as expected.

raffig:
And the second doubt: @AfterParam method must be static. If this method is used to cleanup some resources passed as parameters (e.g. test container) it causes the resources must be kept in static field. Won't this break anything like e.g. parallel test execution?

Could you please take a closer look at this:
https://github.com/testcontainers/testcontainers-java/issues/1774

panchenko:
@raffig both must be public static and optionally can take parameters (however that's not clear from the documentation).

raffig:
@panchenko Thanks! Thats a great hint with parameters. I will fix my workaround and paste it in testcontainers thread. However I think maybe @ClassRule/TestRule could be fixed as well, so TestContainer could implement properly the container lifecycle in parametrized test?

panchenko:
@raffig `@ClassRule` works as expected - for the whole class

raffig:
@panchenko Ah, ok, so if ClassRule is for whole class, so ParamRule is needed right? Do you think it will get introduced in 4.13?

Thanks to your advice I think I've created suitable solution for TestContainers: https://github.com/testcontainers/testcontainers-java/issues/1774#issuecomment-524254875

Probably the same could be implemented with something like ParamRule.

panchenko:
@raffig It feels like quite a rare case when a `TestRule` instance is used as a test parameter.

raffig:
@panchenko I didn't think about TestRule as parameter. Rather about something that would allow  parameters lifecycle management more out-of-the box as in case of TestContainers. However the longer I'm thinking about this the closer I get to the conclusion that is not possible to do in any useful generic way. As for now I think the @BeforeParam/@AfterParam is the best option to solve this.

Everything started when I found out that ClassRule provided by TestContainers does not work in case of Parameterized tests. That's why I've been searching for solution in the beginning, which I found using new @BeforeParam/@AfterParam. Now I'm wondering if TestContainers could provide something like ClassRule, out-of-the-box, but working with Parameterized tests.

Issue #1432, opened 2017-03-18
Asserts for CharSequence contents


sabi0:
A lot of APIs in Android deal with `CharSequence`. And using `assertEquals()` with `CharSequence` arguments is cumbersome and unstable. One has to reproduce the exact same type which might change later.
And sometimes it just does not work (e.g. `StringBuilder` uses default `Object.equals()` implementation).

Therefore I would suggest to add new methods like
`assertContentEquals(CharSequence expected, CharSequence actual)`
that would compare just the content, irregardless of the type it's been wrapped into.

(Pull request #949 was trying to do that with generic `assertEquals()` which is indeed questionable)

I can create a pull request if you find this proposal reasonable.

kcooney:
Important issue, but I would prefer to do it in `assertEquals` or an overload of `assertEquals`. Having to know which assert message to use for which use case is error-prone.

You might want to consider using assertion frameworks with a fluent API (like Truth or Fest).

sabi0:
The problem with `assertEquals` is that sometimes you do need to compare the types too.
So I believe `assertEquals("abc", new StringBuilder("abc"))` must fail. Therefore the need for separate methods to compare just the content.
This is IMO very similar to the current `assertArrayEquals` approach.

I agree that picking the right method would become a bit trickier. But it's anyway better than the current lack of proper `CharSequence` support.

kcooney:
Could you list what specific methods you are proposing?

sabi0:
So far I've been thinking of just these two:
`Assert.assertContentEquals(CharSequence expected, CharSequence actual)`
`Assert.assertContentEquals(String message, CharSequence expected, CharSequence actual)`

kcooney:
Could you give better examples? I would never write `assertEquals("abc", new StringBuilder("abc"))` (I would call `build()` on the builder)

Also I do think DSL based frameworks are often a better fit. In Truth I would write:

    assertThat(myCharSequence).hasValue("hello");

sabi0:
Here is an example of how it might look in a test:
`assertContentEquals("abc", bundle.getCharSequence(Notification.EXTRA_TEXT));`

Adding also a `HasValue` `Matcher` is a good idea.

kcooney:
Thanks. Since `CharSequence` doesn't provide a contract for `equals()` how do you propose we consider two `CharSequence` values equal?  Is it simply that they have the same `toString()` representation? If so why not make the expected value a `String`?

sabi0:
[...]

PR #1438, opened 2017-04-12
Asserts for CharSequence contents #1432

joseph-mccarthy:
PR for #1432 ready for feedback

### DIFF
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -813,39 +813,34 @@ public class Assert {
     }
 
     /**
-     * Asserts that two {@link CharSequence} are the same. If they are not the same, an {@link AssertionError} is
-     * thrown.
+     * Asserts that two {@link CharSequence} instances have the same contentsa.
+     * If they do not, an {@link AssertionError} is thrown.
      *
-     * @param expected the object you expect
+     * @param expected the expected contents
      * @param actual the object to compare to
-     * @see CharSequence
-     * @see AssertionError
      */
-    public static void assertContentEquals(CharSequence expected, CharSequence actual){
-        assertContentEquals(null,expected,actual);
+    public static void assertContentEquals(CharSequence expected, CharSequence actual)  {
+        assertContentEquals(null, expected, actual);
     }
 
     /**
-     * Asserts that two {@link CharSequence} are the same. If they are not the same, an {@link AssertionError} is
-     * thrown with the given message.
+     *Asserts that two {@link CharSequence} instances have the same contentsa.
+     * If they do not, an {@link AssertionError} is thrown with the given message.
      *
      * @param message the identifying message for the {@link AssertionError}
-     * @param expected the object you expect
+     * @param expected the expected contents
      * @param actual the object to compare to
-     * @see CharSequence
-     * @see AssertionError
      */
-    public static void assertContentEquals(String message, CharSequence expected, CharSequence actual){
-        if((expected == null) && (actual == null)){
+    public static void assertContentEquals(String message, CharSequence expected, CharSequence actual) {
+        if (expected == null && actual == null) {
             return;
-        }else{
-            if(expected == null || actual == null){
-                failNotEquals(message,expected,actual);
-            }
+        }
+        if (expected == null || actual == null) {
+            failNotEquals(message, expected, actual);
         }
 
-        assertEquals(expected.length(),actual.length());
-        assertEquals(expected.toString(),actual.toString());
+        assertEquals(expected.length(), actual.length());
+        assertEquals(expected.toString(), actual.toString());
     }
 
     private static void failSame(String message) {
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -812,6 +812,42 @@ public class Assert {
         assertNotSame(null, unexpected, actual);
     }
 
+    /**
+     * Asserts that two {@link CharSequence} are the same. If they are not the same, an {@link AssertionError} is
+     * thrown.
+     *
+     * @param expected the object you expect
+     * @param actual the object to compare to
+     * @see CharSequence
+     * @see AssertionError
+     */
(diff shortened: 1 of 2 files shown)


src/main/java/org/junit/Assert.java
```diff
@@ -812,6 +812,37 @@ public static void assertNotSame(Object unexpected, Object actual) {
...
+    public static void assertContentEquals(String message, CharSequence expected, CharSequence actual) {
+        if (expected == null && actual == null) {
+            return;
+        }
+        if (expected == null || actual == null) {
+            failNotEquals(message, expected, actual);
+        }
+
+        assertEquals(expected.length(), actual.length());
+        assertEquals(expected.toString(), actual.toString());
```
kcooney: This should include the message, no?

src/main/java/org/junit/Assert.java
```diff
@@ -812,6 +812,37 @@ public static void assertNotSame(Object unexpected, Object actual) {
...
+     */
+    public static void assertContentEquals(String message, CharSequence expected, CharSequence actual) {
+        if (expected == null && actual == null) {
+            return;
+        }
+        if (expected == null || actual == null) {
+            failNotEquals(message, expected, actual);
+        }
+
+        assertEquals(expected.length(), actual.length());
```
kcooney: If this assertion fails, the error message will be confusing. Could you construct a message that mentions that the length is not equal, including the passed-in message like assertEquals(String, String, String) does?

src/main/java/org/junit/Assert.java
```diff
@@ -812,6 +812,37 @@ public static void assertNotSame(Object unexpected, Object actual) {
...
     }
 
+    /**
+     * Asserts that two {@link CharSequence} instances have the same contentsa.
+     * If they do not, an {@link AssertionError} is thrown.
+     *
+     * @param expected the expected contents
+     * @param actual the object to compare to
+     */
+    public static void assertContentEquals(CharSequence expected, CharSequence actual)  {
```
kcooney: I'm wondering if the expected value should be a String (the test author would usually pass in a literal string or a constant of type `Strnig`). What do you think?

src/test/java/org/junit/tests/assertion/AssertionTest.java
```diff
@@ -905,6 +906,41 @@ public void expectThrowsUsesCanonicalNameWhenRequiredExceptionNotThrown() {
...
+    }
+
+    @Test
+    public void assertContentEqualsPassBothNull() throws Exception {
+        CharSequence expected = null;
+        CharSequence charSequence = null;
+        assertContentEquals(expected,charSequence);
+    }
+
+    @Test(expected = AssertionError.class)
```
kcooney: Instead of using the "expected" attribute of `@Test` I think the tests should catch the exception and verify that the exception message looks right:

    try {
      assertContentEquals(expected, charSequence);
    } catch (AssertionError e) {
      // make assertions on the message

      return;
    }
    fail("Expected an AssertionError");

src/test/java/org/junit/tests/assertion/AssertionTest.java
```diff
@@ -905,6 +906,41 @@ public void expectThrowsUsesCanonicalNameWhenRequiredExceptionNotThrown() {
...
+    public void assertContentsSingleNull() {
+        CharSequence expected = "StringValue";
+        CharSequence charSequence = null;
+        assertContentEquals(expected, charSequence);
```
kcooney: Please add similar tests that pass in a message

joseph-mccarthy:
Thanks for the review feedback, will make the changes. :+1:

src/main/java/org/junit/Assert.java
joseph-mccarthy: Well I decided to use `CharSequence` as the expected as String implements this. So passing in a literal would be fine.

kcooney: I think I would prefer a String. It would make it more likely that the person didn't get the parameters out of order, and it makes it more clear that, inline with assertEquals(), we won't fail if the CharSequences have different types.

src/main/java/org/junit/Assert.java
joseph-mccarthy: Part of me is even thinking what the value of the length of the comparisons. Could we not just have the `.toString()`. Should the comparison of the lengths be removed?

kcooney: @sabi0 suggested checking the lengths first

joseph-mccarthy:
Closing Pull Request

src/main/java/org/junit/Assert.java
joseph-mccarthy: Addressed in PR #1439

src/main/java/org/junit/Assert.java
joseph-mccarthy: Addressed in PR #1439

src/main/java/org/junit/Assert.java
joseph-mccarthy: Addressed in PR #1439

src/test/java/org/junit/tests/assertion/AssertionTest.java
joseph-mccarthy: Addressed in PR #1439

src/test/java/org/junit/tests/assertion/AssertionTest.java
joseph-mccarthy: Addressed in PR #1439

src/main/java/org/junit/Assert.java
sabi0: Actually, I suggested comparing the lengths first only to avoid calling `toString()`.
But since `assertEquals(..., actual.toString())` is used for comparing the content there is no point to do a separate length check (`String.equals()` will do it anyway).

Besides, not checking the lengths (via `assertEquals`) would result in a nicer message. Compare:
> Expected: <5> but was: <2>

vs

> Expected: "Hello" but was: "Hi"

PR #1439, opened 2017-04-16
Issue #1432 Asserts for charsequence

joseph-mccarthy:
Closed Pull Request #1438 opened this one as requested

### DIFF
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -813,39 +813,34 @@ public class Assert {
     }
 
     /**
-     * Asserts that two {@link CharSequence} are the same. If they are not the same, an {@link AssertionError} is
-     * thrown.
+     * Asserts that two {@link CharSequence} instances have the same contentsa.
+     * If they do not, an {@link AssertionError} is thrown.
      *
-     * @param expected the object you expect
+     * @param expected the expected contents
      * @param actual the object to compare to
-     * @see CharSequence
-     * @see AssertionError
      */
-    public static void assertContentEquals(CharSequence expected, CharSequence actual){
-        assertContentEquals(null,expected,actual);
+    public static void assertContentEquals(CharSequence expected, CharSequence actual)  {
+        assertContentEquals(null, expected, actual);
     }
 
     /**
-     * Asserts that two {@link CharSequence} are the same. If they are not the same, an {@link AssertionError} is
-     * thrown with the given message.
+     *Asserts that two {@link CharSequence} instances have the same contentsa.
+     * If they do not, an {@link AssertionError} is thrown with the given message.
      *
      * @param message the identifying message for the {@link AssertionError}
-     * @param expected the object you expect
+     * @param expected the expected contents
      * @param actual the object to compare to
-     * @see CharSequence
-     * @see AssertionError
      */
-    public static void assertContentEquals(String message, CharSequence expected, CharSequence actual){
-        if((expected == null) && (actual == null)){
+    public static void assertContentEquals(String message, CharSequence expected, CharSequence actual) {
+        if (expected == null && actual == null) {
             return;
-        }else{
-            if(expected == null || actual == null){
-                failNotEquals(message,expected,actual);
-            }
+        }
+        if (expected == null || actual == null) {
+            failNotEquals(message, expected, actual);
         }
 
-        assertEquals(expected.length(),actual.length());
-        assertEquals(expected.toString(),actual.toString());
+        assertEquals(expected.length(), actual.length());
+        assertEquals(expected.toString(), actual.toString());
     }
 
     private static void failSame(String message) {
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -839,8 +839,8 @@ public class Assert {
             failNotEquals(message, expected, actual);
         }
 
-        assertEquals(expected.length(), actual.length());
-        assertEquals(expected, actual.toString());
+        assertEquals(message, expected.length(), actual.length());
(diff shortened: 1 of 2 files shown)


src/main/java/org/junit/Assert.java
```diff
@@ -812,6 +812,41 @@ public static void assertNotSame(Object unexpected, Object actual) {
...
+    /**
+     * Asserts that two {@link CharSequence} instances have the same contents.
+     * If they do not, an {@link AssertionError} is thrown with the given message.
+     *
+     * @param message the identifying message for the {@link AssertionError}
+     * @param expected the expected contents
+     * @param actual the object to compare to
+     */
+    public static void assertContentEquals(String message, String expected, CharSequence actual) {
+        if (expected == null && actual == null) {
```
kcooney: I think we should consider writing it like this:

    if (expected == actual) {
      return;
    }
    if (expected == null || actual == null) {
      failNotEquals(message, expected, actual
    }
    ...

That way, we avoid calling `actual.toString()` when `expected` and `actual` point to the same instance.

That being said, since the method is named "assertContentEquals" (and not "assertEquals") perhaps passing in `null` for `expected` should be an error (resulting in a `NullPointerException`). We don't do that anywhere else in this class, but I thought I'd put the idea out there in case someone might want to argue for it.

src/main/java/org/junit/Assert.java
```diff
@@ -812,6 +812,41 @@ public static void assertNotSame(Object unexpected, Object actual) {
...
+            return;
+        }
+        if (expected == null || actual == null) {
+            failNotEquals(message, expected, actual);
+        }
+
+        assertEquals(contentsDifferentMessage(message), expected.length(), actual.length());
```
kcooney: Moving the conversation about checking the length first to this pull request...

@sabi0 apparently suggested to check for the lengths to avoid the `toString()` call. I agree that failing early if the lengths differ with a message that just prints the lengths might not be all that helpful. We do, however, have one close precedent in JUnit, and that's when comparing arrays. If the lengths are different, we use a message like this:

`"<user supplied message>: array lengths differed, expected.length=2 actual.length=1; arrays first differed at element [1]; expected:<false> but was:<end of array>"`

(with no user-supplied message, the message starts with "array lengths...")

But I think `assertArrayEquals` does that because 1) arrays are often large, 2) the printed representation of arrays can be hard to read and 3) visually seeing the difference between two arrays in an error message is difficult. Here, we throw a `ComparisonFailure`, and the exception message generated by JUnit will show which parts of the string differ (plus IDEs often show a nice UI for `ComparisonFailure`). None of that applies here, so I think that even stating that the lengths are different in the exception isn't that useful for the method here.

src/main/java/org/junit/Assert.java
joseph-mccarthy: Should I make these changes, or are they still open for discussion from others?

src/main/java/org/junit/Assert.java
joseph-mccarthy: Should I make these changes, or are they still open for discussion from others?

src/main/java/org/junit/Assert.java
kcooney: As a potential user of this method, how should it handle null values?

(I assume short-circuiting the method if expected == actual is not controversial)

src/main/java/org/junit/Assert.java
kcooney: No one chimed in, so let's not mention the length difference in the exception message. I do think we should optimize for the expected == actual.

src/main/java/org/junit/Assert.java
```diff
@@ -812,6 +812,41 @@ public static void assertNotSame(Object unexpected, Object actual) {
...
+        assertEquals(message, expected, actual.toString());
+    }
+
+    private static String contentsDifferentMessage(String message) {
+        return message == null ? "Contents different length" : message + " contents different length";
```
ashleyfrieze: This could be simplified:

```java
return (message != null ? message + ": " : "") + "Content of different length";
```

src/main/java/org/junit/Assert.java
```diff
@@ -812,6 +812,41 @@ public static void assertNotSame(Object unexpected, Object actual) {
...
     }
 
+    /**
+     * Asserts that two {@link CharSequence} instances have the same contents.
+     * If they do not, an {@link AssertionError} is thrown.
+     *
+     * @param expected the expected contents
+     * @param actual the object to compare to
+     */
+    public static void assertContentEquals(String expected, CharSequence actual)  {
```
ashleyfrieze: Suggestion - why not make the `expected` a `CharSequence` too?

sabi0: See the original discussion: https://github.com/junit-team/junit4/issues/1432#issuecomment-287558614

sabi0: ... and also here: https://github.com/junit-team/junit4/pull/1438#discussion_r111419678

src/main/java/org/junit/Assert.java
IgorNosach25: Signature of methods is not correct. We must have opportunity to compare 2 instances of Char Sequence interface.

kcooney: We decided that the expected value should be a string. According to the CharSequence Javadoc "The result of comparing two objects that implement CharSequence is therefore, in general, undefined"

IgorNosach25: You are right. I was confused because the question is open. Please close topic if we don't need compare 2 CharSequence instances. How I can see actual open issues? Please answer me in private message) thanks

PR #1442, opened 2017-04-26
pom.xml modified to resolve lifecycle configuration error.

nikshinde1996:
While importing this project in different IDE's life cycle configuration error occurred. This PR resolves that error by adding pluginManagement tag in build section.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,6 +136,7 @@
                 </includes>
             </resource>
         </resources>
+        <pluginManagement>
         <plugins>
             <!--
             Both "org.apache" and "org.codehaus" are default providers of MOJO plugins
@@ -380,6 +381,7 @@
                 <version>2.7</version>
             </plugin>
         </plugins>
+        </pluginManagement>
     </build>
 
     <reporting>


Tibor17:
I don't think this is a problem with POM.
I think you use Eclipse and this IDE has problem to use Maven.
Unless you develop JUnit but your project is dependent on this library then it is really issue of the IDE.

marcphilipp:
`pluginManagement` is not the right place to declare these plugins.

Which IDEs did you have trouble with? It works fine with IntelliJ IDEA.

nikshinde1996:
Eclipse Oxygen. 
Is it right to configure IDE, not to issue life-cycle configuration error for plugins ?

Tibor17:
@nikshinde1996 
Try to use IntelliJ IDEA. Not because I use it but because IDEA detects configuration of Maven plugins in this project and configures project files upon POM. You should not have problems then.

kcooney:
I have no problems developing in Eclipse. I think I used the `.classpath` to setup my project. Please provide steps to reproduce. It may be a bug in Oxygen.

kcooney:
This is probably relevant:

http://stackoverflow.com/q/6352208/95725

This pull appears to implement the highest-rated solution, but many of the comments on that answer indicate that the "fix" will cause other problems.

@nikshinde1996 I suggest you try one of the other fixes and reach out to us as to which ones work

marcphilipp:
@nikshinde1996 Can you provide more information which Maven plugin caused the error in Eclipse?

nikshinde1996:
This error is issued when you try importing project as existing maven project. Following Maven problem is issued at line 202 of pom.xml
```
Plugin execution not covered by lifecycle configuration: com.google.code.maven-replacer-plugin:replacer:
 1.5.3:replace (execution: default, phase: process-sources)
```
Importing project as cloned git repository works good in eclipse IDE.

panchenko:
The change is not needed.
This "error" means m2eclipse does not know if those plugins are important for importing project into IDE or not. There is [documentation](http://www.eclipse.org/m2e/documentation/m2e-execution-not-covered.html) for it, as I understand it is possible to disable this message via Eclipse preferences.

marcphilipp:
I'm closing this pull request since it's an IDE issue.

PR #1443, opened 2017-04-27
Made ExpectedExceptio.isAnyExceptionExpected public.

jokaah:
Is there any harm in making this method public?

I have a base class for our test classes that has both an ExpectedException rule and a Verifier rule. The problem is that I don't want to run the Verifier if any exception is expected. So this is my current solution:

```
public abstract class TestBase {
    private Method isAnyExceptionExpected;

    public TestBase() throws Exception {
        this.isAnyExceptionExpected = ExpectedException.class.getDeclaredMethod("isAnyExceptionExpected");
        this.isAnyExceptionExpected.setAccessible(true);
    }

    @Rule
    public ExpectedException thrown = ExpectedException.none();

    @Rule
    public Verifier verifier = new Verifier() {
        @Override
        public void verify() {
            try {
                if (!(boolean) isAnyExceptionExpected.invoke(thrown)) {
                    verifyStuff();
                    verifyOtherStuff();
                    alsoVerifyThis();
                }
            } catch (IllegalAccessException | InvocationTargetException e) {
                System.err.print("Reflection exception: " + e.getMessage());  // ugh...
            }
        }
    };

    @Before
    public void beforeEach() throws Exception {
        doAwesomeSetup();
    }
}
```

### DIFF
--- a/src/main/java/org/junit/rules/ExpectedException.java
+++ b/src/main/java/org/junit/rules/ExpectedException.java
@@ -247,8 +247,9 @@ public class ExpectedException implements TestRule {
 
     /**
      * Check if any Exception is expected.
+     * @since 4.13
      */
-    public boolean isAnyExceptionExpected() {
+    public final boolean isAnyExceptionExpected() {
         return matcherBuilder.expectsThrowable();
     }
 
--- a/src/main/java/org/junit/rules/ExpectedException.java
+++ b/src/main/java/org/junit/rules/ExpectedException.java
@@ -245,6 +245,13 @@ public class ExpectedException implements TestRule {
         return this;
     }
 
+    /**
+     * Check if any Exception is expected.
+     */
+    public boolean isAnyExceptionExpected() {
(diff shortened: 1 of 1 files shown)


src/main/java/org/junit/rules/ExpectedException.java
```diff
@@ -245,6 +245,13 @@ public ExpectedException expectCause(Matcher<?> expectedCause) {
         return this;
     }
 
+    /**
+     * Check if any Exception is expected.
+     */
+    public boolean isAnyExceptionExpected() {
```
kcooney: Let's make this final

src/main/java/org/junit/rules/ExpectedException.java
```diff
@@ -245,6 +245,13 @@ public ExpectedException expectCause(Matcher<?> expectedCause) {
         return this;
     }
 
+    /**
+     * Check if any Exception is expected.
+     */
```
kcooney: Add `@since 4.13`

PR #1445, opened 2017-05-03
@Rule(order=N)

panchenko:
Yet another attempt to order `@Rule`s, based on the #1417 idea, issue #1221.

No breaking API changes, with a cost of using `ThreadLocal` :-)
both MethodRule & TestRule are supported.
Will add javadocs later, when we agree on the approach.

### DIFF
--- a/src/main/java/org/junit/ClassRule.java
+++ b/src/main/java/org/junit/ClassRule.java
@@ -85,4 +85,12 @@ import java.lang.annotation.Target;
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.FIELD, ElementType.METHOD})
 public @interface ClassRule {
+
+    /**
+     * Specifies the order in which rules are applied. The rules with a higher value are inner.
+     *
+     * @since 4.13
+     */
+    int order() default Rule.DEFAULT_ORDER;
+
 }
--- a/src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
+++ b/src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
@@ -3,6 +3,7 @@ package org.junit.runners;
 import static org.junit.internal.runners.rules.RuleMemberValidator.RULE_METHOD_VALIDATOR;
 import static org.junit.internal.runners.rules.RuleMemberValidator.RULE_VALIDATOR;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -22,12 +23,13 @@ import org.junit.internal.runners.statements.InvokeMethod;
 import org.junit.internal.runners.statements.RunAfters;
 import org.junit.internal.runners.statements.RunBefores;
 import org.junit.rules.MethodRule;
-import org.junit.rules.RunRules;
 import org.junit.rules.TestRule;
 import org.junit.runner.Description;
 import org.junit.runner.notification.RunNotifier;
+import org.junit.runners.model.FrameworkMember;
 import org.junit.runners.model.FrameworkMethod;
 import org.junit.runners.model.InitializationError;
+import org.junit.runners.model.MemberValueConsumer;
 import org.junit.runners.model.MultipleFailureException;
 import org.junit.runners.model.Statement;
 import org.junit.runners.model.TestClass;
@@ -391,29 +393,23 @@ public class BlockJUnit4ClassRunner extends ParentRunner<FrameworkMethod> {
                 target);
     }
 
-    private Statement withRules(FrameworkMethod method, Object target,
-            Statement statement) {
-        List<TestRule> testRules = getTestRules(target);
-        Statement result = statement;
-        result = withMethodRules(method, testRules, target, result);
-        result = withTestRules(method, testRules, result);
-
-        return result;
-    }
-
-    private Statement withMethodRules(FrameworkMethod method, List<TestRule> testRules,
-            Object target, Statement result) {
-        Statement withMethodRules = result;
-        for (org.junit.rules.MethodRule each : getMethodRules(target)) {
-            if (!(each instanceof TestRule && testRules.contains(each))) {
-                withMethodRules = each.apply(withMethodRules, method, target);
+    private Statement withRules(FrameworkMethod method, Object target, Statement statement) {
+        RuleContainer ruleContainer = new RuleContainer();
+        CURRENT_RULE_CONTAINER.set(ruleContainer);
+        try {
+            List<TestRule> testRules = getTestRules(target);
+            for (MethodRule each : rules(target)) {
+                if (!(each instanceof TestRule && testRules.contains(each))) {
+                    ruleContainer.add(each);
+                }
             }
+            for (TestRule rule : testRules) {
+                ruleContainer.add(rule);
+            }
+        } finally {
+            CURRENT_RULE_CONTAINER.remove();
         }
-        return withMethodRules;
-    }
-
-    private List<org.junit.rules.MethodRule> getMethodRules(Object target) {
-        return rules(target);
+        return ruleContainer.apply(method, describeChild(method), target, statement);
     }
 
     /**
@@ -422,27 +418,12 @@ public class BlockJUnit4ClassRunner extends ParentRunner<FrameworkMethod> {
      *         test
      */
     protected List<MethodRule> rules(Object target) {
-        List<MethodRule> rules = getTestClass().getAnnotatedMethodValues(target, 
-                Rule.class, MethodRule.class);
-
-        rules.addAll(getTestClass().getAnnotatedFieldValues(target,
-                Rule.class, MethodRule.class));
-
-        return rules;
-    }
-
-    /**
-     * Returns a {@link Statement}: apply all non-static fields
-     * annotated with {@link Rule}.
-     *
-     * @param statement The base statement
-     * @return a RunRules statement if any class-level {@link Rule}s are
-     *         found, or the base statement
-     */
-    private Statement withTestRules(FrameworkMethod method, List<TestRule> testRules,
-            Statement statement) {
-        return testRules.isEmpty() ? statement :
-                new RunRules(statement, testRules, describeChild(method));
+        RuleCollector<MethodRule> collector = new RuleCollector<MethodRule>();
+        getTestClass().collectAnnotatedMethodValues(target, Rule.class, MethodRule.class,
+                collector);
+        getTestClass().collectAnnotatedFieldValues(target, Rule.class, MethodRule.class,
+                collector);
+        return collector.result;
     }
 
     /**
@@ -451,13 +432,10 @@ public class BlockJUnit4ClassRunner extends ParentRunner<FrameworkMethod> {
      *         test
      */
     protected List<TestRule> getTestRules(Object target) {
-        List<TestRule> result = getTestClass().getAnnotatedMethodValues(target,
-                Rule.class, TestRule.class);
-
-        result.addAll(getTestClass().getAnnotatedFieldValues(target,
-                Rule.class, TestRule.class));
-
-        return result;
+        RuleCollector<TestRule> collector = new RuleCollector<TestRule>();
+        getTestClass().collectAnnotatedMethodValues(target, Rule.class, TestRule.class, collector);
+        getTestClass().collectAnnotatedFieldValues(target, Rule.class, TestRule.class, collector);
+        return collector.result;
     }
 
     private Class<? extends Throwable> getExpectedException(Test annotation) {
@@ -473,5 +451,23 @@ public class BlockJUnit4ClassRunner extends ParentRunner<FrameworkMethod> {
             return 0;
         }
         return annotation.timeout();
+    }
+
+    private static final ThreadLocal<RuleContainer> CURRENT_RULE_CONTAINER =
+            new ThreadLocal<RuleContainer>();
+
+    private static class RuleCollector<T> implements MemberValueConsumer<T> {
+        final List<T> result = new ArrayList<T>();
+
+        public void accept(FrameworkMember member, T value) {
+            Rule rule = member.getAnnotation(Rule.class);
+            if (rule != null) {
+                RuleContainer container = CURRENT_RULE_CONTAINER.get();
+                if (container != null) {
+                    container.setOrder(value, rule.order());
+                }
+            }
+            result.add(value);
+        }
     }
 }
--- a/src/main/java/org/junit/runners/ParentRunner.java
+++ b/src/main/java/org/junit/runners/ParentRunner.java
@@ -36,9 +36,11 @@ import org.junit.runner.manipulation.Sortable;
 import org.junit.runner.manipulation.Sorter;
 import org.junit.runner.notification.RunNotifier;
 import org.junit.runner.notification.StoppedByUserException;
+import org.junit.runners.model.FrameworkMember;
 import org.junit.runners.model.FrameworkMethod;
 import org.junit.runners.model.InitializationError;
 import org.junit.runners.model.InvalidTestClassError;
+import org.junit.runners.model.MemberValueConsumer;
 import org.junit.runners.model.RunnerScheduler;
 import org.junit.runners.model.Statement;
 import org.junit.runners.model.TestClass;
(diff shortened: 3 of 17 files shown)


src/main/java/org/junit/internal/runners/rules/RuleContainer.java
```diff
@@ -0,0 +1,97 @@
...
+    private abstract static class RuleEntry<T> {
+        final T rule;
+        final int order;
+
+        RuleEntry(T rule, Integer order) {
+            this.rule = rule;
+            this.order = order != null ? order.intValue() : -1;
+        }
+
+        public int getInterfacePriority() {
```
kcooney: I suggest just making this method abstract

src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
```diff
@@ -380,29 +382,32 @@ protected Statement withAfters(FrameworkMethod method, Object target,
...
+    private Statement withRules(FrameworkMethod method, Object target, Statement statement) {
+        RuleContainer ruleContainer = new RuleContainer() {
+            protected Integer getOrder(FrameworkMember member) {
+                Rule rule = member.getAnnotation(Rule.class);
+                return rule != null ? rule.order() : null;
+            }
+        };
+        List<TestRule> testRules;
+        List<MethodRule> methodRules;
+        AnnotatedValueListener.CURRENT.set(ruleContainer);
```
kcooney: We would probably want to do something similar in `ParentRunner.withClassRules()`

panchenko:
@kcooney Thanks for the feedback, I think I've adjusted most of the concerns, could you please give another look? It terms of code flow it should be better and we could possibly discuss how it should be named and in what packages the new classes should go :-)

src/main/java/org/junit/runners/model/TestClass.java
```diff
@@ -255,14 +284,13 @@ public String getName() {
                  */
                 if (valueClass.isAssignableFrom(each.getReturnType())) {
                     Object fieldValue = each.invokeExplosively(test);
-                    results.add(valueClass.cast(fieldValue));
+                    consumer.accept(each, valueClass.cast(fieldValue));
```
kcooney: One thing I noticed in my attempt is that we will call `MethodRule` methods twice if the return value implements both `TestRule` and `MethodRule` (one of the values would be thrown away in `withRules()`. My `MethodValueCollector` class avoided this, because the collector subclass can filter out the values before the method value was called.

I'm not sure how much we should worry about that. Your approach is quite nice.

src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
```diff
@@ -380,29 +382,22 @@ protected Statement withAfters(FrameworkMethod method, Object target,
...
-        Statement withMethodRules = result;
-        for (org.junit.rules.MethodRule each : getMethodRules(target)) {
-            if (!(each instanceof TestRule && testRules.contains(each))) {
-                withMethodRules = each.apply(withMethodRules, method, target);
+    private Statement withRules(FrameworkMethod method, Object target, Statement statement) {
+        RuleContainer ruleContainer = new RuleContainer();
+        RuleContainer.CURRENT.set(ruleContainer);
+        try {
+            List<TestRule> testRules = getTestRules(target);
+            List<MethodRule> methodRules = rules(target);
```
kcooney: Since `methodRules` is used once, I suggest inlining it:

    for (MethodRule each : rules(target)) {
      ...

(no idea why MethodRule was fully qualified here)

src/main/java/org/junit/runners/RuleContainer.java
```diff
@@ -0,0 +1,136 @@
...
+import org.junit.rules.MethodRule;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.FrameworkMethod;
+import org.junit.runners.model.Statement;
+
+/**
+ * @since 4.13
+ */
+public final class RuleContainer {
```
kcooney: Let's make this package-scope.

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -460,4 +467,16 @@ public int compare(T o1, T o2) {
...
         this.scheduler = scheduler;
     }
+
+    private static class ClassRuleCollector implements MemberValueConsumer<TestRule> {
+        final List<TestRule> result = new ArrayList<TestRule>();
+
+        public void accept(FrameworkMember member, TestRule value) {
+            ClassRule rule = member.getAnnotation(ClassRule.class);
+            if (rule != null) {
+                RuleContainer.recordRuleOrder(value, rule.order());
```
kcooney: It's not obvious here that we are modifying a thread local. I would prefer that `recordRuleOrder()` be a non-static method.

You can use [`ThreadLocal.inititalValue()`](https://docs.oracle.com/javase/7/docs/api/java/lang/ThreadLocal.html#initialValue()) to initialize the thread local to a no-op value

src/main/java/org/junit/runners/RuleContainer.java
```diff
@@ -0,0 +1,136 @@
...
+
+    public void add(MethodRule methodRule) {
+        methodRules.add(methodRule);
+    }
+
+    public void add(TestRule testRule) {
+        testRules.add(testRule);
+    }
+
+    public void addTestRules(List<? extends TestRule> rules) {
```
kcooney: nit: let's either name this `addAll()` or rename the `add()` methods to `addTestRule()` and `addMethodRule()`

Or just get rid of it. It's trivial to inline, and you don't have `addMethodRules()`

src/main/java/org/junit/runners/RuleContainer.java
```diff
@@ -0,0 +1,136 @@
...
+        private int compareInt(int a, int b) {
+            return (a < b) ? 1 : (a == b ? 0 : -1);
+        }
+    };
+
+    /**
+     * Returns entries in the order how they should be applied, i.e. inner-to-outer.
+     */
+    private List<RuleEntry> getSortedEntries() {
+        List<RuleEntry> ruleEntries = new ArrayList<RuleEntry>();
```
kcooney: Nit: `new ArrayList<RuleEntry>(methodRules.size() + testRules.size())`

src/main/java/org/junit/runners/model/MemberValueConsumer.java
```diff
@@ -0,0 +1,8 @@
+package org.junit.runners.model;
+
+/**
+ * @since 4.13
+ */
+public interface MemberValueConsumer<T> {
```
kcooney: I like the name and the packaging. Could you add a short description of what it does in the class Javadoc, and add method Javadoc to every new public method that is in a public class?

src/test/java/org/junit/runners/RuleContainerTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.runners.model.Statement;
+
+public class RuleContainerTest {
+    private final RuleContainer container = new RuleContainer();
+
+    @Test
+    public void methodRulesOnly() {
+        container.add(MRule.M1);
+        container.add(MRule.M2);
+        assertEquals("[M1, M2]", container.getSortedRules().toString());
```
kcooney: That's clever making the rules an enum and using `toString()`.

src/main/java/org/junit/ClassRule.java
```diff
@@ -85,4 +85,12 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.FIELD, ElementType.METHOD})
 public @interface ClassRule {
+
+    /**
+     * Specifies the order in which rules are applied. The rules with a higher value are inner.
+     *
+     * @since 4.13
+     */
+    int order() default -1;
```
kcooney: I personally like `priority` best. I can live with `order`. I don't like `level`

@junit-team/junit-committers what are your preferences?

src/main/java/org/junit/runners/ParentRunner.java
kcooney: Actually, instead of doing all the work to create a no-op version of `RuleCollector`, just do the `null` check of the thread-local here.

kcooney:
If we merge this, should we deprecate `RuleChain`?

src/main/java/org/junit/runners/ParentRunner.java
panchenko: It's used twice - for `@Rule` and `@ClassRule`, are we OK with the small code duplication?

src/main/java/org/junit/runners/RuleContainer.java
panchenko: It's used twice, so I've renamed the `add()` methods.

panchenko:
> If we merge this, should we deprecate RuleChain?

I was also thinking of this.

src/main/java/org/junit/runners/RuleContainer.java
panchenko: The only question is how likely someone might want to reuse it? The only part which should stay "hidden" here is the `ThreadLocal` field, the rest of the class can be moved into the `internal` package.

panchenko:
I've made some more changes here, particularly ordering of `@ClassRule` happens in a more direct way, no ThreadLocal is needed there, so it's moved into `BlockJUnit4ClassRunner`.

kcooney:
@panchenko thanks. I've busy with my primary job, so I might not have time to look into this until the weekend.

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -243,14 +244,7 @@ protected Statement withAfterClasses(Statement statement) {
...
     private Statement withClassRules(Statement statement) {
-        RuleContainer ruleContainer = new RuleContainer();
-        RuleContainerHelper.CURRENT.set(ruleContainer);
-        try {
-            ruleContainer.addTestRules(classRules());
-        } finally {
-            RuleContainerHelper.CURRENT.remove();
-        }
-        return ruleContainer.apply(null, getDescription(), null, statement);
+        return RunRules.applyAll(statement, classRules(), getDescription());
```
kcooney: We should not be applying rules in this method. Instead, we should return a `Statement` which applies the rules, so the rules are applied when the final statement is evaluated. That's why `RunRules` is a `Statement`.

panchenko:
@kcooney currently RunRules applies all the statements in the **constructor**, so the only difference between
* `new RunRules(statement, rules)`
* `RunRules.applyAll(statement, rules)`

is an additional `RunRules` instance with the `evaluate`method like this
```java
    statement.evaluate();
}
```
So as an optimization I've removed this instance and make use of the constructed `statement` directly. I agree this change is not strictly necessary, and at the same time I can't understand why an extra object is needed here :-)

kcooney:
@panchenko ah didn't realize `RunRules` did the work in the constructor. Still, I  prefer not to change it because 1) the change is unrelated to this pull 2) we should avoid depreciating things unless there is a clear reason for the current users to change, 3) the overhead of creating an object is low and 4) maybe we should apply the rules at statement execution time.

panchenko:
@kcooney reverted that change, will suggest it is separately :-)
"4)" would be a behavior change, I don't think we should do it.

kcooney:
This LGTM. Thanks!

The only open issue in my mind is the naming of the attribute. Order might be confusing, since if you ordered two `ExternalResource` rules, I would expect the `before()` method of the rule with the lower order to happen first, but it happens the other way around. But for `Verifier` rules, I think I would expect the opposite. I think `priority` makes it clear that the attribute affects which one is applied first, but forces the user to read the Javadoc to get clarity (though we might want to provide an example of what it means to be "inner", since if you use rules but don't write them you might not understand)

panchenko:
I agree that naming is always an issue.

Currently the rules with lower values are _executed_ first, so for some kind of rules it's already intuitively clear, from the test:
```java
@Rule(order = 1)
public final TestRule a = new LoggingTestRule(ruleLog, "outer");
@Rule(order = 2)
public final TestRule z = new LoggingTestRule(ruleLog, "inner");
....
assertEquals(" outer.begin inner.begin foo inner.end outer.end", ruleLog.toString());
```

`priority` might cause more confusion, e.g. in Windows higher value means higher priority, so one might assume it would be executed the other way around. Reading documentation is a good thing, but sometimes a question to GitHub/StackOverflow/forum is posted before doing that :-)

I would suggest optimizing for the common use case, which I think is about creating resources depending on each other. That's why `order` and sorting this way.

src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
```diff
@@ -463,4 +441,22 @@ private long getTimeout(Test annotation) {
         }
         return annotation.timeout();
     }
+
+    private static final ThreadLocal<RuleContainer> CURRENT_RULE_CONTAINER =
+            new ThreadLocal<RuleContainer>();
```
marcphilipp: Why does this have to be a `ThreadLocal`? Can't we use an instance variable and pass it to the `RuleCollector` constructor?

src/main/java/org/junit/runners/RuleContainer.java
```diff
@@ -0,0 +1,140 @@
...
+        if (methodRules.isEmpty() && testRules.isEmpty()) {
+            return statement;
+        }
+        Statement result = statement;
+        for (RuleEntry ruleEntry : getSortedEntries()) {
+            if (ruleEntry instanceof TestRuleEntry) {
+                result = ((TestRuleEntry) ruleEntry).rule.apply(result, description);
+            } else {
+                result = ((MethodRuleEntry) ruleEntry).rule.apply(result, method, target);
+            }
```
marcphilipp: The `if` can be replaced my introducing an `apply(FrameworkMethod method, Description description, Object target, Statement statement)` method in `RuleEntry`.

src/main/java/org/junit/runners/RuleContainer.java
```diff
@@ -0,0 +1,140 @@
...
+
+        RuleEntry(T rule, Integer order) {
+            this.rule = rule;
+            this.order = order != null ? order.intValue() : -1;
+        }
+
+        RuleEntry(T rule, int order) {
+            this.rule = rule;
+            this.order = order;
+        }
```
marcphilipp: Do we really need both constructors (same below)?

src/main/java/org/junit/ClassRule.java
marcphilipp: I'm fine with `order`.

src/main/java/org/junit/runners/BlockJUnit4ClassRunner.java
panchenko: This complexity comes from different methods returning `MethodRule` and `TestRule` instances - both are merged and ordered together.

src/main/java/org/junit/runners/RuleContainer.java
panchenko: It is possible, however I am not sure that would improve readability. The number of lines would be increased for sure.

src/main/java/org/junit/runners/RuleContainer.java
panchenko: Changed this.

src/main/java/org/junit/runners/RuleContainer.java
marcphilipp: Well, if you don't want to change this, then please remove `TestRuleEntry` and `MethodRuleEntry` and do the `instanceof` check on the contained rule class directly.

marcphilipp:
@kcooney Are you fine with `order` now? PTAL!

panchenko:
@marcphilipp What's your opinion on deprecating `RuleChain`?

marcphilipp:
FWIW I think we should deprecate RuleChain.

kcooney:
Changes look great!

I'd like to merge this as two commits. The second commit would have the `RuleChain` changes and the first would have anything else. @panchenko are you comfortable enough with git to do this? If not, I can do it, but that would mess up the history in your repo.

panchenko:
@kcooney Sure, I've changed this to 2 commits.

kcooney:
@panchenko would it be possible to put a few more details in the first commit message? No rush; I won't be spending time on JUnit this weekend.

Thanks for all of the work recently!

kcooney:
@marcphilipp merged. Would you please update the release notes on the wki?

panchenko:
@kcooney I've added this to release notes.

PR #1449, opened 2017-05-05
Parameterized reuses TestClass

kcooney:


kcooney:
@stefanbirkner could you review this?

stefanbirkner:
I have a look.

src/main/java/org/junit/runners/Suite.java
```diff
@@ -14,6 +14,7 @@
 import org.junit.runner.notification.RunNotifier;
 import org.junit.runners.model.InitializationError;
 import org.junit.runners.model.RunnerBuilder;
+import org.junit.runners.model.TestClass;
```
panchenko: This import seems not needed.

src/main/java/org/junit/internal/Checks.java
```diff
@@ -0,0 +1,37 @@
...
+     * as the exception message if an exception is thrown.
+     *
+     * @param value object reference to check
+     * @param message message to use if {@code value} is {@code null}
+     * @return the passed-in value, if not {@code null}
+     * @throws NullPointerException if {@code value} is {@code null}
+     */
+    public static <T> T notNull(T value, String message) {
+        if (value == null) {
+            throw new NullPointerException(String.valueOf(message));
```
panchenko: I would suggest passing `message` directly without `String.valueOf()`.

kcooney: `message` can  be `null`

panchenko: In this case this method would behave exactly as the one above, similarly to `java.util.Objects.requireNonNull()`

src/main/java/org/junit/internal/builders/JUnit4Builder.java
```diff
@@ -1,12 +1,12 @@
...
 
 public class JUnit4Builder extends RunnerBuilder {
     @Override
     public Runner runnerForClass(Class<?> testClass) throws Throwable {
-        return new BlockJUnit4ClassRunner(testClass);
+        return new JUnit4(testClass);
     }
-}
\ No newline at end of file
+}
```
stefanbirkner: This change is unrelated to the reuse of TestClass. It should be a separate commit.

src/main/java/org/junit/internal/Checks.java
```diff
@@ -0,0 +1,37 @@
+package org.junit.internal;
+
+/** @since 4.13 */
+public final class Checks {
```
stefanbirkner: I think we should not extract a Checks class and simply inline the validations, in order to not provide methods someone may use and depend on. I know that is a very weak argument. I leave it up to you whether you want to keep it or not.

src/main/java/org/junit/internal/Checks.java
stefanbirkner: I conform to @panchenko's suggestion. I normally expect that `someMethod(someValue)` and `someMethod(someValue, null)` behave the same.

src/main/java/org/junit/runners/Suite.java
kcooney: Removed

src/main/java/org/junit/internal/Checks.java
kcooney: The behavior I have here matches Guava's `Preconditions.checkNotNull(T, String)` method. I think the Guava team has given this much more thought than I have, but I can take a stab at explaining why the behavior I have is better.

IMHO it is a programmer error to pass in a `null` as the second parameter. If we have this method behave like `notNull(T)` when you pass in a `null` message we would be ignoring that programmer error. Having the message be `"null"` makes it clear that 1) someone tried to specify a real message, and 2) they messed up. Hopefully someone would notice and fix it.

src/main/java/org/junit/internal/builders/JUnit4Builder.java
kcooney: It is a separate commit. See https://github.com/junit-team/junit4/pull/1449/commits/11822c4fd1d2388079ed6a2eadd61ed7b3d44035

I have this commit in this pull since there's no reason for `ParentRunner.createTestClass()` to be called for JUnit4-style tests that aren't annotated with `RunWith`.

src/main/java/org/junit/internal/Checks.java
kcooney: As you can see in the commits, we have multiple classes that do `null` checks. Quite often when I write code in JUnit I find myself wanting a one-liner to check for null.

I doubt anyone would depend on this class, since 1) it's in an internal package, 2) it has an odd name, and 3) most everyone that is using JDK 6 or later would use `Objects.requireNotNull()` or Guava's `Preconditions.checkNotNull()`.

src/main/java/org/junit/internal/Checks.java
panchenko: Yes, please, lets keep it simple. Also the 1.8 version with `Supplier<String>` passes the result directly to the NullPointerException constructor.

src/main/java/org/junit/internal/Checks.java
stefanbirkner: I would like to go with the JDK way.

src/main/java/org/junit/internal/builders/JUnit4Builder.java
stefanbirkner: Sorry, my fault. I was not aware that the PR is made up of multible commits.

src/main/java/org/junit/internal/Checks.java
kcooney: Changed `notNull(T, String)` to not use `String.valueOf()` on the message.

tasomaniac:
How does this work when we provide the parameters through constructor? In such cases, the properties in test instance are `private final`

panchenko:
@tasomaniac This issue is about a _runner_ for parameterizzed tests - the tests are developed the same way as before.

tasomaniac:
I see. The title says reuses TestClass. I thought TestClass is the instance of the test class we create.

PR #1451, opened 2017-05-13
Support suppressed exceptions when trimming stack trace

### LINKED ISSUE #1426: Trimmed stack traces do not include suppressed throwables
Sometimes suppressed exceptions contain critical context.  Trimmed stack traces should include suppressed exceptions if any are present.

The workaround is to disable trimmed stack traces.

Ref: getTrimmedStackTrace in org.junit.internal.Throwables

panchenko:
Fixes #1426

### DIFF
--- a/src/main/java/org/junit/internal/Throwables.java
+++ b/src/main/java/org/junit/internal/Throwables.java
@@ -5,6 +5,7 @@ import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.StringReader;
 import java.io.StringWriter;
+import java.lang.reflect.Method;
 import java.util.AbstractList;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


src/main/java/org/junit/internal/Throwables.java
```diff
@@ -104,8 +105,31 @@ public static String getTrimmedStackTrace(Throwable exception) {
...
+    private static Method initGetSuppressed() {
+        try {
+            return Throwable.class.getMethod("getSuppressed");
+        } catch (Throwable e) {
+            return null;
+        }
+    }
+
+    private static boolean hasSuppressed(Throwable exception) {
+        if (getSuppressed != null) {
```
kcooney: nit: I personally prefer existing out on error conditions early, to avoid over-intentation:

    if (getSuppressed == null) {
      return false;
    }
    try {
      ...
    } catch (Throwable e) {
      return false;
    }

PR #1452, opened 2017-05-14
Test for failures in ExternalResouce.after() when multiple rules

panchenko:
When constructing MultipleFailureException and some of the exceptions are already MultipleFailureException they are unwrapped.

This is an improvement which came to my mind while doing a test for 2 `ExternalResource` instances failing in `after()` like mentioned in #1334.

### DIFF
--- a/src/main/java/org/junit/runners/model/MultipleFailureException.java
+++ b/src/main/java/org/junit/runners/model/MultipleFailureException.java
@@ -33,13 +33,9 @@ public class MultipleFailureException extends Exception {
         this.fErrors = new ArrayList<Throwable>(errors.size());
         for (Throwable error : errors) {
             if (error instanceof AssumptionViolatedException) {
-                fErrors.add(
-                        new TestCouldNotBeSkippedException((AssumptionViolatedException) error));
-            } else if (error instanceof MultipleFailureException) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/junit/runners/model/MultipleFailureException.java
```diff
@@ -33,9 +33,13 @@ public MultipleFailureException(List<Throwable> errors) {
         this.fErrors = new ArrayList<Throwable>(errors.size());
         for (Throwable error : errors) {
             if (error instanceof AssumptionViolatedException) {
-                error = new TestCouldNotBeSkippedException((AssumptionViolatedException) error);
+                fErrors.add(
+                        new TestCouldNotBeSkippedException((AssumptionViolatedException) error));
+            } else if (error instanceof MultipleFailureException) {
```
kcooney: Any thrown `MultipleFailureException` will be unwrapped, recursively, in `EachTestNotifier.addFailure()`. Are there any benefits to doing it here as well?

panchenko:
@kcooney You are absolutely right.
This started as an additional test demonstrating that #1334 is fully fixed and I've implemented it in a different way.

kcooney:
Thanks! Pushed some minor style fixes. Will wait until the Travis build passes, then will squash and merge.

PR #1454, opened 2017-05-18
fix assertEquals() if toString() returns null

### LINKED ISSUE #1453: NPE when asserting classes that return null in toString
```
    public static class NotString
    {

        private Object data;

        public NotString(Object data)
        {
            this.data = data;
        }

        @Override
        public boolean equals(Object obj)
        {
            if (this == obj)
                return true;
            if (obj == null)
                return false;
            if (getClass() != obj.getClass())
                return false;
            NotString other = (NotString) obj;
            if (this.data == null)
            {
                if (other.data != null)
                    return false;
            }
            else if (!this.data.equals(other.data))
                return false;
            return true;
        }

        @Override
        public String toString()
        {
            return null;
        }

    }

    @Test
    public void testNotStringClass()
    {
        assertEquals(new NotString(1), new NotString(2));
    }
```

This code results in a NPE from JUnit because the ToString method of the class is returning null.

Tested in JUnit 4.12

panchenko:
Fixes #1453

### DIFF
--- a/src/test/java/org/junit/tests/assertion/AssertionTest.java
+++ b/src/test/java/org/junit/tests/assertion/AssertionTest.java
@@ -728,12 +728,14 @@ public class AssertionTest {
     public void nullToString() {
         try {
             assertEquals(new NullToString(), new NullToString());
-            fail();
         } catch (AssertionError e) {
             assertEquals("expected: org.junit.tests.assertion.AssertionTest$NullToString<null> but "
                             + "was: org.junit.tests.assertion.AssertionTest$NullToString<null>",
                     e.getMessage());
+            return;
         }
+
+        fail("Failed on assertion.");
     }
 
     @Test(expected = AssertionError.class)
--- a/src/test/java/org/junit/tests/assertion/AssertionTest.java
+++ b/src/test/java/org/junit/tests/assertion/AssertionTest.java
@@ -717,6 +717,25 @@ public class AssertionTest {
         assertEquals("null", null);
     }
 
+    private static class NullToString {
+        @Override
+        public String toString() {
+            return null;
+        }
+    }
+
+    @Test
+    public void nullToString() {
+        try {
+            assertEquals(new NullToString(), new NullToString());
+            fail();
+        } catch (AssertionError e) {
+            assertEquals("expected: org.junit.tests.assertion.AssertionTest$NullToString<null> but "
+                            + "was: org.junit.tests.assertion.AssertionTest$NullToString<null>",
+                    e.getMessage());
+        }
+    }
+
     @Test(expected = AssertionError.class)
     public void compareBigDecimalAndInteger() {
         final BigDecimal bigDecimal = new BigDecimal("1.2");
--- a/src/test/java/org/junit/tests/assertion/AssertionTest.java
+++ b/src/test/java/org/junit/tests/assertion/AssertionTest.java
@@ -724,18 +724,19 @@ public class AssertionTest {
         }
     }
 
+    private static class NothingWasThrown extends RuntimeException {
+    }
+
     @Test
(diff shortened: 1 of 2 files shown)


src/test/java/org/junit/tests/assertion/AssertionTest.java
```diff
@@ -717,6 +717,27 @@ public void stringNullAndNullWorksToo() {
...
+        try {
+            assertEquals(new NullToString(), new NullToString());
+        } catch (AssertionError e) {
+            assertEquals("expected: org.junit.tests.assertion.AssertionTest$NullToString<null> but "
+                            + "was: org.junit.tests.assertion.AssertionTest$NullToString<null>",
+                    e.getMessage());
+            return;
+        }
+
+        fail("Failed on assertion.");
```
marcphilipp: Why do we need this `fail` statement?

panchenko: Only in case `assertEquals()` does nothing.

marcphilipp: Well, the same holds true for all the other tests in this class, right? If so, please remove the `return` and the `fail()` statements.

panchenko: @marcphilipp there are different kind of tests in this class:
* some have `@Test(expected = AssertionError.class)`
* some have `return` + `fail()`
* some have only `fail()` inside `try-catch`
* some don't have `fail()` / `return`

As this test is expected to throw an exception and that exception is being checked, I think it's better to make sure an exception was thrown. I agree this variant is too verbose, the best approach would be throwing some new runtime exception from the `try-catch` block.

kcooney: Calling `fail` from inside of a `try` block that has `catch (AssertionError` doesn't do what many people think it does, so I'm strongly against that option. Maybe I'll find time to fix the tests that do that.

I also dislike using  `@Test(expected = Something.class)` for any test method where there is more than one call in the body, since it isn't clear which method should throw the exception. Because using `expected` should usually only be used for single-method tests, I tend to completely avoid it in my personal projects, but it's usage is very common in the JUnit tests.

Throwing a custom exception inside the `try` block for something that should be an assertion error is a bit odd, so I'm not thrilled with that option.

I personally strongly in favor of the `return` and `fail` idiom for any tests that verify that a test method throws `AssertionError` unless the test can be written with one line of code (in which case  `@Test(expected = AssertionError.class)` is fine).

Luckily, with Java 8 we can all use lambdas so this will all be moot some day.

panchenko: I would agree with "a bit odd", however it's shorter. Will revert back to `return` :-)

src/test/java/org/junit/tests/assertion/AssertionTest.java
```diff
@@ -717,6 +717,27 @@ public void stringNullAndNullWorksToo() {
...
+        try {
+            assertEquals(new NullToString(), new NullToString());
+        } catch (AssertionError e) {
+            assertEquals("expected: org.junit.tests.assertion.AssertionTest$NullToString<null> but "
+                            + "was: org.junit.tests.assertion.AssertionTest$NullToString<null>",
+                    e.getMessage());
+            return;
+        }
+
+        fail("AssertionError expected");
```
panchenko: @kcooney I would like this extracted as a function, and will update other tests then separately.

kcooney: If we could could do that without hurting readability that is fine. For tests I am fine with some duplication if it makes the code clear.

PR #1455, opened 2017-05-19
consistent exception testing in AssertionTest

panchenko:
That was discovered while discussing #1454, so I would like it be more consistent.

### DIFF
--- a/src/test/java/org/junit/tests/assertion/AssertionTest.java
+++ b/src/test/java/org/junit/tests/assertion/AssertionTest.java
@@ -34,6 +34,10 @@ public class AssertionTest {
 //      assert false;
 //  }
 
+    private void failAssertionErrorExpected() {
+        throw new AssertionError("AssertionError expected");
+    }
+
     @Test(expected = AssertionError.class)
     public void fails() {
         Assert.fail();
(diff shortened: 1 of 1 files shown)


src/test/java/org/junit/tests/assertion/AssertionTest.java
```diff
@@ -34,6 +34,10 @@
 //      assert false;
 //  }
 
+    private void failAssertionErrorExpected() {
+        throw new AssertionError("AssertionError expected");
```
panchenko: It's intentional that exception is thrown directly here, as this class also have tests for the `fail()` method. So it not using fail() here in this particular case is a good thing :-)

kcooney: I think I would prefer to extract a static constant for "AssertionError expected" and inline this method. That's because it's 100% clear that this code is wrong:

    try {
      fail("woops');
      throw new AssertionError(ASSERTION_ERROR_EXPECTED);
    } catch (AssertionError expected) {
      ...

src/test/java/org/junit/tests/assertion/AssertionTest.java
```diff
@@ -367,9 +375,11 @@ public void arraysWithNullElementEqual() {
     public void stringsDifferWithUserMessage() {
         try {
             assertEquals("not equal", "one", "two");
-        } catch (Throwable exception) {
+        } catch (AssertionError exception) {
```
kcooney: You can change this to catch `ComparisonFailure` instead (in which case, it's perfectly fine to call `fail()` in the `try` block)

kcooney:
Thanks!

PR #1460, opened 2017-05-25
Handle AssumptionViolatedException in @Parameters method

### LINKED ISSUE #1329: assumeTrue doesn't work in @Parameters methods
My test class loads some expensive resources in the parameters function, so I want to skip that step if the entire class of tests is going to be skipped anyway. However, when I use `assumeTrue` in the parameters function the tests fail with the following:

```
org.junit.AssumptionViolatedException: got: <false>, expected: is <true>
    at org.junit.Assume.assumeThat(Assume.java:95)
    at org.junit.Assume.assumeTrue(Assume.java:41)
    at ExampleTest.getParams(ExampleTest.java:17)
```

I am running JUnit 4.12 within Eclipse. MWE below:

```
import static org.junit.Assume.assumeTrue;

import java.util.Collection;
import java.util.Collections;

import org.junit.runner.RunWith;
import org.junit.runners.Parameterized;
import org.junit.runners.Parameterized.Parameters;

@RunWith(Parameterized.class)
public class ExampleTest {

    @Parameters(name = "{index}: {1}")
    public static Collection<Object[]> getParams() {
//      assumeTrue(false);//bad
        return Collections.emptyList();
    }

    public ExampleTest(){
        assumeTrue(false);//okay
    }
}
```

panchenko:
Fixes #1329

### DIFF
--- a/src/main/java/org/junit/runners/Parameterized.java
+++ b/src/main/java/org/junit/runners/Parameterized.java
@@ -330,10 +330,12 @@ public class Parameterized extends Suite {
             this.exception = exception;
         }
 
+        @Override
         public Description getDescription() {
             return description;
         }
 
+        @Override
         public void run(RunNotifier notifier) {
             notifier.fireTestAssumptionFailed(new Failure(description, exception));
         }
--- a/src/main/java/org/junit/runners/Parameterized.java
+++ b/src/main/java/org/junit/runners/Parameterized.java
@@ -352,8 +352,8 @@ public class Parameterized extends Suite {
 
         private RunnersFactory(Class<?> klass) throws Throwable {
             testClass = new TestClass(klass);
-            List<Object> allParametersResult;
             parametersMethod = getParametersMethod(testClass);
+            List<Object> allParametersResult;
             AssumptionViolationRunner assumptionViolationRunner = null;
             try {
                 allParametersResult = allParameters(testClass, parametersMethod);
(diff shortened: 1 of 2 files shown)


src/main/java/org/junit/runners/Parameterized.java
```diff
@@ -365,7 +396,13 @@ private TestWithParameters createTestWithNotNormalizedParameters(
         @SuppressWarnings("unchecked")
         private static List<Object> allParameters(
                 TestClass testClass, FrameworkMethod parametersMethod) throws Throwable {
-            Object parameters = parametersMethod.invokeExplosively(null);
+            Object parameters;
+            try {
+                parameters = parametersMethod.invokeExplosively(null);
+            } catch (AssumptionViolatedException e) {
```
kcooney: Couldn't you catch this exception in the caller (`RunnersFactory`)? If the exception gets caught, you can set a field in `RunnersFactory` named `runnerOverride` to a new `AssumptionViolationRunner`. If `runnerOverride` is set it would be returned from `RunnersFactory.createRunners()`.

As a bonus, we could possibly catch any exception from there, and for the exceptions that aren't `AssumptionViolatedException ` use `ErrorReportingRunner`.

panchenko: I've implemented the first part. 

`ErrorReportingRunner` is already applied in `RunnerBuilder.safeRunnerForClass()`, seems like we should not care about other exceptions.

kcooney:
@panchenko Thanks! Could you update the release notes on the wiki?

garfieldnate:
Thank you!

panchenko:
@kcooney I've updated release notes for this.

PR #1463, opened 2017-06-08
Call methodBlock() after fireTestStarted()

### LINKED ISSUE #1462: BlockJUnit4ClassRunner.methodBlock() should be called after fireTestStarted()
In the fix for #1066 (https://github.com/junit-team/junit4/commit/a90b496a6595856066504baf4f737fb853a6e45d) `methodBlock()` is being called before `fireTestStarted()`. If creating the test instance is slow or has visible side-effects, it should be done after we send the notification that the test has started.

panchenko:
Fixes #1462


kcooney:
Thank you so much for getting this done so quickly!

PR #1465, opened 2017-06-13
Provide helpful message if parameter cannot be set.

stefanbirkner:
For private @Parameter fields is users get an exception like "java.lang.IllegalAccessException: Class ... can not access a member of class X with modifiers private" The new message "Cannot set parameter 'parameter'. Ensure that the the field 'parameter' is public." tells the user what they should do.

The reason for adding this feature is the Stackoverflow question https://stackoverflow.com/questions/44522046/reflection-exception-in-parameterized-junit-test-using-array-parameter/44522988

### DIFF
--- a/src/main/java/org/junit/runners/parameterized/BlockJUnit4ClassRunnerWithParameters.java
+++ b/src/main/java/org/junit/runners/parameterized/BlockJUnit4ClassRunnerWithParameters.java
@@ -73,6 +73,13 @@ public class BlockJUnit4ClassRunnerWithParameters extends
             int index = annotation.value();
             try {
                 field.set(testClassInstance, parameters[index]);
+            } catch (IllegalAccessException e) {
+                IllegalAccessException wrappedException = new IllegalAccessException(
+                        "Cannot set parameter '" + field.getName()
+                                + "'. Ensure that the field '" + field.getName()
+                                + "' is public.");
+                wrappedException.initCause(e);
+                throw wrappedException;
             } catch (IllegalArgumentException iare) {
                 throw new Exception(getTestClass().getName()
                         + ": Trying to set " + field.getName()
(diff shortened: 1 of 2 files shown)


src/main/java/org/junit/runners/parameterized/BlockJUnit4ClassRunnerWithParameters.java
```diff
@@ -73,6 +73,10 @@ private Object createTestUsingFieldInjection() throws Exception {
             int index = annotation.value();
             try {
                 field.set(testClassInstance, parameters[index]);
+            } catch (IllegalAccessException e) {
+                throw new Exception("Cannot set parameter '" + field.getName()
```
kcooney: Are you sure we shouldn't set the cause of this exception?

Also, throwing ``Exception` is rare. Perhaps we should throw a new `IllegalAccessException`

stefanbirkner: You're right with the cause. The problem with creating a new `IllegalAccessException` is that it doesn't have a constructor with a `cause` argument.

stefanbirkner:
When I have your okay for the pull request I will squash the commits and merge to master.

src/main/java/org/junit/runners/parameterized/BlockJUnit4ClassRunnerWithParameters.java
kcooney: You can use https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html#initCause(java.lang.Throwable) to set the cause.on the `IllegalAccessException`. Can we do that?

stefanbirkner: And again you're right. I fixed it and already squashed the commits.

PR #1470, opened 2017-06-27
Support non-public base classes containing public @Before or @Rule members.

### LINKED ISSUE #1468: Junit @before order is not honor when parent class is package private
According to junit 4.12 documentation

>The @Before methods of superclasses will be run before those of the current class, unless they are overridden in the current class. No other ordering is defined.

But when the parent class is package private derived @before method is executed before. i.e:

```
package com.jcrada.junit.test;

import static org.junit.Assert.assertEquals;
import org.junit.Before;
import org.junit.Test;

public class DerivedTestClass extends ParentTestClass {

    @Before
    public void setup(){
        System.out.println("setup derived class");
    }

    @Test
    public void testDummy(){
        assertEquals(1, 1);
    }
}
```
```
package com.jcrada.junit.test;

import org.junit.Before;

class ParentTestClass {

    @Before
    public void init(){
        System.out.println("setup parent class");
    }
}

```

output is

> setup derived class
setup parent class

kcooney:
Fixes #1468


src/main/java/org/junit/runners/model/FrameworkField.java
```diff
@@ -21,6 +21,15 @@
                     "FrameworkField cannot be created without an underlying field.");
         }
         this.field = field;
+
+        if (isPublic()) {
+            // This method could be a public field in a package-scope base class
```
marcphilipp: > This *field* could be ...

panchenko:
TBH `.setAccessible(true)` would allow using even not public members.

src/main/java/org/junit/runners/model/FrameworkField.java
kcooney: Fixed. Thanks!

kcooney:
@panchenko the pull request has written would not change the behavior of fields or methods that are not public.

PR #1476, opened 2017-08-03
Each li tag should have a corresponding close tag

fanofxiaofeng:
It seems that the are two missing </li> tags.


kcooney:
Thanks for the contribution

You do not need to close <li> tags in JavaDoc

See https://www.w3.org/TR/html401/intro/sgmltut.html#h-3.2.1 and the examples in http://www.oracle.com/technetwork/articles/java/index-137868.html

fanofxiaofeng:
OK

PR #1478, opened 2017-08-07
Fix dead link to the ant task in FAQ

rybak:
Documentation to the ant junit task has been moved to another location.


kcooney:
Thanks!

PR #1486, opened 2017-10-04
Remove oraclejdk7 since it is longer supported by travi…

jxmai:
…sCI & add openjdk8 to travis.yml

openjdk6 is no longer supported by travisCI: 

see: https://github.com/travis-ci/travis-ci/issues/8199

---

oraclejdk7 is longer supported by travisCI:

see: https://github.com/travis-ci/travis-ci/issues/7884

---

-- openJDk8 has been available for some time

see: https://github.com/travis-ci/travis-ci/issues/6483

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: java
 script: mvn verify javadoc:javadoc site:site
 jdk:
-  - oraclejdk7
   - oraclejdk8
+  - openjdk8
   - openjdk7
-  - openjdk6


.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: java
 script: mvn verify javadoc:javadoc site:site
 jdk:
-  - oraclejdk7
   - oraclejdk8
+  - openjdk8
   - openjdk7
-  - openjdk6
```
stefanbirkner: We should not stop building on JDK6 because it provides us fast feedback whether JUnit can still be used with Java 6. #1487 installs OpenJDK6 on Travis.

Tibor17: You have Couldbees build with jdk 1.5?
Do not worry because you have `animal-sniffer-maven-plugin` which checks all Java API calls in source code whether they match with Java 1.5 signatures.

jxmai: I agree that we do have `animal-sniffer-maven-plugin` that can check all the compatibility with older Java versions, but it also makes sense continue to use travis to provide fast feedback.

However, the installed maven is not compatible with Java 6 so only putting `openjdk` in the add-on is not gonna to work.

In #1487 

> The failing build shows that the installed Maven cannot be used with JDK6. I fix this after October, 15th.

I guess the fix is to force to use maven 3.2.5 during `install` which is the same trick as https://github.com/stefanbirkner/system-rules/commit/012552c80d8ca8b01b2939740e8719a18b944385

PR #1491, opened 2017-11-29
Fix TemporaryFolder deletion after test execution

udaysagar2177:
Fixed a no-op loop that could occur after we fail to delete one of the files under the current TemporaryFolder. I thought it is better to delete as many files as possible before returning back to TemporaryFolder#delete() caller function.

### DIFF
--- a/src/main/java/org/junit/rules/TemporaryFolder.java
+++ b/src/main/java/org/junit/rules/TemporaryFolder.java
@@ -295,8 +295,8 @@ public class TemporaryFolder extends ExternalResource {
         boolean result = true;
         File[] files = file.listFiles();
         if (files != null) {
-            for (File each : files) {
-                result = result && recursiveDelete(each);
+            for (int i = 0; i < files.length && result; i++) {
+                result = recursiveDelete(files[i]);
             }
         }
         return result && file.delete();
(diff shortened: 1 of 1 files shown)


udaysagar2177:
Let me know if we want to fail fast as soon as we are unable to delete any file inside the TemporaryFolder and I can update the code accordingly.

kcooney:
The current code fails fast, which seems reasonable.

udaysagar2177:
Actually, if it fails to delete a file, it turns into a no-op loop meaning it will still iterate through all the files and just does not call recursive delete as result is false in the and condition.

kcooney:
I am not too worried about the no-op loop, given it is rare that there would be a failure deleting a file. In addition, the performance cost of the extra times through the loop is dwarfed by the cost of hitting the filesystem to get the file list.

udaysagar2177:
Ok, I updated the code to fail fast when there is a failure. Feel free to close this PR if this fix is not needed but please let me know why. Thanks!

kcooney:
@udaysagar2177 I think your proposal makes the code a bit harder to read. I added a commit with a more simpler approach to your remote branch. Let me know what you think.

Note that since none of these changes affect the number of calls to `File.delete()` or `File.listFiles()` I don't personally think that any of these micro-optimizations are worth it.

udaysagar2177:
Yes, the new commit is much simpler. I agree that we don't get any performance improvement here but the new commit you added greatly simplifies reading the initial code.

PR #1493, opened 2017-12-31
Revert "Fluent expectations for ExpectedException".

kcooney:
The change was not binary compatible with JUnit 4.12.
In addition, the preferred way to check for exceptions
in JUnit 4.13 is with Assert.assertThrows() so we will
be not adding new major features to ExpectedException.

This reverts commit 1db00d5a965951ea3667480a96467b43a80823c1.


stefanbirkner:
Good catch. I removed the according section from the release notes: https://github.com/junit-team/junit4/wiki/4.13-Release-Notes/_compare/7c424430fbe072ace0c2ccb3ba0cb42270e55727...09c3dc24255f2095325b0f7606b1f8a4dc90f0db

Issue #1436, opened 2017-04-02
Update TestWatcher Javadoc to specify that it should be the "outer" rule


Gaibhne:
Depending on the **field name** of the ExpectedException rule, the rule works or breaks when used in conjunction with a TestWatcher. More details, as well as a test case demonstrating the faulty behaviour, can be seen at:

http://stackoverflow.com/questions/43088543/junit-test-both-passes-and-fails-conflict-using-both-expectedexception-and-tes

kcooney:
The ordering of rules is undefined. If you need to enforce ordering between rules, you can use `RuleChain`. We could perhaps make it easier to discover this. Any suggestions?

kcooney:
@Gaibhne do you have any suggestions for how you could have discovered `RuleChain` as the solution to your problem (Java doc improvements, updates to the wki, etc)?

panchenko:
The ordering of rules is address be #1445

Gaibhne:
While ordering fixes the issue, what happens is not actually something that SHOULD be something that depends on the order of the rules, so there is no way I would have ever gotten the idea to read any documentation relating to rule orders.

Don't get me wrong, as a programmer, I see why it happens the way it does, but that is only in hind sight. Nothing but a meaningful error or warning message could have pointed me towards looking into rule ordering mechanisms.

I'm not sure if it is feasible to produce any sort of error or warning for a scenario like this. In the end, the bug is that the TestWatcher follows different rules, no pun intended, about what is considered a failure or a success than JUnit.

panchenko:
This issue can be addressed in the following ways:
1. If `TestWatcher` is implemented as a rule it can access the final result only it is the outermost one. This limitation can be added to the class javadoc.
1. Extend ordering of rules, so a rule implementation can specify how it should be ordered.
1. Deprecate `TestWatcher` and implement replacement via `RunListener`. Would be the best approach.

kcooney:
I think updating the Javadoc of `TestWatcher` makes since so I've updated the summary of this issue accordingly.

stefanbirkner:
I started work on this issue.

stefanbirkner:
I'm currently tidying up the tests for TestWatcher. This is just a preparatory refactoring that allows me to add a test for the upcoming arguments about ordering in  the Javadoc.

PR #1502, opened 2018-02-11
order TestWatcher to be outermost

panchenko:
As a possible way to address ordering in #1436

The change of spaces is done to make `LoggingTestRule` consistent with `LoggingTestWatcher`.

### DIFF
--- a/src/main/java/org/junit/rules/DefaultOrder.java
+++ b/src/main/java/org/junit/rules/DefaultOrder.java
@@ -9,5 +9,6 @@ import java.lang.annotation.Target;
 @Inherited
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
-public @interface OutermostRule {
+public @interface DefaultOrder {
+    int value();
 }
--- a/src/main/java/org/junit/rules/OutermostRule.java
+++ b/src/main/java/org/junit/rules/OutermostRule.java
@@ -1,0 +1,13 @@
+package org.junit.rules;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Inherited
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface OutermostRule {
+}
--- a/src/main/java/org/junit/rules/TestWatcher.java
+++ b/src/main/java/org/junit/rules/TestWatcher.java
@@ -43,6 +43,7 @@ import org.junit.runners.model.Statement;
  *
  * @since 4.9
  */
+@OutermostRule
 public abstract class TestWatcher implements TestRule {
     public Statement apply(final Statement base, final Description description) {
         return new Statement() {
--- a/src/main/java/org/junit/rules/TestWatcher.java
+++ b/src/main/java/org/junit/rules/TestWatcher.java
@@ -43,7 +43,7 @@ import org.junit.runners.model.Statement;
  *
  * @since 4.9
  */
-@OutermostRule
+@DefaultOrder(Integer.MIN_VALUE)
 public abstract class TestWatcher implements TestRule {
     public Statement apply(final Statement base, final Description description) {
         return new Statement() {
--- a/src/main/java/org/junit/runners/RuleContainer.java
+++ b/src/main/java/org/junit/runners/RuleContainer.java
@@ -8,6 +8,7 @@ import java.util.List;
 
 import org.junit.Rule;
 import org.junit.rules.MethodRule;
+import org.junit.rules.OutermostRule;
 import org.junit.rules.TestRule;
 import org.junit.runner.Description;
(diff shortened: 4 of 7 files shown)


kcooney:
Interesting idea. I worry that this could break existing tests. If it does, there may not be a way for the user to fix their tests.

panchenko:
I agree it's tricky. Also agree with deprecating `ExpectedException`.

A possible improvement could be doing this only if `@Rule` has default order.

kcooney:
@panchenko so perhaps a `DefaultOrder` annotation?

panchenko:
There are 2 parts: 
* preferred/default order for some rules
* how it can be overridden if needed, e.g. if `@Rule` has `order != Rule.DEFAULT_ORDER`

kcooney:
Agreed. I am just suggesting that the class level annotation have an attribute which is the default order value.

stefanbirkner:
TLDR: I propose to not merge this pull request and use `@Rule(order = Integer.MIN_VALUE)` for solving #1436. See pull request #1501.

I think that the annotation `@OutermostRule` adds a lot of complexity considering the problem that it solves. Currently we have one mechanism for ordering rules: the `order` attribute of the `@Rule` annotation. (You could also consider a ordering mechanism that rules created by methods are collected first, but for now I leave this aside.) `@OutermostRule` adds a second mechanism by separating the rules into two classes and enforcing the following order: outermost - low order, outermost - high order, standard rule - low order, standard rule - high order. Unfortunately this is not apparent to the reader of a test:

    public class SomeTest {
      @Rule(order = 1)
      public final Testrule firstRule = new FirstRule();

      @Rule(order = 2)
      public final Testrule secondRule = new SecondRule(); //this is a outermost rule

      @Rule(order = 3)
      public final Testrule thirdRule = new ThirdRule();

      @Rule(order = 4)
      public final Testrule fourthRule = new FourthRule(); //this is a outermost rule

      ...
    }

The order of these rules is `secondRule`, `fourthRule`, `firstRule`, `thirdRule`. In order to understand this the reader must know about `@OutermostRule` and they must know that `SecondRule` and `ThirdRule` are annotated with `@OutermostRule`. The same rule order can be achieved by 

    public class SomeTest {
      @Rule(order = 1)
      public final Testrule firstRule = new FirstRule();

      @Rule(order = Integer.MAX_VALUE)
      public final Testrule secondRule = new SecondRule();

      @Rule(order = 3)
      public final Testrule thirdRule = new ThirdRule();

      @Rule(order = Integer.MAX_VALUE + 1)
      public final Testrule fourthRule = new FourthRule();

      ...
    }

The reader must only know about rule ordering. What is lost is that without the `order` attribute the rules `secondRule` and `fourthRule` would still enclose `firstRule` and `thirdRule`.

The reason for implementing this annotation was the problem that TestWatcher should always be the outermost rule (see #1436). Currently I don't know another use case for `@OutermostRule`. But the `TestWatcher` problem can already be solved by using `@Rule(order = Integer.MIN_VALUE)` without implementing a second mechanism for ordering.

Unfortunately the `order` approach has a drawback, too. The test writer is now responsible for adding the `order` attribute to the rule. The best we can do is documentation in the class' Javadoc. Hopefully the popular IDEs implement a warning when a `TestWatcher` is used without an explicit `order` attribute.

If we still want `TestWatcher`s automatically to be the outermost rules than we should do this without the additional annotation. We can check rules for being a `TestWatcher` and adjust the order of rules accordingly.

kcooney:
@stefanbirkner my proposal was to have the `order` attribute of `Rule` override this behavior of having`TestWatcher` be the outermost rule.

That being said, I agree that adding documentation might be the best approach.

panchenko:
I've changed annotation to `@DefaultOrder`.
The next question if it should be specified for other rules, like `Verifier`, etc.

stefanbirkner:
The changed code shows pretty clear how `@DefaultOrder` works. Thanks for that piece of work. Nevertheless I'm still in favor for fixing the problem with documentation and hopefully IDE support.

Although the classification issue is gone it has been replaced by a new issue: the value of `order` is hard to choose because it is not absolute. Only for the outermost and the innermost rule the `order` is kind of clear.

The rest of my preceding comment still applies to `@DefaultOrder`.

kcooney:
If we solve this problem with code, the current pull is the way I would like to solve it.

The value for `DefaultOrder.order` is hard to choose, but that doesn't bother me because it's a default value which users can override via `Rule.order`

What bothers me is we would yet again be changing the order that things happen for existing tests. For large code bases these kids of changes slow down people from upgrading.

stefanbirkner:
The change of the order of existing tests together with the added complexity makes me reluctant to merge this pull request. With the doc fix all existing tests keep successful while new tests can fix the `TestWatcher` problem and even the `ErrorCollector/ExpectedException` problem by setting `order`. If we go for the doc fix then I create a doc update for `ErrorCollector/ExpectedException`, too.

panchenko:
I think an update for those classes documentation would be useful in any case (with or without this).

kcooney:
Since #1501 merged, and people were at best luke-warm on this approach, I am closing.

PR #1505, opened 2018-02-15
Run ErrorCollectorTest with Parameterized runner

stefanbirkner:
The array for the Parameterized gives a better overview of the tests.

### DIFF
--- a/src/test/java/org/junit/rules/ErrorCollectorTest.java
+++ b/src/test/java/org/junit/rules/ErrorCollectorTest.java
@@ -1,24 +1,94 @@
 package org.junit.rules;
 
-import org.hamcrest.CoreMatchers;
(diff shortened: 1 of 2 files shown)


panchenko:
I agree it looks shorter, how convenient is this for debugging?

stefanbirkner:
The debugging experience depends on the IDE. With IntelliJ it is possible to debug a single parameter set. I have no actual experience with other IDEs.

marcphilipp:
Now the build seems to fail on JDK 5: https://junit.ci.cloudbees.com/job/JUnit/10/console

@stefanbirkner Can you please take a look?

stefanbirkner:
I will have a look at it, but maybe not today.

PR #1507, opened 2018-02-19
Fix bug on JDK 9

stefanbirkner:


src/test/java/org/junit/runners/model/TestClassTest.java
```diff
@@ -132,26 +132,41 @@ public String methodA() {
...
-    	    annotatedMethods.size(), is(3));
-    	assertThat("First annotated method is wrong.", annotatedMethods
-    	    .iterator().next().getName(), is("methodA"));
+        TestClass tc = new TestClass(MethodsAnnotated.class);
+        List<FrameworkMethod> annotatedMethods = tc.getAnnotatedMethods();
+        assertThat(extractNames(annotatedMethods), anyOf(
+            //Java < 9:
+            equalTo(asList("methodA", "methodB", "methodC")),
+            //Java >= 9 has annotations on methods of java.lang.Object:
+            equalTo(asList("clone", "finalize", "getClass", "hashCode",
```
panchenko: I would suggest ignoring the `jdk.internal.*` annotations completely.

stefanbirkner: Although I understand your motivation I would like to keep it like it is. Just for simplicity because it works fine and there is no need for an additional filtering. IMHO the problem here is that the test is not good enough. I think about a better test.

kcooney: I agree with Alex. tte tests would be less.fragile if you filtered out the `jdk.internal.*` annotations. They are, after all, internal.

stefanbirkner:
I rewrote the tests so that they are less fragile. Therefore I don't have to change `getAnnotatedMethods` by adding a filter.

src/test/java/org/junit/runners/model/TestClassTest.java
stefanbirkner: I rewrote the tests so that they are less fragile. Therefore I don't have to change `getAnnotatedMethods` by adding a filter.

PR #1514, opened 2018-03-28
Shadowed static methods/fields should override parent class behavior.

### LINKED ISSUE #1509: [JUnit 4.13 regression] @BeforeClass shadowing in superclass doesn't work
In JUnit 4.13 version the `@BeforeClass` method of superclass is run even if it is shadowed by current class, which is not as expected according to current JUnit 4.12 behaviour and `@BeforeClass` javadoc:
```
The "@BeforeClass" methods of superclasses will be run before those of the current class,
unless they are shadowed in the current class.
```

Repro code:
```
class Junit413RegressionSuperclass {
  @BeforeClass
  public static void beforeAll() {
    System.out.println("beforeAll - superclass");
  }
}

public class Junit413Regression extends Junit413RegressionSuperclass {
  @BeforeClass
  public static void beforeAll() {
    System.out.println("beforeAll - class");
  }

  @Test
  public void test() {
    System.out.println("test - class");
  }
}
```

Expected output (JUnit 4.12):
```
beforeAll - class
test - class
```

Actual output (JUnit 14.13-SNAPSHOT):
```
beforeAll - superclass
beforeAll - class
test - class
```

kcooney:
Fixes #1509

### DIFF
--- a/src/main/java/org/junit/runners/model/FrameworkField.java
+++ b/src/main/java/org/junit/runners/model/FrameworkField.java
@@ -47,12 +47,7 @@ public class FrameworkField extends FrameworkMember<FrameworkField> {
 
     @Override
     public boolean isShadowedBy(FrameworkField otherMember) {
-        return false;
-    }
-
-    @Override
-    FrameworkField handlePossibleBridgeMethod(List<FrameworkField> members) {
-        return this; // fields are never bridge methods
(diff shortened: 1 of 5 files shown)


kcooney:
@medb any feedback?

medb:
LGTM

PR #1515, opened 2018-04-06
Recommend to set order for ExpectedException

### LINKED ISSUE #1466: ErrorCollector will not fail tests if used in conjunction with an ExpectedException
I stumbled across a bit of an obscure edge case in one of our tests using Junit 4.12 and I'm wondering if this is a bug or if it's expected behavior.

It appears that any errors collected by the `ErrorCollector` are not reported if test throws an exception and the exception is also expected by the `ExpectedException`

For an example, see the first test method here:

```java
package com.alexblanck;

import static org.hamcrest.CoreMatchers.equalTo;

import org.junit.Rule;
import org.junit.Test;
import org.junit.rules.ErrorCollector;
import org.junit.rules.ExpectedException;

public class ConflictingRules {

    @Rule
    public ErrorCollector errors = new ErrorCollector();

    @Rule
    public ExpectedException exception = ExpectedException.none();

    @Test
    public void succeeds_butExpectedToFailBasedOnErrorCollector() {
        errors.checkThat(3, equalTo(4));
        exception.expect(IllegalStateException.class);
        throw new IllegalStateException();
    }

    @Test(expected = IllegalStateException.class)
    public void fails_basedOnAnnotation_asExpected() {
        errors.checkThat(3, equalTo(4));
        throw new IllegalStateException();
    }

    @Test
    public void fails_basedOnErrorCollector_asExpected() {
        errors.checkThat(3, equalTo(4));
    }

    @Test
    public void succeeds_basedOnExpectedException_asExpected() {
        exception.expect(IllegalStateException.class);
        throw new IllegalStateException();
    }
}
```

As you can see in the example, the test does fail as expected when `@Test(expected=` annotation instead of the `ExpectedException` rule.

Digging into the code a bit, it seems that the problem is an ordering issue. `ErrorCollector` extends`Verifier` and therefore only checks for the presence of errors after the base statement evaluates without an exception. If the `expected` annotation is used, the "base" statement is a `ExpectException` statement, so it does not throw and the `ErrorCollector` runs.

Using a RuleChain, I can get my example to behave as expected by forcing the ErrorCollector to be the "outerRule":

```
@Rule
public RuleChain chain = RuleChain.outerRule(errors).around(exception);
```

I'm not sure if this causes other unintended behavior though. I found [another issue](https://github.com/junit-team/junit4/issues/1012) that talks about a similar ordering issue, but it proposed the reverse ordering.

stefanbirkner:
A test that should fail because of ErrorCollector may not fail when this test throws an exception that is expected by ExpectedException.

Fixes #1466.

### DIFF
--- a/src/main/java/org/junit/rules/ExpectedException.java
+++ b/src/main/java/org/junit/rules/ExpectedException.java
@@ -68,6 +68,13 @@ import org.junit.runners.model.Statement;
  *     throw new NullPointerException(&quot;What happened?&quot;);
  * }</pre>
  *
+ * <p>It is recommended to set the {@link org.junit.Rule#order() order} of the
+ * {@code ExpectedException} to {@code Integer.MAX_VALUE} if it is used together
+ * with another rule that handles exceptions, e.g. {@link ErrorCollector}.
+ * Otherwise failing tests may be successful.
+ * <pre> &#064;Rule(order = Integer.MAX_VALUE)
+ * public ExpectedException thrown = ExpectedException.none();</pre>
+ *
  * <h3>AssumptionViolatedExceptions</h3>
  * <p>JUnit uses {@link AssumptionViolatedException}s for indicating that a test
  * provides no useful information. (See {@link org.junit.Assume} for more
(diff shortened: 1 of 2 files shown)


kcooney:
Thanks!

Should we deprecate `ExpectedException`?

stefanbirkner:
I created #1519 for further discussion about deprecation of `ExpectedExeption`.

PR #1518, opened 2018-04-27
Fix testRemoveElement of ListTest 

clumsyfox-z9o9z:
The test failed due to IndexOutOfBoundsException in fullList.remove(3). It seems that the value 3 should be removed but unfortunately ArrayList.remove() takes the index of the element and not the element itself. This fix ensures that element 3 is removed by its index.

### DIFF
--- a/src/test/java/junit/samples/ListTest.java
+++ b/src/test/java/junit/samples/ListTest.java
@@ -38,7 +38,7 @@ public class ListTest extends TestCase {
 
     public void testContains() {
         assertTrue(fullList.contains(1));
-        assertTrue(!emptyList.contains(1));
+        assertFalse(emptyList.contains(1));
     }
 
     public void testElementAt() {
@@ -61,7 +61,7 @@ public class ListTest extends TestCase {
     }
 
     public void testRemoveElement() {
-        fullList.remove(Integer.valueOf(3));
-        assertTrue(!fullList.contains((Integer) 3));
+        fullList.remove(2);
+        assertFalse(fullList.contains(3));
     }
 }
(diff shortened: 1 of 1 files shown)


src/test/java/junit/samples/ListTest.java
```diff
@@ -61,7 +61,8 @@ public void testRemoveAll() {
     }
 
     public void testRemoveElement() {
-        fullList.remove(3);
```
panchenko: `.remove(Integer.valueOf(3))` then

panchenko: `.remove((Integer) 3)`

src/test/java/junit/samples/ListTest.java
```diff
@@ -61,7 +61,7 @@ public void testRemoveAll() {
...
-        assertTrue(!fullList.contains(3));
+        fullList.remove(Integer.valueOf(3));
+        assertTrue(!fullList.contains((Integer) 3));
```
kcooney: No need to cast here (the test doesn't cast in other calls to contains)

Also while you are here, please replace every place that does `assertTrue(!expression)` with `assertFalse(expression)`

src/test/java/junit/samples/ListTest.java
```diff
@@ -61,7 +61,7 @@ public void testRemoveAll() {
     }
 
     public void testRemoveElement() {
-        fullList.remove(3);
-        assertTrue(!fullList.contains(3));
+        fullList.remove(2);
```
kcooney: fullList.remove(Integer.valueOf(3));

stefanbirkner:
This PR has been merged into master. The commits are cceaf19c554f9c0bcd51223ad99fc77f31f9e877 and 5aa5f1641387daba3e566c54c83e674abc683d6c.

Thanks for improving and fixing the test.

PR #1519, opened 2018-05-03
Deprecate rule ExpectedException

stefanbirkner:
The method Assert.assertThrows provides a nicer way for verifying exceptions. In addition the use of ExpectedException is error-prone when used with other rules like TestWatcher because the order of rules is important in that case.

### Pro

`assertThrows` is in any way better than `ExpectedException`:
- the code that throws the exception can be exactly specified
- no dependency on rule ordering
- the exception is exposed so that any of its properties can be checked

### Contra

`assertThrows` is only easy to use with Java 8. For older versions of Java you need boiler-plate code for creating a `ThrowingRunnable`

### DIFF
--- a/src/main/java/org/junit/rules/ExpectedException.java
+++ b/src/main/java/org/junit/rules/ExpectedException.java
@@ -14,11 +14,7 @@ import org.junit.runners.model.Statement;
 
 /**
  * The {@code ExpectedException} rule allows you to verify that your code
- * throws a specific exception. Note that, starting with Java 8,
- * {@link org.junit.Assert#assertThrows(java.lang.Class, org.junit.function.ThrowingRunnable)
- * Assert.assertThrows}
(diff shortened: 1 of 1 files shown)


stefanbirkner:
This is a PR for the question raised by @kcooney in #1515: "Should we deprecate ExpectedException?"

It should be used for discussion about the suggested deprecation.

stefanbirkner:
The argument against deprecation only applies to codebases that update to JUnit 4.13 but don't change their build to JDK 8. I think that codebases that don't change the Java version are also not updating their JUnit dependency.

kcooney:
There are enough problems with `ExpectedException` that I would recommend try..catch over `ExpectedException` for any codebase where you can't use lambdas.

stefanbirkner:
Good point.

PR #1524, opened 2018-05-25
Nullaway intellij

yh73:


yh73:
I am so sorry. I accidentally push the files to the wrong location. It should be pushed to my forked version. I will delete it ASAP.

PR #1527, opened 2018-05-31
Add a new feature to let test methds in a class run as user specified order

JacobCN:
This is a new feature to let JUnit run as you specified order. What we need do is two steps:
1. Adding annotation 'FixMethodOrder(MethodSorters.SPECIFIED_PRIORITY)'.
2. Specify 'priority' value for methed annotation 'Test'.

This feature is an optional featue to let JUnit users specify method order in some special scenarios.
More details please refer to the comments in changed code.

### DIFF
--- a/src/main/java/org/junit/Test.java
+++ b/src/main/java/org/junit/Test.java
@@ -95,4 +95,16 @@ public @interface Test {
      * </p>
      */
     long timeout() default 0L;
+
+    /**
+     * Optionally specify <code>priority</code>, a int value, to let the methods run as <code>priority</code> ascending order,
+     * <p>
+     * <b>NOTE:</b>
+     * <br/>
+     * If you want it works, you must use <code>@FixMethodOrder(MethodSorters.SPECIFIED_PRIORITY)</code> in your test class.
+     * <br/>
+     * If two methods have the same priority, they will run as the {@link org.junit.internal.MethodSorter#DEFAULT} order.
+     * </p>
+     */
+    int priority() default 0;
 }
--- a/src/main/java/org/junit/internal/MethodSorter.java
+++ b/src/main/java/org/junit/internal/MethodSorter.java
@@ -41,7 +41,6 @@ public class MethodSorter {
      * If <code>priority</code> have the same value, it will run as {@link #DEFAULT} order.
      */
     public static final Comparator<Method> SPECIFIED_PRIORITY = new Comparator<Method>() {
-        @Override
         public int compare(Method m1, Method m2) {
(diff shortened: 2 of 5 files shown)


JacobCN:
Add a new feature for @Test, to let JUnit specify method order in one method.

kcooney:
Thanks for the contribution.

I am hoping we could include #1130 in 4.13, which is a more general and flexible way to solve this problem. With that pull, you could create a custom`Ordering` that ordered the methods by any criteria you specify (including annotations on the method). Some users will.want to specify order via a priority, others with a unique priority, and others via some dependency.

JacobCN:
@kcooney Thanks for your comment.

I read the code of #1130, which it is a good solution. And as a JUnit user I hope to add the order feature ASAP, cause it is really a pressing need. But I think the solution in this PR is more convenient in some scenario, especially when I could not describe the order by programming. What's more, this solution is an optional way to describe the test method order.

**Notice** This feature in this PR has the same behavior as another unit test framework **TestNG**. This is a convenient way as my experience.
  * _There is no worry about about the license conflict with TestNG License Apache 2.0. My code is from my own design and there is no any code reference to TestNG, otherwise it will not work well._

kcooney:
I'm curious what the other maintainers think. @junit-team/junit-committers 

While adding an attribute is convenient, I would prefer we don't merge this pull for 4.13, for these reasons:

1. Some people strongly feel that tests should be independent and not depend on ordering. For those users, adding a `priority` or `order` attribute to `@Test` is too easy and therefore sends the wrong message
2. This pull relies on `@FixMethodOrder`, which was a temporary work-around for Java no longer guaranteeing the order that methods are returned from the reflection APIs
3. It's not that hard to write a custom annotation and an associated Ordering
4. We could update`@Test` later after we see how people use `Ordering`

dsaff:
Momentarily delurking to say that I think it would be good for the JUnit design if @FixMethodOrder was used less and less, not more and more.  (Basically everything @kcooney said.)

JacobCN:
@kcooney @dsaff , thanks for your explanations.
I think I caught the key point.  '<@FixMethodOrder>' is only a temporary work-around. @kcooney Do you mean this annotation is a deprecatory solution, or it will be removed or marked by '<@deprecated>' in the future?

kcooney:
@JacobCN if Ordering is added in 4.12 then I will probably deprecate `@FixMethodOrder`. If past behavior is any predictor of future behavior, we won't remove it until after the thermal death of our sun.

JacobCN:
Got it, I close this PR now.

Issue #1525, opened 2018-05-27
Assumption violation is not reflected on text UI


davido:
It would be helpful if assumption violations would be reflected on the text UI. For this test class with three test methods:

```java
package org.gerritcon.mv2016;

import static org.junit.Assert.assertEquals;
import static org.junit.Assume.assumeTrue;

import org.junit.Test;

public class PrintyTest {
  @Test
  public void test_passed() {
    assertEquals("foo", "foo");
  }

  @Test
  public void test_failed() {
    assertEquals("foo", "bar");
  }

  @Test
  public void test_assumption_violation() {
    assumeTrue(false);
  }
}
```

JUnit text UI produces this result:

```
$ java -classpath buck-out/gen/lib__printy_tests#testsjar__output/printy_tests#testsjar.jar:/home/davido/Downloads/hamcrest-core-1.3.jar:/home/davido/Downloads/junit-4.12.jar org.junit.runner.JUnitCore org.gerritcon.mv2016.PrintyTest
JUnit version 4.12
..E.
Time: 0.008
There was 1 failure:
1) test_failed(org.gerritcon.mv2016.PrintyTest)
org.junit.ComparisonFailure: expected:<[foo]> but was:<[bar]>
	at org.junit.Assert.assertEquals(Assert.java:115)
[...]
	at org.junit.runner.JUnitCore.main(JUnitCore.java:36)

FAILURES!!!
Tests run: 3,  Failures: 1
```

The impression is that 2 tests were successful even though, only single test was successful and another one failed with assumption violation. This information is lost.
 
Compared to it the output from [Buck build tool](https://buckbuild.com/) is:

```
$ buck test :printy_tests
FAILURE org.gerritcon.mv2016.PrintyTest test_failed: expected:<[foo]> but was:<[bar]>
[-] PROCESSING BUCK FILES...FINISHED 0.0s 🏖  (Watchman reported no changes)
[-] DOWNLOADING... (0.00 B/S AVG, TOTAL: 0.00 B, 0 Artifacts)
[-] BUILDING...FINISHED 0.0s [100%] (6/6 JOBS, 0 UPDATED, 0 [0.0%] CACHE MISS)
[-] TESTING...FINISHED 0.3s (1 PASS/1 SKIP/1 FAIL)
RESULTS FOR //:printy_tests
FAIL    <100ms  1 Passed   1 Skipped   1 Failed   org.gerritcon.mv2016.PrintyTest
[...]

PR #1528, opened 2018-05-31
Now tests with assumption violations are flagged with 'I'

JxJxJxMxNxS:
According to #1525 now the tests that has assumption violations are flagged with 'I' in the textUI output

### DIFF
--- a/src/main/java/org/junit/runner/Result.java
+++ b/src/main/java/org/junit/runner/Result.java
@@ -137,7 +137,7 @@ public class Result implements Serializable {
 
         @Override
         public void testAssumptionFailure(Failure failure) {
-            ignoreCount.getAndIncrement();
+            // do nothing: same as passing (for 4.5; may change in 4.6)
         }
     }
 
--- a/src/main/java/org/junit/runner/Result.java
+++ b/src/main/java/org/junit/runner/Result.java
@@ -137,7 +137,7 @@ public class Result implements Serializable {
 
         @Override
         public void testAssumptionFailure(Failure failure) {
-            // do nothing: same as passing (for 4.5; may change in 4.6)
+            ignoreCount.getAndIncrement();
         }
     }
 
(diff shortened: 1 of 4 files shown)


stefanbirkner (approved):

src/main/java/org/junit/runner/Result.java
```diff
@@ -137,7 +137,7 @@ public void testIgnored(Description description) throws Exception {
 
         @Override
         public void testAssumptionFailure(Failure failure) {
-            // do nothing: same as passing (for 4.5; may change in 4.6)
+            ignoreCount.getAndIncrement();
```
kcooney: In #1294 (currently targeted for 4.13) we are proposing updating this class to collect assumption failure counts, so I am not sure we should make this change.

kcooney (commented):
Thanks for taking this on.

Could we hold off a bit on this? There were recent comments on the bug.

src/main/java/org/junit/runner/Result.java
JxJxJxMxNxS: Oh great! in my opinion add a counter for the assumption failures would be great in order to know exactly how many did the test have

src/main/java/org/junit/runner/Result.java
```diff
@@ -137,7 +137,7 @@ public void testIgnored(Description description) throws Exception {
 
         @Override
         public void testAssumptionFailure(Failure failure) {
-            ignoreCount.getAndIncrement();
+            // do nothing: same as passing (for 4.5; may change in 4.6)
```
Tibor17: This is absolutely wrong. We rely on it!
You cannot remove the line. There is no field counter for assumptions and even if it was there, it is a change in behavior in all other frameworks and their integration tests will fail.

kcooney: @Tibor17 I suggest commenting on the bug.

PR #1533, opened 2018-06-08
Added supporting for comparing CharSequence objects. Added tests.

IgorNosach25:
"Asserts for CharSequence contents #1432" is resolved. Please apply my changes.

### DIFF
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -649,6 +649,36 @@ public class Assert {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


kcooney:
Thanks for the contribution.

If JUnit has assertions for `CharSequence` (and I am not 100% convinced it should) I prefer to iterate on #1439

Note there are many projects that focus on making comprehensive assertion APIs (Hamcrest, Fest, Truth...), so we sometimes decline requests to add new kinds of assertions to JUnit

PR #1545, opened 2018-07-27
fixes issue #1432 about asserting charSequence equal

spector5:
adds assertContentsEqual for CharSequences


kcooney:
This has some I releated changes (.classpath, . project...) which would need to be reverted before we could consider merging this.

We also have pull #1485 for this same issue. Not sure why no one on the JUnit team commented on that pull.

PR #1558, opened 2018-10-03
Merge two if branches with the implementation

renuka-fernando:
- merge two if branches with the same implementation (`return true` in the both branches).
- use `Collections.singletonList` instead of `Arrays.asList` since it is only one item.


marcphilipp:
Thanks again, @renuka-fernando!

PR #1559, opened 2018-10-03
Remove generic wildcard in return parameters

renuka-fernando:
- remove generic wildcard in return parameters.
- remove explicit array creation in `newInstance(Object ... initargs)` method.

### DIFF
--- a/src/main/java/junit/framework/TestSuite.java
+++ b/src/main/java/junit/framework/TestSuite.java
@@ -56,12 +56,12 @@ public class TestSuite implements Test {
         Object test;
         try {
             if (constructor.getParameterTypes().length == 0) {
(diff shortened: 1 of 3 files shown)


marcphilipp:
If I'm not mistaken, removing the wildcard breaks source compatibility so we shouldn't do it.

@renuka-fernando Could you please explain why you want to remove it?

renuka-fernando:
It is recommended not to use wildcard types as return types. @marcphilipp could you please refer
- https://rules.sonarsource.com/java/RSPEC-1452

marcphilipp:
I agree in general, but using a raw type is not a solution and we cannot use a concrete type here, either. Thus, I'm closing this PR. Feel free to submit a new one that just contains the varargs change! 🙂

PR #1560, opened 2018-10-07
Replace for with foreach

renuka-fernando:
- Simplify `If` statements in `Double` and `Float` checking methods with of `Assert` class.

### DIFF
--- a/src/main/java/org/junit/rules/TemporaryFolder.java
+++ b/src/main/java/org/junit/rules/TemporaryFolder.java
@@ -200,8 +200,8 @@ public class TemporaryFolder extends ExternalResource {
         File relativePath = null;
         File file = root;
         boolean lastMkdirsCallSuccessful = true;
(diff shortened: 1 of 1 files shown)


kcooney:
Thanks for the contribution.

The changes to Assert take fewer lines, but I am not sure iI agree that the newer code is simpler. I could see replacing the second "if (something) { return false } return true" with "return inverted something". That being said, presumably the original author and reviewer(s) were happy with what was there...

renuka-fernando:
Yes, I also felt it's easy to understand with if blocks.

kcooney:
@renuka-fernando could you delete the first commit, keep the second commit (the "for-each" commit) and force-push the changes to GitHub?

renuka-fernando:
@kcooney
I have deleted the first commit, thank you.

kcooney:
@renuka-fernando thanks! Merged.

PR #1561, opened 2018-10-11
Fixed small incorrect javadoc link

lbroman:


src/main/java/org/junit/runners/model/MemberValueConsumer.java
```diff
@@ -11,7 +11,7 @@
     /**
      * Receives the next value and its declaring member.
      *
-     * @param member declaring member ({@link FrameworkMethod or {@link FrameworkField}}
+     * @param member declaring member ({@link FrameworkMethod} or {@link FrameworkField}}
```
marcphilipp: The last character should be `)`, not `}`.

PR #1562, opened 2018-10-11
Removed eclipse config files and added them to gitignore

lbroman:
It is considered bad practice to handle IDE configuration in GIT so I have cleaned up the eclipse files and added them to the ignore file.

If you have policies to keep eclipse files specifically, please feel free to ignore my pull request.


kcooney:
I personally use the checked-in eclipse files. JUnit also often gets contributions from less-experianced developers that do not know how to import projects into Eclipse without the project files.

At the very least if we remove the files we should provide documentation on how to generate them or otherwise get started with Eclipse and IntelliJ.

lbroman:
Good point. I can write a simple guide on how to get started with Eclipse and maven. Is this something that should be on the github wiki or in the source tree?

panchenko:
To be honest: there is no need to change anything. 
Why fixing something that's not broken?

lbroman:
Depends on your view on if unclean is broken or not. Or what is even an unclean project. I'm don't think I have any say on that in this project anyway. Feel free to close my pull request.

kcooney:
If we want to add documentation about working with Eclipse, I suggest updatinghttps://github.com/junit-team/junit4/blob/master/CONTRIBUTING.md

Currently the Eclipse files specify dependencies that are already listed in the pom, so checking them in does violate Don't Repeat Yourself. So there are some minor advantages to removing them.

That being said, I wouldn't expect the dependencies to change, since 4.13 is likely to be the last release of the 4.x code line.

wangminpku:
Dear lbroman (@lbroman )and Junit team (@kcooney)(@marcphilipp )(@stefanbirkner ) (@panchenko )(@Tibor17 )(@aslakhellesoy )

We are working on automatically identifying and clustering the related pull requests for code review. We have found that [PR#821](https://github.com/junit-team/junit4/pull/821) might be related to this one, because both of them focus on the same problem. 

We would really appreciate it if you could help us to validate and give us some feedback if the information from [PR#821](https://github.com/junit-team/junit4/pull/821) is helpful when you review the current PR. 

marcphilipp:
Closing due to lack of consensus and inactivity.

PR #1563, opened 2018-10-11
Changed Assert.assertThat to MatcherAssert.assertThat

lbroman:
Since Assert.assertThat only calls MatcherAssert.assertThat
it makes sense to start cleaning this up to make
 future deprecation an easier process.


marcphilipp:
At this point it's very unlikely that we're going to remove `Assert.assertThat`. Besides, it's not even deprecated. In the future, please create an issue so we can discuss potential code changes before starting to work on them.

PR #1568, opened 2018-10-26
Improve message if file exists instead of folder 

stefanbirkner:
When `TemporaryFolder(String path)` or `TemporaryFolder(String... paths)` is
called with a path that matches the path of an existing file then an
`IOException` with a message like "a file with path <path> exists" is
thrown.

### DIFF
--- a/src/main/java/org/junit/rules/TemporaryFolder.java
+++ b/src/main/java/org/junit/rules/TemporaryFolder.java
@@ -206,8 +206,13 @@ public class TemporaryFolder extends ExternalResource {
 
             lastMkdirsCallSuccessful = file.mkdirs();
             if (!lastMkdirsCallSuccessful && !file.isDirectory()) {
-                throw new IOException(
-                        "could not create a folder with the path \'" + relativePath.getPath() + "\'");
+                if (file.exists()) {
+                    throw new IOException(
+                            "a file with the path \'" + relativePath.getPath() + "\' exists");
+                } else {
+                    throw new IOException(
+                            "could not create a folder with the path \'" + relativePath.getPath() + "\'");
+                }
             }
         }
         if (!lastMkdirsCallSuccessful) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/junit/rules/TemporaryFolder.java
```diff
@@ -205,9 +205,16 @@ public File newFolder(String... paths) throws IOException {
...
             lastMkdirsCallSuccessful = file.mkdirs();
-            if (!lastMkdirsCallSuccessful && !file.isDirectory()) {
-                throw new IOException(
-                        "could not create a folder with the path \'" + relativePath.getPath() + "\'");
+            if (!lastMkdirsCallSuccessful) {
+                if (file.exists() && file.isDirectory()) {
+                    continue;
+                } else if (file.exists()) {
+                    throw new IOException(
+                            "a file with the path \'" + relativePath.getPath() + "\' exists");
```
npathai: We can include `already` in the message, as `createNewFile` includes it in error message.

stefanbirkner: I was thinking about the word "already" for this message two. But it is wrong here because the user wants to create a folder and not a file. Therefore we cannot say the file is already there.

src/main/java/org/junit/rules/TemporaryFolder.java
```diff
@@ -205,9 +205,16 @@ public File newFolder(String... paths) throws IOException {
             file = new File(root, relativePath.getPath());
 
             lastMkdirsCallSuccessful = file.mkdirs();
-            if (!lastMkdirsCallSuccessful && !file.isDirectory()) {
-                throw new IOException(
-                        "could not create a folder with the path \'" + relativePath.getPath() + "\'");
+            if (!lastMkdirsCallSuccessful) {
+                if (file.exists() && file.isDirectory()) {
+                    continue;
```
marcphilipp: I think we could avoid the `continue` by moving the `!file.isDirectory()` back into the enclosing `if` and remove the first inner `if`.

stefanbirkner: You're right. Fixed.

PR #1569, opened 2018-10-27
Resolves #1365

### LINKED ISSUE #1365: Interrupted test cases cause random failures with thread reuse
Whenever a test case is interrupted (Thread.interrupt()), the test runner thread seems to stay in interrupted state. This causes random failures later when the thread is reused for other test cases.

Expected behaviour: When the thread is reused, interrupted flag for the thread is cleaned before starting a new test case. This can be done by calling `Thread.interrupted()`, which returns the interrupted state, and clears the interrupted flag.

The following code should reproduce the problem, given that test1 is run before test2

``` java
import org.junit.Test;

public class ThreadTest {
  @Test
  public void test1() {
    Thread.currentThread().interrupt();
  }

  @Test
  public void test2() throws InterruptedException {
    Thread.sleep(1000);
  }
}
```

The problem I encountered is similar to the one described in https://www.thoughtwire.com/junit-interupt/

npathai:
Resolves #1365 
 - `runLeaf` method of `ParentRunner` clears the interrupt status flag after execution of each test
 - Added test case to verify interrupt status

### DIFF
--- a/src/main/java/org/junit/runners/ParentRunner.java
+++ b/src/main/java/org/junit/runners/ParentRunner.java
@@ -7,7 +7,6 @@ import static org.junit.internal.runners.rules.RuleMemberValidator.CLASS_RULE_VA
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
@@ -215,6 +214,7 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
             statement = withBeforeClasses(statement);
             statement = withAfterClasses(statement);
             statement = withClassRules(statement);
+            statement = withInterruptIsolation(statement);
         }
         return statement;
     }
@@ -290,6 +290,22 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
             @Override
             public void evaluate() {
                 runChildren(notifier);
+            }
+        };
+    }
+
+    /**
+     * @return a {@link Statement}: clears interrupt status of current thread after execution of statement
+     */
+    protected final Statement withInterruptIsolation(final Statement statement) {
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                try {
+                    statement.evaluate();
+                } finally {
+                    Thread.interrupted(); // clearing thread interrupted status for isolation
+                }
             }
         };
     }
--- a/src/test/java/org/junit/tests/running/classes/ThreadsTest.java
+++ b/src/test/java/org/junit/tests/running/classes/ThreadsTest.java
@@ -1,0 +1,91 @@
+package org.junit.tests.running.classes;
+
+import org.junit.AfterClass;
+import org.junit.Test;
+import org.junit.internal.runners.ErrorReportingRunner;
+import org.junit.runner.Description;
+import org.junit.runner.JUnitCore;
+import org.junit.runner.Request;
+import org.junit.runner.Result;
+import org.junit.runner.RunWith;
+import org.junit.runner.Runner;
+import org.junit.runner.notification.RunListener;
+import org.junit.runners.BlockJUnit4ClassRunner;
+import org.junit.runners.model.InitializationError;
+
+import static org.junit.Assert.assertEquals;
+
+public class ThreadsTest {
+    private String log = "";
+
+    public static class TestWithInterrupt {
+
+        @Test
+        public void interruptCurrentThread() {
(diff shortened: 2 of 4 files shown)


src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -346,7 +345,7 @@ protected final void runLeaf(Statement statement, Description description,
         EachTestNotifier eachNotifier = new EachTestNotifier(notifier, description);
         eachNotifier.fireTestStarted();
         try {
-            statement.evaluate();
```
kcooney: I wonder if instead of updating `runLeaf()` we should update `methodBlock()` to return a `Statement` that calls `Thread.interrupted()` at the end of the statement chain.

kcooney: We should almost certainly modify `classBlock()` to return a `Statement` that calls T`hread.interrupted()` at the end of the statement chain.

npathai: @kcooney Updated `methodBlock()` to return a statement that clears interrupt status and reverted `runLeaf`. I am not sure how updating `classBlock()` will help?

kcooney: @npathai I assume that we would want to handle the case where the interrupt bit was set  in a class rule or `@AfterClass` method. Those are handled by `classBlock()`

npathai:
@kcooney I have incorporated the review comments and cleared interrupt status flag from `methodBlock()` and `classBlock()`.

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -294,6 +294,22 @@ public void evaluate() {
         };
     }
 
+    /**
+     * @return a {@link Statement}: clears interrupt status of current thread after execution of statement
+     */
+    protected Statement withInterruptIsolation(final Statement statement) {
```
kcooney: I personally would prefer this is be `final`

npathai:
@kcooney Yes I agree. Made the method `final`. Should I squash the commits?

kcooney:
@npathai we usually do a squash and merge unless the author asks us to keep the commits separate. If you think these changes should be broken up a certain way on the main branch, please squash them.

kcooney:
Here are the logs:

```
$ ~/bin/install-jdk.sh --target "/home/travis/openjdk6" --workspace "/home/travis/.cache/install-jdk" --feature "6" --license "GPL"
install-jdk.sh 2018-10-17
Expected feature release number in range of 9 to 12, but got: 6
The command "~/bin/install-jdk.sh --target "/home/travis/openjdk6" --workspace "/home/travis/.cache/install-jdk" --feature "6" --license "GPL"" failed and exited with 3 during .
```

src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -356,7 +372,7 @@ protected final void runLeaf(Statement statement, Description description,
         }
     }
 
-    /**
+  /**
```
kcooney: Unintentional change?

src/test/java/org/junit/tests/running/classes/ThreadsTest.java
```diff
@@ -0,0 +1,86 @@
+package org.junit.tests.running.classes;
+
+import org.junit.AfterClass;
+import org.junit.Test;
+import org.junit.internal.runners.ErrorReportingRunner;
+import org.junit.runner.*;
```
kcooney: For newer code we avoid wildcard imports.

src/main/java/org/junit/runners/ParentRunner.java
npathai: Done. Escaped from my review.

src/test/java/org/junit/tests/running/classes/ThreadsTest.java
npathai: Done. IDE behavior :disappointed:

npathai:
@kcooney All done. Raise issue for failing openjdk-6 build as well

marcphilipp:
@kcooney This should make it into 4.13, right?

kcooney:
@marcphilipp your call.

ijuma:
It would be great to get this in so the long awaited 4.13 can be released. :)

marcphilipp:
Since this PR seems to have stalled, I've resubmitted the changes in #1586.

npathai:
@marcphilipp Apologies, I was away for some time so couldn't reply.

marcphilipp:
@npathai No worries! I picked up your work and finished it in #1586.

PR #1572, opened 2018-11-25
Fix UnsupportedOperationException on class rules

dherges:
Collections.emptyList() returns an immutable list. Calls to `BlockJUnit4ClassRunner#classRules().add(...)` used to work before, now result in UnsupportedOperationException. This fix returns a mutable, empty list.

See https://circleci.com/gh/spektrakel-blog/sparkles/752?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

### DIFF
--- a/src/main/java/org/junit/runners/ParentRunner.java
+++ b/src/main/java/org/junit/runners/ParentRunner.java
@@ -542,7 +542,7 @@ public abstract class ParentRunner<T> extends Runner implements Filterable,
 
         public List<TestRule> getOrderedRules() {
             if (entries.isEmpty()) {
-                return Collections.emptyList();
+                return new ArrayList<TestRule>();
             }
             Collections.sort(entries, RuleContainer.ENTRY_COMPARATOR);
             List<TestRule> result = new ArrayList<TestRule>(entries.size());


src/main/java/org/junit/runners/ParentRunner.java
```diff
@@ -542,7 +542,7 @@ public void accept(FrameworkMember member, TestRule value) {
 
         public List<TestRule> getOrderedRules() {
             if (entries.isEmpty()) {
```
panchenko: I would suggest removing "if" then and adding a test.

panchenko:
> Is this a regression introduced in 4.13-beta-1?

yes, that's code for decoratively ordering rules.

Tibor17:
one way or another this is not a bug because this getter is not real getter since it does not return `field`. In real, this getter returns local variable and always a new one by every call. Modifications do not modify the status of `ParentRunner` so any modifications do not make sense. Copy the reference in subclass. It is really strange that this method has two reasons for `return` which is the reason of the bug and unnecessary performance tuning which does not make visible effect.
There is no performance problem to specify the size of `ArrayList` because nobody write a million `rules` in a class.
Think simple and start the method with:
` List<TestRule> result = new ArrayList<TestRule>()`
then logic here
`return result;`

dherges:
Technically, yes, it's a regression. Working solution is to wrap the list:

https://github.com/spektrakel-blog/sparkles/blob/737ca09755ffa00a66a65434d81676366191d06d/support/javalin/testing/src/main/java/sparkles/support/javalin/testing/JavalinTestRunner.java#L22-L27

kcooney:
It's a regression. While I understand that it might not make sense to return a mutable list, but I don't want to release 4.13 with a known regression.

kcooney:
Incidentally, I don't agree that there is no benefit to specifying the initial size if the `ArrayList`.

marcphilipp:
@dherges Do you have time to finish this PR or shall someone else take over?

dherges:
Yes, can somebody else please take over. I don't understand points about what makes sense and what is not a bug and so on.

kcooney:
I think Tibor is the only who stated that they thought this wasn't a bug.

This is clearly a regression from 4.12.Given the number of third-party runners out there that could be affected, I (as a JUnit maintainer) think we should fix it.

panchenko:
I'll do it today, that would be fair, as I've introduced that optimization.

kcooney:
Fixed in #1578

PR #1578, opened 2018-12-17
ParentRunner.classRules() returns modifiable List

panchenko:
See #1572


dherges:
Looks good! Thanks! :)

PR #1586, opened 2019-01-12
Clear thread interrupt status after each test

### LINKED ISSUE #1365: Interrupted test cases cause random failures with thread reuse

marcphilipp:
The thread interrupt status flag is now cleared from `classBlock()` and
`methodBlock()`. The flag is cleared after each test case completes and
after `AfterClass` methods and `ClassRules` have been executed.

Fixes #1365.

---

Originally submitted as #1569.

### DIFF
--- a/src/test/java/org/junit/tests/running/classes/ThreadsTest.java
+++ b/src/test/java/org/junit/tests/running/classes/ThreadsTest.java
@@ -2,21 +2,25 @@ package org.junit.tests.running.classes;
 
 import org.junit.AfterClass;
 import org.junit.Test;
-import org.junit.internal.runners.ErrorReportingRunner;
 import org.junit.runner.Description;
 import org.junit.runner.JUnitCore;
 import org.junit.runner.Request;
 import org.junit.runner.Result;
 import org.junit.runner.RunWith;
-import org.junit.runner.Runner;
 import org.junit.runner.notification.RunListener;
 import org.junit.runners.BlockJUnit4ClassRunner;
-import org.junit.runners.model.InitializationError;
 
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.util.Arrays.asList;
+import static java.util.Collections.singletonList;
 import static org.junit.Assert.assertEquals;
 
 public class ThreadsTest {
-    private String log = "";
+
+    private List<Boolean> interruptedFlags = new ArrayList<Boolean>();
+    private JUnitCore core = new JUnitCore();
 
     public static class TestWithInterrupt {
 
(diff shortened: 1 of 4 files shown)


kcooney:
I unfortunately do not have time to help with JUnit at this time

src/test/java/org/junit/tests/running/classes/ThreadsTest.java
```diff
@@ -0,0 +1,98 @@
...
+                interruptedFlags.add(Thread.currentThread().isInterrupted());
+            }
+        });
+
+        Request request = new Request() {
+            @Override
+            public Runner getRunner() {
+                try {
+                    return new BlockJUnit4ClassRunner(TestWithInterruptFromAfterClass.class) {
+                    };
```
panchenko: is this anonymous class necessary/intentional?

src/test/java/org/junit/tests/running/classes/ThreadsTest.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+    }
+
+    @Test
+    public void currentThreadInterruptedStatusIsClearedAfterEachTestExecution() {
+        core.addListener(new RunListener() {
+            @Override
+            public void testFinished(Description description) {
+                interruptedFlags.add(Thread.currentThread().isInterrupted());
```
panchenko: 1) This listener is duplicated in both tests. As JUnitCore is a field, might be add this listener from a `@Before` method?
2) Another thing this listener can do is just `assertFalse()`, WDYT?

marcphilipp: The listeners actually implement a different method but have the same implementation. Just doing `assertFalse()` would have the downside that we're not checking that the listener was actually called.

src/test/java/org/junit/tests/running/classes/ThreadsTest.java
marcphilipp: Good catch, it wasn't necessary. I simplified the test in 20a130ef8dfdd533202e9e5938ccc111b6e832f1.

PR #1592, opened 2019-02-18
Update JDK build matrix

sormuras:
- remove JDK 9 (EOL)
- remove JDK 10 (EOL)
- add JDK 12 (RC)


sormuras:
Either Java 12 xor Java 6.

sormuras:
`javac` says: `Source option 6 is no longer supported. Use 7 or later.`

PR #1594, opened 2019-03-10
Document order attribute of Rule and ClassRule

### LINKED ISSUE #1593: 4.13 docs: Rule refers to deprecated RuleChain for order despite having order() property
Version: 4.13-beta2

`Rule` JavaDoc
> You can use a `{@link org.junit.rules.RuleChain}` if you want to have control over the order in which the Rules are applied.

`ClassRule` JavaDoc doesn't mention `order()`

It would be nice to have the same example as on `RuleChain` written on `Rule` and `ClassRule` using the `order()` property to ease migration of the deprecation.

marcphilipp:
Instead of mentioning the now deprecated `RuleChain`, the Javadoc of the
`Rule` and `ClassRule` annotation now contains an example on how to use
the new `order` attribute.

Resolves #1593.

### DIFF
--- a/src/main/java/org/junit/ClassRule.java
+++ b/src/main/java/org/junit/ClassRule.java
@@ -30,6 +30,8 @@ import java.lang.annotation.Target;
  * undefined, in general. However, Rules defined by fields will always be applied
  * after Rules defined by methods, i.e. the Statements returned by the former will
  * be executed around those returned by the latter.
+ *
+ * <h3>Usage</h3>
  * <p>
  * For example, here is a test suite that connects to a server once before
  * all the test classes run, and disconnects after they are finished:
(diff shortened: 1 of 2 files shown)


src/main/java/org/junit/ClassRule.java
```diff
@@ -80,6 +82,26 @@
...
+ * <pre>
+ * public class ThreeClassRules {
+ *     &#064;ClassRule(order = 0)
+ *     public static LoggingRule outer = new LoggingRule("outer rule");
+ *
+ *     &#064;ClassRule(order = 1)
+ *     public static LoggingRule middle = new LoggingRule("middle rule");
+ *
+ *     &#064;ClassRule(order = 2)
+ *     public static LoggingRule inner = new LoggingRule("middle rule");
```
TWiStErRob: copy-paste

src/main/java/org/junit/Rule.java
```diff
@@ -63,6 +63,26 @@
...
+ * <pre>
+ * public class ThreeRules {
+ *     &#064;Rule(order = 0)
+ *     public LoggingRule outer = new LoggingRule("outer rule");
+ *
+ *     &#064;Rule(order = 1)
+ *     public LoggingRule middle = new LoggingRule("middle rule");
+ *
+ *     &#064;Rule(order = 2)
+ *     public LoggingRule inner = new LoggingRule("middle rule");
```
TWiStErRob: copy-paste

marcphilipp: Indeed... 😬

marcphilipp: Fixed in de2ebb35.

PR #1597, opened 2019-03-13
Annotate ExpectedException with @Deprecated.

kcooney:
The Javadoc already indicates that it is deprecated.


kcooney:
@marcphilipp thanks for the quick approval. Could we add this to 4.13? This is so trivial that I don't think we need a new release candidate.

marcphilipp:
Agreed. 👍

Issue #1496, opened 2018-02-06
Release JUnit 4.13


kcooney:
This issue tracks the work needed to release JUnit 4.13

kcooney:
@junit-team/junit-committers any thoughts on a 4.13 release? I have unfortunately been very busy with my day job so I can't drive a 4.13 release, but I can certainly help.

I've associated some open issues with Milestone 4.13 as a starting point, but it appears that milestones cannot have comment threads, so I figured we can discuss scheduling and which features should be included here

marcphilipp:
One thing I would like to do is aligning `Assert.assertThrows`/`expectThrows` with Jupiter's `Assertions.assertThrows`. I know there was some controversy around that a while back but now that Jupiter's been released I think we should consider migration scenarios and remove `Assert.assertThrows` and rename `expectThrows` to `assertThrows`. Any objections?

panchenko:
The last attempt to remove one of those was #1396

kcooney:
I can see both sides of the debate about merging `assertThrows` and `expectThrows`, so I wouldn't object to a pull removing `expectThrows` and updating `assertThrows` to return the caught exception. I suggest reaching out the people involved in the original pull adding these methods to 4.x, to inform them of the resolution of this issue. I realize some people would have been upset regardless of what direction we decided to take with the APIs, but still think that it's good to make sure everyone knows their voice is being heard.

While we are at it, it would be great to port `assertThrows(Class<T> expectedType, Executable executable, String message)` to JUnit4 (see https://github.com/junit-team/junit5/commit/ea67ca0)

zjffdu:
any progress on 4.13 release ?

kcooney:
@zjffdu yes. See https://github.com/junit-team/junit4/pulse/monthly for recent work and https://github.com/junit-team/junit4/milestone/3 for the proposed remaining work.

Tibor17:
It's been almost 4 years after last release 4.12.
Cut 4.13-beta-1 release version for deeper testing.

cljohnso:
[...]

PR #1599, opened 2019-03-21
Add test to verify failed is called when withBefores fails

TimvdLippe:
This is a regression test for a change in behavior between JUnit 4.12
and 4.13.

This test fails on `master`, but passes on `r4.12`. You can run `mvn test` on this PR to see it failing. You can run `mvn test` on https://github.com/TimvdLippe/junit4/tree/run-regression-test-on-4-12 to see it passing.

The difference in behavior seems to be related to the handling of the exception in `withBefores`. The `fail()` ensures that that fails fast. 4.12 will then state that the test was not started, did fail and did not finish, but 4.13 will report all as true.

Related to #1496

### DIFF
--- a/src/test/java/org/junit/runner/AllRunnerTests.java
+++ b/src/test/java/org/junit/runner/AllRunnerTests.java
@@ -11,7 +11,8 @@ import org.junit.runners.Suite.SuiteClasses;
         FilterFactoriesTest.class,
         FilterOptionIntegrationTest.class,
         JUnitCommandLineParseResultTest.class,
-        JUnitCoreTest.class, RequestTest.class
+        JUnitCoreTest.class, RequestTest.class,
+        RunnerOrderTest.class
 })
 public class AllRunnerTests {
 }
--- a/src/test/java/org/junit/runner/RunnerOrderTest.java
+++ b/src/test/java/org/junit/runner/RunnerOrderTest.java
@@ -1,0 +1,73 @@
+package org.junit.runner;
+
+import static junit.framework.TestCase.assertTrue;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.fail;
+
+import org.junit.Test;
+import org.junit.runner.notification.Failure;
+import org.junit.runner.notification.RunListener;
+import org.junit.runner.notification.RunNotifier;
+import org.junit.runners.BlockJUnit4ClassRunner;
+import org.junit.runners.model.FrameworkMethod;
+import org.junit.runners.model.InitializationError;
+import org.junit.runners.model.Statement;
+
+public class RunnerOrderTest {
+
+  @Test
+  public void runnerWithFailingTests() throws InitializationError {
+    ReportingRunner runner = new ReportingRunner(FailingTests.class);
+    runner.run(new RunNotifier());
+
+    assertFalse(runner.started);
+    assertTrue(runner.failed);
+    assertFalse(runner.finished);
+  }
+
+  public static class FailingTests {
+    @Test
+    public void this_test_is_supposed_to_fail() {
+      fail();
+    }
+  }
+
+  class ReportingRunner extends BlockJUnit4ClassRunner {
+
+    private boolean started;
+    private boolean failed;
+    private boolean finished;
+
+    public ReportingRunner(Class<?> testClass) throws InitializationError {
+      super(testClass);
+    }
+
+    @Override
+    protected Statement withBefores(FrameworkMethod method, Object target, Statement statement) {
+      throw new IllegalArgumentException();
+    }
+
+    @Override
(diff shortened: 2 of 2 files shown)


kcooney:
I think the 4.13 behavior is correct. If a `@Before` method is called--even if it fails--the test has already started. The work between "start" and "finish" should include `@Before` methods, `@After` methods and rules annotated with `@Rule` (because they are all part of the test fixture)

TimvdLippe:
While it is probably more correct, my concerns are largely to the change in behavior. I am not sure what the extents are of the change in logic. There might be other projects such as Mockito who rely on the specific ordering. We can roll forward a fix, but that would leave projects using an older version of Mockito with JUnit 4.13 with erroneous behavior.

If you decide to go ahead, I do acknowledge that. I just wanted to make you aware of this change in subtle behavior that might have unforeseen larger consequences.

kcooney:
@TimvdLippe understood that any change can have an impact. Now that we know what broke the Mockito tests, have you been able to determine the change(s) to JUnit that triggered the change in behavior? That might give us an idea of the potential impact, and perhaps revisit the change. I tried to find it by browsing the code, but didn't see anything obvious.

I believe this is the code that has the problem:

https://github.com/mockito/mockito/blob/a323b8132de6f6e1c29d738de245469f8ce009b0/src/main/java/org/mockito/internal/runners/DefaultInternalRunner.java#L56

Am I correct? If so, what's the implication of not ever getting inside that `if` statement?

Many tools check the JUnit version to work-around behavior changes in JUnit. You can get the version by calling `junit.runner.Version.id()`. If this part of the Mockito code shouldn't have a compile-time dependency on JUnit, you could call it via reflection.

kcooney:
BTW, thanks for raising this issue. As a developer in an extremely large code base that uses JUnit, I am well aware of the cost that small changes in JUnit behavior can have.

TimvdLippe:
> Am I correct? If so, what's the implication of not ever getting inside that if statement?

The Mockito runner has to clean up listeners to prevent memory starvation. In JUnit 4.12, we did not get a signal when the test finished. We thus, in `testFailed` implemented the logic to handle that case. However, it also assumed that `testFinished` was not called as well, as that should not also clean up the listener.

Moreover, we didn't call `testFinished` on our Mockito listeners. This callback is usually used to reset mocks or perform cleanup work. Since Mockito was unable to initialize any mocks, I think listeners would incorrectly try to clean up mocks (and then fail).

In any case, I think we can work quite easily around this new behavior in Mockito. I am just hesitant for making such subtle changes that could trip people over. If this change can be "easily" reverted, that would be my suggested path forward.

Let me know what you think :smile:

kcooney:
@junit-team/junit-committers any thoughts?

stefanbirkner:
I agree that the behavior of JUnit 4.13 is what I would have expected. I still try to get an idea how much code bases would have failing tests after an upgrade to JUnit 4.13. @TimvdLippe Do i get it right that all tests that use Mockito will fail?

TimvdLippe:
No not all tests. The change behaviour changes the contract we have with Mockito framework listeners. E.g. projects that listen for Mockito events and act accordingly. For example, it allows frameworks to perform some initialization/cleanup before and after a test is run with Mockito.

marcphilipp:
I've done some digging and found this commit to be the reason for the behavior change: https://github.com/junit-team/junit4/commit/a90b496a6595856066504baf4f737fb853a6e45d. It seems it's a fix we should probably have released in a patch release. Thus, I think we should go through with the change.

TimvdLippe:
Once JUnit 4.13 is officially released, I can update Mockito and publish a new release. Then we just have to say to our users that they need to update their Mockito version if they do run into issues.

marcphilipp:
I've submitted https://github.com/mockito/mockito/pull/1672 which should make the behavior consistent for 4.12 and 4.13.

I might be missing something but I don't think this will break anyone since the behavior only changed for failures when initializing mocks. Instead of seeing a class-level error, all tests will now be reported as failed. Everything else should work as before, right?

TimvdLippe:
> I've submitted mockito/mockito#1672 which should make the behavior consistent for 4.12 and 4.13.

That's awesome. Merged 😄 
> Everything else should work as before, right?

We now throw more events than before. If framework integrations weren't expecting them, then it might have broken them. I imagining a report system here which keeps track of `testFinished` and assume it means a success instead of failure. That is an imaginary example though, as we don't know how and if people integrate with Mockito that way.

At least now we are consistent with 4.12 and 4.13 and I think that is the most important. Let's hope we won't get any reports on Mockito about the regression 😛

marcphilipp:
Please note that `testFinished` was already called for failing tests before, just not for those that failed during initialization.

PR #1605, opened 2019-04-13
Restore that fields shadow fields from super class

stefanbirkner:
This commit restores the shadowing behaviour of JUnit 4.12. Some users of JUnit 4 have tests that rely on this behaviour. Their tests would fail with JUnit 4.13.

This commit finally reverts 39b8a92d24e6c81e4fc614cdf13003ce17bbc166. Some parts have already been reverted by ed6813d3bbe2da4d67892f73e641fa80ab4d126e.


stefanbirkner:
I already removed #1414 from the realase notes.

PR #1608, opened 2019-05-05
Fix deprecation messages for Hamcrest dependencies

sbrannen:


sbrannen:
I actually noticed that the referenced `MatcherAssert` class also exists under referenced package; however, that is in the dedicated `hamcrest-junit` artifact.

So if you continue to reference that package, I think it would at least be beneficial to the reader to mention the artifact it comes from.

Similarly, for the deprecated `assumeThat()` method, it would be nice if the deprecation message mentions the `hamcrest-junit` artifact explicitly. Otherwise users will be confused why `MatcherAssert` is available in `hamcrest-core` while `MatcherAssume` is not.

marcphilipp:
There are other references that were also changed in 3289d9422b535666786aa3201ea0f31ad5e1f7e3. AFAIK they reference [hamcrest-junit](https://github.com/hamcrest/hamcrest-junit) which has been discontinued.

@tumbarumba Right?

tumbarumba:
Yes, that's correct. `hamcrest-junit` was an early, discontinued experiment in repackaging the Hamcrest classes into a more coherent form. It should not have been published, and we don't recommend that anyone uses it.

The "official" Hamcrest releases are:
- version `1.3`. This is a last version code compatible and compiled with Java 1.5. It has artifacts `hamcrest-core` and `hamcrest-library`
- version `2.1`. This is the latest version, it is compatible with Java 1.7. It has a single artifact called `hamcrest`, though we still published artifacts for `hamcrest-core` and `hamcrest-library`, which are just there to provide transitive dependencies on `hamcrest`.

More details are available on the [Hamcrest Distributables](http://hamcrest.org/JavaHamcrest/distributables) page.

TWiStErRob:
I agree with https://github.com/junit-team/junit4/pull/1608#issuecomment-489444641, I'm really confused, using 4.13-beta-3 + Hamcrest 2.1 and my project doesn't compile because `MatcherAssume`  is gone (@tumbarumba where is it?).

@marcphilipp @sbrannen Please add the artifacts AND versions to the deprecation message, and promote a stable fully functional combination of junit and hamcrest.

---

The current dependency structure for 2 simple libraries is far from ideal:
```gradle
configurations.all {
	resolutionStrategy.eachDependency { DependencyResolveDetails dep ->
		if (dep.requested.group == 'org.hamcrest') {
			switch (dep.requested.name) {
				case 'java-hamcrest':
					dep.useTarget("org:hamcrest:hamcrest:${dep.target.version}")
					dep.because("2.0.0.0 shouldn't have been published")
					break
				case 'hamcrest-core':
					dep.useTarget("org:hamcrest:hamcrest:${dep.target.version}")
					dep.because("hamcrest-core doesn't contain anything")
					break
				case 'hamcrest-library':
					dep.useTarget("org:hamcrest:hamcrest:${dep.target.version}")
					dep.because("hamcrest-library doesn't contain anything")
					break
			}
		}
	}
}
dependencies { // test
	testImplementation "junit:junit:4.13-beta-3"
	testImplementation "org.hamcrest:hamcrest:2.1"
	testImplementation "org.hamcrest:hamcrest-junit:2.0.0.0"
}
```
`junit` still depends on `hamcrest-core`, `hamcrest` 2.1 doesn't have `MatcherAssume` so still need `hamcrest-junit`, but `hamcrest-junit` depends on `java-hamcrest` for fun. All hail `resolutionStrategy` for helping with this 🥇. This is actually so complicated that it would be really nice if both JUnit and Hamcrest dedicated a section in their setup guides.

marcphilipp:
In the light of that it would probably be most convenient for users if we undeprecated `Assume.assumeThat()` and change its implementation to use `describeMismatch()`. It's not as easy to add it to Hamcrest because it requires and exception from JUnit.

@junit-team/junit-committers @tumbarumba Thoughts?

TWiStErRob:
I think the missing link is a `hamcrest-junit` artifact. I'm all for decoupling the two libs, but it's a pretty common combination to have a dedicated artifact for supporting the combo.

sbrannen:
> In the light of that it would probably be most convenient for users if we undeprecated `Assume.assumeThat()` and changed its implementation to use `describeMismatch()`. It's not as easy to add it to Hamcrest because it requires and exception from JUnit.

I agree that undeprecating it is probably the best option since there is nothing else (supported) that you can refer users to.

sbrannen:
In summary, somebody needs to review **all** deprecations regarding Hamcrest and ensure that the "deprecation message" does not advise users to use something that is itself already deprecated and/or no longer available.

tumbarumba:
Hi @TWiStErRob, I'm just looking at your issue<sup>&sect;</sup>. As mentioned previously in this thread, the `hamcrest-junit` artifact is a completely unsupported experiment in restructuring the Hamcrest project. One of the previous maintainers was trying different things, and probably shouldn't have published that to Maven Central. Nat publishd a blog about this at http://natpryce.com/articles/000806.html, and the code can be found at https://github.com/hamcrest/hamcrest-junit.

With the release of JUnit 5, and the removal of the tight dependencies between JUnit and Hamcrest, I'm not sure that `hamcrest-junit` is required any more, though I appreciate that there will be people still using JUnit 4 for the next infinity or so years.

Looking at your Gradle example above, it looks much more complex than it needs to be. The insight that might help is that `hamcrest-2.1` is the new version, perfect for anyone using JUnit 5 or above. However, if you are using JUnit 4 (or anything else that has an existing transitive dependency on Hamcrest 1.3), the Hamcrest published 2 other artifacts to simplify upgrading: `hamcrest-core-2.1` and `hamcrest-library.2.1`. You can use these *instead of* `hamcrest-2.1` to automatically upgrade. This is documented at http://hamcrest.org/JavaHamcrest/distributables#upgrading-from-hamcrest-1x, though I appeciate it's a subtle thing. In your case, I think the following Gradle snippet would be sufficient (I have not tested this):

```groovy
dependencies { // test
	testImplementation "junit:junit:4.13-beta-3"
	testImplementation "org.hamcrest:hamcrest-core:2.1"
	testImplementation "org.hamcrest:hamcrest-junit:2.0.0.0"
}
``` 

<sup>&sect;</sup> Well, not quite *just*. I was looking at this a few months ago, and started writing some notes, and got distracted by life, the universe and everything. Trying to pick up the dropped threads again now. Apologies for being so slow.

TWiStErRob:
It's complex because we need a clean classpath. For example Android UI tests can't handle duplicate classes (dex). `org.hamcrest:hamcrest:2.1` and `org.hamcrest:java-hamcrest:2.0.0.0` are both containing the exact same package (`org.hamcrest.*`), but they're obviously not the same implementation. Even on JVM there's no guarantee which of the class files will be picked up while running the test. Cannot remove `java-hamcrest` because that's the only lib that contains `assumeThat` (one that's not deprecated anyway). There are also two `assertThat`s so IDE experience is not great.
```java
org.hamcrest.MatcherAssert.assertThat
org.hamcrest.junit.MatcherAssert.assertThat
```
Tip: run `gradle dependencies --configuration testCompileClasspath` on the dependencies you wrote.

---

To sum up, what I would love to see:
 * simple consumption of JUnit 4 and Hamcrest (as simple as you told above)
 * no duplicate classes on classpath
 * `assumeThat` and `assertThat` both available as a non-deprecated functions
 * no duplicate functions available on classpath (e.g. `assertThat` I listed above)
 * deprecations point to the right place when using the latest library version of JUnit 4 (4.13+) and Hamcrest (2.1+).

I think one way would be to publish a `hamcrest-junit:2.1` that would be just an alias to `hamcrest:2.1` and contain `assumeThat`?!

tumbarumba:
Hi @TWiStErRob, what you're asking for cannot be done

`assumeThat` has a dependency on `org.junit.AssumptionViolatedException`. The base Hamcrest library will never have any transitive dependencies, and we definitely won't add in a dependency on JUnit 4, which would result in a circular dependency.

`hamcrest-junit` was an unsupported experiment, and we don't intend to continue support in the future - though if someone wants to issue a pull request to update it, someone would definitely have a look.

My suggestion: don't use `hamcrest-junit`. I you really want the `assumeThat()` method, the  [`MatcherAssume`](https://github.com/hamcrest/hamcrest-junit/blob/master/src/main/java/org/hamcrest/junit/MatcherAssume.java) implementation is relatively simple. You could implement a replacement yourself.

marcphilipp:
Closed in favor of #1627.

PR #1610, opened 2019-05-17
ValidateWith custom use bug

SixPenny:
When I write a test to test `ValidateWith`, it didn't work. I found that there is a bug in file`src/main/java/org/junit/validator/AnnotationsValidator.java`.


panchenko:
`ValidateWith` is documented as 

  > Allows for an {@link AnnotationValidator} to be attached to an **annotation**.

It seems like the actual bug is missing `@Target(ElementType.ANNOTATION_TYPE)` on it.

An own annotation should be defined first:
```
@Retention(RetentionPolicy.RUNTIME)
@ValidateWith(ExampleAnnotationValidator.class)
public @interface ExampleAnnotationWithValidator {
}
```

SixPenny:
Yes. My commit isn't the real bug. I didn't see the test `AnnotationsValidatorTest`. It has a demonstration of use of `ValidateWith`. You can close this issue.

PR #1615, opened 2019-05-31
Undeprecate RuleChain

vRallev:
While the order in tests can be naturally specified with the order annotation attribute, RuleChain serves a second purpose as a composite TestRule.


stefanbirkner:
I created another PR #1617 that rewrites more of the Javadoc so that the ordering is not mentioned as primary use case anymore.

marcphilipp:
Closed in favor of #1617.

PR #1617, opened 2019-06-04
Rewrite Javadoc of RuleChain

### LINKED ISSUE #1614: Provide a `CompositeTestRule`
`RuleChain` has been deprecated in version 4.13 and the replacement is to use the order attribute in the `@Rule` annotation. But `RuleChain` also served as a composite test rule and there is no replacement for that. Let's say I have a method like this:
```java
public TestRule myRules() { .. }
```
Previously I was returning a `RuleChain`, that combined multiple rules. There is no replacement for this purpose.

Yes, it would be easy to create my own test rule. But I feel like there should be something available in the framework. I don't know how much it would differ from `RuleChain`, though.

stefanbirkner:
The purpose of this rule is no longer the ordering of rules but it can still be used for composing rules into a single rule. Therefore we remove the deprecation.

Fixes #1614.

### DIFF
--- a/src/main/java/org/junit/rules/RuleChain.java
+++ b/src/main/java/org/junit/rules/RuleChain.java
@@ -4,26 +4,34 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
+import org.junit.Rule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 
 /**
- * The RuleChain rule allows ordering of TestRules. You create a
+ * The {@code RuleChain} can be used for creating composite rules. You create a
  * {@code RuleChain} with {@link #outerRule(TestRule)} and subsequent calls of
  * {@link #around(TestRule)}:
  *
  * <pre>
- * public class UseRuleChain {
- * 	&#064;Rule
- * 	public RuleChain chain= RuleChain
- * 	                       .outerRule(new LoggingRule("outer rule"))
- * 	                       .around(new LoggingRule("middle rule"))
- * 	                       .around(new LoggingRule("inner rule"));
+ * public abstract class CompositeRules {
+ *   public static TestRule extendedLogging() {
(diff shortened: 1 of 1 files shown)


src/main/java/org/junit/rules/RuleChain.java
```diff
@@ -4,26 +4,35 @@
 import java.util.Collections;
 import java.util.List;
 
+import org.junit.Rule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 
 /**
- * The RuleChain rule allows ordering of TestRules. You create a
+ *
```
marcphilipp: nit: extraneous blank line

src/main/java/org/junit/rules/RuleChain.java
```diff
@@ -37,28 +46,14 @@
...
- *                          .around(new LoggingRule("inner rule"));
- * </pre>
+ * In older versions of JUnit (before 4.13) {@code RuleChain} was used for
+ * ordering rules. We recommend to not use it for this purpose anymore. You can
+ * use the attribute {@code order} of the annotation {@link Rule#order() Rule}
+ * or {@link org.junit.ClassRule#order() ClassRule} for ordering rules.
  *
- * @deprecated Since 4.13 ordering of rules can be naturally specified with an annotation attribute.
- * @see org.junit.Rule#order()
- * @see org.junit.ClassRule#order()
```
marcphilipp: :o: Shall we keep the two `@see` links?

src/main/java/org/junit/rules/RuleChain.java
stefanbirkner: fixed

src/main/java/org/junit/rules/RuleChain.java
stefanbirkner: I restored the links.

PR #1625, opened 2019-09-03
Adding new method needed on API for cucumber-jvm integration

vtatai:
Cucumber-jvm needs just an extra method added so that we can add annotations to the descriptions.

For more info please check https://github.com/cucumber/cucumber-jvm/pull/1761

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: java
+dist: trusty
 
 script: ./mvnw verify javadoc:javadoc site:site
 
--- a/src/main/java/org/junit/runner/Description.java
+++ b/src/main/java/org/junit/runner/Description.java
@@ -110,6 +110,18 @@ public class Description implements Serializable {
         return new Description(null, formatDisplayName(name, className), uniqueId);
     }
 
+    /**
+     * Create a <code>Description</code> of a single test named <code>name</code> in the class <code>clazz</code>.
+     * Generally, this will be a leaf <code>Description</code>.
+     *
+     * @param name the name of the test (a method name for test annotated with {@link org.junit.Test})
+     * @return a <code>Description</code> named <code>name</code>
+     */
+    public static Description createTestDescription(String className, String name, Serializable uniqueId,
+                                                    Annotation... annotations) {
+        return new Description(null, formatDisplayName(name, className), uniqueId, annotations);
+    }
+
     private static String formatDisplayName(String name, String className) {
         return String.format("%s(%s)", name, className);
     }
--- a/src/main/java/org/junit/runner/Description.java
+++ b/src/main/java/org/junit/runner/Description.java
@@ -114,7 +114,10 @@ public class Description implements Serializable {
      * Create a <code>Description</code> of a single test named <code>name</code> in the class <code>clazz</code>.
      * Generally, this will be a leaf <code>Description</code>.
      *
+     * @param className the class name of the test
      * @param name the name of the test (a method name for test annotated with {@link org.junit.Test})
+     * @param uniqueId an arbitrary object used to define uniqueness (in {@link #equals(Object)}
+     * @param annotations meta-data about the test, for downstream interpreters
      * @return a <code>Description</code> named <code>name</code>
      */
     public static Description createTestDescription(String className, String name, Serializable uniqueId,
--- a/src/test/java/org/junit/tests/experimental/max/DescriptionTest.java
+++ b/src/test/java/org/junit/tests/experimental/max/DescriptionTest.java
@@ -69,7 +69,7 @@ public class DescriptionTest {
     @Test
     public void parseClassNameAndMethodUniqueIdWithAnnotations() {
         Description description = Description.createTestDescription("not a class name", "aTestMethod", 123,
-            DescriptionTest.class.getAnnotations());
+            this.getClass().getAnnotations());
 
         assertThat(description.getClassName(), equalTo("not a class name"));
         assertThat(description.getMethodName(), equalTo("aTestMethod"));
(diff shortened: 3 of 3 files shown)


vtatai:
This is failing due to unrelated changes. For instance on Java 6:
```
$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install openjdk-6-jdk
Reading package lists...
Building dependency tree...
Reading state information...
Package openjdk-6-jdk is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  apt:i386 apt
E: Package 'openjdk-6-jdk' has no installation candidate
apt-get.diagnostics
apt-get install failed
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install openjdk-6-jdk" failed and exited with 100 during .
```
Last time tests passed was 3 months ago, so my guess is something broke since then. I'm not sure how to fix this, but will take a look at the Travis config files.

.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: java
+dist: trusty
```
vtatai: Had to force trusty since the packages specified below do not exist on more recent releases. Probably needs another ticket to upgrade the integration to use xenial (default) or bionic.

marcphilipp:
@kcooney I think this looks harmless enough that we could include it in 4.13. WDYT?

src/main/java/org/junit/runner/Description.java
```diff
@@ -110,6 +110,18 @@ public static Description createTestDescription(String className, String name, S
         return new Description(null, formatDisplayName(name, className), uniqueId);
     }
 
+    /**
+     * Create a <code>Description</code> of a single test named <code>name</code> in the class <code>clazz</code>.
+     * Generally, this will be a leaf <code>Description</code>.
+     *
+     * @param name the name of the test (a method name for test annotated with {@link org.junit.Test})
```
marcphilipp: Please also document the other parameters.

vtatai: Thanks for the review, fixed!

mpkorstanje:
Would there be any objection to making the `ParentRunner.filter` method also apply the filter to it's own description? It would remove the need to pass the annotation from the annotated runner class to the individual test descriptions.

vtatai:
Yes that seems like an interesting approach, let me give it a shot.

marcphilipp:
I don't think we should change this in `ParentRunner`. But you could override `filter` in `Cucumber` and check there, couldn't you?

kcooney:
@mpkorstanje The reason why `ParentRunner.filter` doesn't apply the filter to it's own description is because that filtering would be done by the containing suite.

`JUnitCore.main` uses `Request.classes()` to create a suite `Request`. I believe most build systems would use `Request.classes()` as well, so having `ParentRunner.filter` not apply the filter generally doesn't cause too many problems.

It's possible there's some odd interaction with categories somewhere.

vtatai:
@marcphilipp do you see any issues if we have this code inside `Cucumber`?

```java
    @Override
    public void filter(Filter filter) throws NoTestsRemainException {
        if (filter.shouldRun(getDescription())) {
            super.filter(Filter.ALL);
            return;
        }
        super.filter(filter);
    }
```
Thanks for any feedback!

vtatai:
@marcphilipp / @kcooney : I've tried overriding the `filter` method on `Cucumber`, but could not find a way to make it work - the problem is that it works for included categories, but then excluded categories don't work. That is because the `Categories` filter looks at each `Description`'s child to determine if they can run, and since they are not annotated, they can. I can't see any way around that, other than dropping all the children, which we can't do. Any ideas?

kcooney:
@vtatai It sounds like there might be a bug in categories. Could you write a test to demonstrate the problem? Here are some of the current tests you can use for inspiration:

- https://github.com/junit-team/junit4/blob/master/src/test/java/org/junit/experimental/categories/CategoryTest.java#L155
- https://github.com/junit-team/junit4/blob/master/src/test/java/org/junit/experimental/categories/CategoryTest.java#L182

vtatai:
@kcooney so the issue was again with cucumber - junit integration. What happens is that the `Categories` checks the parent class for any annotations:

https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/experimental/categories/Categories.java#L280

Which lead me to pass the original parent tests class for the description:

https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/runner/Description.java#L109

but the problem is that `Description` on the line right below calls a `formatDisplayName` which messes up with the description for the test being run, by including the "parent" class name. So we're back where we started, which is the `Description` factory methods. Would it be ok to just make the `Description` constructor public? That would make integration so much easier, since `Description` is a closed class and it would be very difficult to cover all the usage scenarios in factory methods. What are your thoughts? Thanks!

wangminpku:
Dear vtatai (@vtatai  )and Junit team (@kcooney)(@marcphilipp )(@stefanbirkner )

We are working on automatically identifying and clustering the related pull requests for code review. We have found that [PR#451](https://github.com/junit-team/junit4/pull/451) and [PR#365](https://github.com/junit-team/junit4/pull/365) might be related to this one, because both of them focus on the same problem.

We would really appreciate it if you could help us to validate and give us some feedback if the information from [PR#451](https://github.com/junit-team/junit4/pull/451) and [PR#365](https://github.com/junit-team/junit4/pull/365) is helpful when you review the current PR.

vtatai:
Sorry for the delayed response, but I'd love to have this merged - it is very similar to PR#451. Not sure if there is any other validation needed?

kcooney:
@vtatai we could merge this, but note we currently do not currently have plans of releasing a new version of the 4.x codebase. You should probably explore a work-around on your side.

vtatai:
Thanks @kcooney, I'm closing this PR then.

PR #1627, opened 2019-09-26
Fix the Hamcrest deprecation messages

stefanbirkner:
I reverted all changes that relied on hamcrest-junit. Assert.assertThat can still be deprecated because MatcherAssert.assertThat can be used instead.


stefanbirkner:
@marcphilipp We are still using `Assert.assertThat` in the method `ExpectedException#handleException`. IMO we should use `MatcherAssert.assertThat` there, too. What do you think?

sbrannen:
I took a quick glance within the browser, and I think the changes look good, but I'll defer to @marcphilipp. 👍

marcphilipp:
TBH I'm not sure anymore that we should deprecate `Assert.assertThat`. I think we should just keep it. What good would it do for anyone to switch to `MatcherAssert` since we won't get rid of the Hamcrest dependency (in JUnit 4) after all.

stefanbirkner:
> What good would it do for anyone to switch to `MatcherAssert` since we won't get rid of the Hamcrest dependency (in JUnit 4) after all.
IMHO the only improvement is that it is clear that the assertion method is part of Hamcrest and not JUnit. Would be nice if that helps people to understand that they have to ask the Hamcrest people for support. I know this is whishful thinking :-) But the price is that people may have to change code in order to upgrade to JUnit 4.13 if some company policy doesn't allow to use deprecated things.

I'm in favor of deprecating assertThat but if you think that the issue about forcing people to upgrade code is more important, I'm fine with it.

marcphilipp:
I think there are two main reasons to deprecate an API: (1) so it can be removed eventually and (2) so point users at better APIs. None of them apply in this case since I don't think we'll ever remove `Assert.assertThat` and `MatcherAssert.assertThat` has exactly the same signature but is harder to remember (because all other assertions are in `Assert`).

@kcooney WDYT?

kcooney:
@marcphilipp not sure I agree with the "harder to remember". If you use Hamcrest, then you should expect to need to import Hamcrest classes (just like if you use Truth, you should expect to import Truth classes). One could argue that having the classes in `org.junit.Assert` confused new users into thinking it's part of JUnit.

I think we could use the deprecation to point people to possibly better APIs. We could have the deprecation comment point to a link on our site that mentions alternative assertion frameworks.

The price mentioned by @stefanbirkner ("change code in order to upgrade to JUnit 4.13") seems really low in this case, since it's a simple change of imports (and unlike with our questionable deprecation of `ExpectedException`, projects can update code away from the deprecated APIs in advance).

True, we will never get rid of Hamcrest, but I don't see the harm in the deprecation. Then again, my company most developers ignore deprecations :-/

stefanbirkner:
I second @kcooney's argument about MatcherAssert. Whenever I use Hamcrest I don't use org.unit.Assert anymore because then I use everything from Hamcrest and do only Hamcrest asserts.

marcphilipp:
Thanks, @stefanbirkner! 👍

PR #1632, opened 2019-10-18
Assume that TemporaryFolderUsageTest.newFolderWithGivenFolderThrowsIO…

DirkMahler:
org.junit.rules.TemporaryFolderUsageTest#newFolderWithGivenFolderThrowsIOExceptionWhenFolderCannotBeCreated fails on Windows systems as a folder cannot be set to read-only. This is a pre-condition for the test and it has been changed such that the test is skipped if this cannot be satisified.


marcphilipp:
@DirkMahler Thanks!

PR #1633, opened 2019-10-18
Only deprecate ExpectedException.none()

### LINKED ISSUE #1609: Undeprecate ExpectedException rule
After upgrading the Spring Framework to JUnit 4.13 beta 3, I noticed that `org.junit.rules.ExpectedException` is now `@Deprecated`, and this change generates a lot of warnings in a large code base like that.

Since 4.13 is the last intended release in the 4.x line, I do not think it makes sense to deprecate such a commonly used and supported feature.

If you keep the deprecation in place, I fear that it will only annoy thousands (millions?) of developers for no good reason.

marcphilipp:
Instead of deprecating the whole class, now only the factory method is
deprecated and points to `assertThrows()`. This will make it easier to
temporarily suppress the exception for codebases that use it
extensively by adding an intermediate factory method of their own.

Resolves #1609.

### DIFF
--- a/src/main/java/org/junit/rules/ExpectedException.java
+++ b/src/main/java/org/junit/rules/ExpectedException.java
@@ -106,18 +106,19 @@ import org.junit.runners.model.Statement;
  * exception. E.g. "Test doesn't throw %s." will fail with the error message
  * "Test doesn't throw an instance of foo.".
  *
- * @deprecated Since 4.13
- * {@link org.junit.Assert#assertThrows(Class, org.junit.function.ThrowingRunnable)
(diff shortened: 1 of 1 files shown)


sbrannen:
LGTM 👍

stefanbirkner:
Thanks.

stefanbirkner:
@marcphilipp Are you updating the release notes or should it do this part.

marcphilipp:
@stefanbirkner If you could that would be great, thanks! 👍

stefanbirkner:
Done.

marcphilipp:
Thanks! 👍

PR #1635, opened 2019-10-21
Made some variable names more consistent with the other parts.

euske:
Hello, we're developing an automated system that detects inconsistent variable names in a large software project. Our system checks if each variable name is consistent with other variables in the project in its usage pattern, and proposes correct candidates if inconsistency is detected. This is a part of academic research that we hope to publish soon, but as a part of the evaluation, we applied our systems to your projects and got a few interesting results. We carefully reviewed our system output and manually created a patch to correct a few variable names. We would be delighted if this patch is found to be useful. If you have a question or suggestion regarding this patch, we'd happily
answer. Thank you.

P.S. our patch is purely for readability purposes and does not change any functionality. A couple of issues that we've noticed were left untouched. For example, mixed use of variable names "msg" and "message" were fairly widespread, but we have only corrected a few notable instances to minimize our impact.

### DIFF
--- a/src/main/java/org/junit/runner/Description.java
+++ b/src/main/java/org/junit/runner/Description.java
@@ -222,11 +222,11 @@ public class Description implements Serializable {
         if (isTest()) {
             return 1;
         }
-        int result = 0;
+        int conut = 0;
         for (Description child : fChildren) {
-            result += child.testCount();
+            conut += child.testCount();
         }
-        return result;
+        return conut;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/main/java/org/junit/runner/Description.java
```diff
@@ -222,11 +222,11 @@ public int testCount() {
         if (isTest()) {
             return 1;
         }
-        int result = 0;
+        int conut = 0;
```
marcphilipp: `count`?

euske: Oops. That's a typo when I was manually copying the system output. Sorry.

The use of count was fairly common as seen in countTestCases() in junit/framework/TestSuite.java.

kcooney: @euske I was going to fix this myself, but I can't because your pull request was made on your master branch. Please fix this and send us a new pull. This should work:

```
$ git branch -f master ce0d7c7^
$ git checkout -b consisent-variable-names ce0d7c7
$ git push --set-upstream origin consisent-variable-names 
```

kcooney: You also may want to change your `master` branch `main`. You can follow the "What Your Teammates Have to Do" instructions at https://www.git-tower.com/learn/git/faq/git-rename-master-to-main/

I suggest fixing your master branch (following my previous instructions) first

PR #1638, opened 2019-11-19
Never reorder classes annotated with @FixMethodOrder

### LINKED ISSUE #1637: sortWith() and orderWith() should respect @FixMethodOrder
When you order a suite using `Request.sortWith()` all of the runners that implement `Sortable` (like `BlockJUnit4ClassRunner`) will reorder their methods using the passed-in `Comparator`, even if the ordering was specified with `@FixMethodOrder`. The same holds true for `Request.orderWith()`. JUnit should instead not reorder methods for classes annotated with `@FixMethodOrder`

**Background:**

Someone at my company was trying to implement code to shuffle method orders to catch tests that produced different pass/fail results based on ordering (these shuffling tests are useful, for instance, if you want to ensure that you could always add `@Ignore` to a broken test without causing otherwise-passing tests to start failing). We have a very large code base, and some tests are annotated with `@FixMethodOrder` because we know that they only pass if run in a specific order. We have enough of these that it will take a significant amount of time to fix them, especially for parts of the code where the original authors have left the company.

We had to write custom shuffling code, because the code in JUnit for ordering did not respect `@FixMethodOrder`. In some cases, our customized shuffling code has to bail out with a "Shuffling not supported for this suite" error message.

Can we please have `ParentRunner.sort(Sorter)` and `ParentRunner.order(Orderer)` do nothing if the Description indicates that the class was annotated with `@FixMethodOrder`?

I do realize that this would be a change in behavior from previous versions of JUnit. I don't think the change I'm proposing would cause tests to fail. The only way that could happen is if there is some code outside of JUnit that specifically orders the suite in an order, but if projects run into this problem when trying to migrate to 4.13, the easy work-around would be to remove the `@FixMethodOrder` annotation.

kcooney:
Fixes #1637

### DIFF
--- a/src/main/java/org/junit/runner/OrderWith.java
+++ b/src/main/java/org/junit/runner/OrderWith.java
@@ -7,6 +7,7 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 import org.junit.runner.manipulation.Ordering;
+import org.junit.validator.ValidateWith;
 
 /**
  * When a test class is annotated with <code>&#064;OrderWith</code> or extends a class annotated
@@ -18,6 +19,7 @@ import org.junit.runner.manipulation.Ordering;
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.TYPE)
 @Inherited
+@ValidateWith(OrderWithValidator.class)
 public @interface OrderWith {
     /**
      * Gets a class that extends {@link Ordering}. The class must have a public no-arg constructor.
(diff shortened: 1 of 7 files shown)


kcooney:
Thanks for the review, @marcphilipp 

I spit the commit into this pull into two (the first commit containing the changes to sorting/ordering, the second commit adding the validation check).

I suggest using "Create a merge commit" (and not "Squash and merge") for this pull but it's up to you.

@stefanbirkner any concerns with these changes?

marcphilipp:
I'd use "Rebase and merge" which would also preserve both commits but avoid the additional merge commit.

marcphilipp:
Thanks, @kcooney! Could you please add an entry to the release notes?

kcooney:
@marcphilipp Done.

PR #1639, opened 2019-11-27
Reordering modifiers, clean up imports, and deleting commented-out code

ullenius:
Minor clean up. Does **3** things:

- Removing commented out lines of code.

> "It makes me crazy to see stretches of code that are commented out. 
> Who knows how old it is?
> Who knows whether or not it's meaningful? 
> Yet no one will delete it because everyone assumes someone else needs it or has plans for it. Commented-out code is an abomination." 

-Clean Code (chapter 17) by [Uncle Bob](https://en.wikipedia.org/wiki/Robert_C._Martin)

- Cleaning up unused imports and deleting imports from the same package.

- Reordering modifiers to comply with the Java Language Specification:

> The Java Language Specification recommends listing modifiers in the following order:
> 
> 1. Annotations
> 2. public
> 3. protected
> 4. private
> 5. abstract
> 6. static
> 7. final
> 8. transient
> 9. volatile
> 10. synchronized
> 11. native
> 12. strictfp

### DIFF
--- a/src/main/java/junit/extensions/ActiveTestSuite.java
+++ b/src/main/java/junit/extensions/ActiveTestSuite.java
@@ -43,7 +43,6 @@ public class ActiveTestSuite extends TestSuite {
             public void run() {
                 try {
                     // inlined due to limitation in VA/Java
-                    //ActiveTestSuite.super.runTest(test, result);
                     test.run(result);
                 } finally {
                     ActiveTestSuite.this.runFinished();
--- a/src/main/java/junit/extensions/ActiveTestSuite.java
+++ b/src/main/java/junit/extensions/ActiveTestSuite.java
@@ -1,6 +1,9 @@
 package junit.extensions;
 
-import junit.framework.*;
+import junit.framework.Test;
+import junit.framework.TestCase;
+import junit.framework.TestResult;
+import junit.framework.TestSuite;
 
 /**
  * A TestSuite for active Tests. It runs each
--- a/src/main/java/junit/extensions/ActiveTestSuite.java
+++ b/src/main/java/junit/extensions/ActiveTestSuite.java
@@ -1,9 +1,6 @@
 package junit.extensions;
 
-import junit.framework.Test;
-import junit.framework.TestCase;
-import junit.framework.TestResult;
-import junit.framework.TestSuite;
+import junit.framework.*;
 
 /**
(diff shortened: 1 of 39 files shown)


stefanbirkner (approved):

marcphilipp (approved):

stefanbirkner:
The change in `FailOnTimeoutOutTest` is already part of 1254795902e56eb545e1fe696827924bc5e82b66. Removing it from the commits in this PR will solve the conflict. I can do this when merging the PR.

src/main/java/junit/extensions/ActiveTestSuite.java
```diff
@@ -43,7 +43,6 @@ public void runTest(final Test test, final TestResult result) {
             public void run() {
                 try {
                     // inlined due to limitation in VA/Java
-                    //ActiveTestSuite.super.runTest(test, result);
```
kcooney, line 46: I think we should add this comment back, since i clarifies what was inlined.

kcooney (changes requested):

src/main/java/junit/extensions/ActiveTestSuite.java
ullenius: Fixed in 460e488

src/main/java/junit/extensions/ActiveTestSuite.java
```diff
@@ -1,9 +1,6 @@
 package junit.extensions;
 
-import junit.framework.Test;
-import junit.framework.TestCase;
-import junit.framework.TestResult;
-import junit.framework.TestSuite;
+import junit.framework.*;
```
kcooney, line 3: I prefer we avoid using globs in imports. I wouldn't go so far as to suggest to remove existing globs, but let's not add more.

ullenius: Reverted in 7b53777.

kcooney (approved):

marcphilipp:
@ullenius Could you please rebase on `main` to resolve the conflict?

ullenius:
> @ullenius Could you please rebase on `main` to resolve the conflict?

No problem. I removed `FailOnTimeoutOutTest` from the PR as @stefanbirkner suggested and rebased.

stefanbirkner:
I'm merging the PR.

PR #1646, opened 2020-01-29
Upgrade Maven to 3.2.5

marcphilipp:
Since Maven 3.1.1 by default uses HTTP instead of HTTPS for resolving
artifacts from Maven Central, which has recently been discontinued,
this commit now uses the latest release that runs on Java 6 and also
upgrades the Maven wrapper.


kcooney:
@marcphilipp sorry, I'm not very knowledgable about maven, so I can't give this a good review

stefanbirkner:
I found another solution that does not require a Maven upgrade: #1649. I prefer not to update Maven because I like to build locally with Java 5 in order to ensure that JUnit 4 keeps compatible with Java 5.

marcphilipp:
Superseded by #1649

PR #1649, opened 2020-02-07
Use Google's Maven Central mirror

stefanbirkner:
We are using Maven 3.1.1 which by default uses HTTP instead of HTTPS for resolving artifacts from Maven Central. Maven Central recently discontinued HTTP support. Therefore the build on Travis started failing. By using an HTTPS mirror of Maven Central the build on Travis will work again.

I chose Google's mirror because Travis uses this mirror by default, too. I did not upgrade to a new version of Maven because there is no newer version with Java 5 support and it should be possible to build JUnit 4 with Java 5 so that we can easily ensure that it works with Java 5.

### DIFF
--- a/.travis.settings.xml
+++ b/.travis.settings.xml
@@ -1,4 +1,12 @@
 <settings>
+    <mirrors xmlns="http://maven.apache.org/SETTINGS/1.1.0">
+        <mirror>
+            <mirrorOf>central</mirrorOf>
+            <name>GCS Maven Central mirror</name>
+            <url>https://maven-central.storage-download.googleapis.com/maven2/</url>
+            <id>google-maven-central</id>
+        </mirror>
+    </mirrors>
     <servers>
         <server>
             <id>junit-snapshot-repo</id>


stefanbirkner:
This is a successful run with the changes: https://travis-ci.org/junit-team/junit4/jobs/647502589 That build also included a change of `.travis.yml` that enabled deployment for the `mirrors` branch.

marcphilipp:
We should also document that `~/.m2/settings.xml` has to be changed locally when using the Maven wrapper.

sormuras:
`http` ceased to work: https://travis-ci.org/junit-team/junit4/jobs/642256743

Let's merge this PR and switch to `https` where possible. In the root [pom.xml](https://github.com/junit-team/junit4/blob/master/pom.xml) are 12 `http` links...

.travis.settings.xml
```diff
@@ -1,4 +1,12 @@
 <settings>
+    <mirrors xmlns="http://maven.apache.org/SETTINGS/1.1.0">
```
sormuras: Is `https://` a valid protocol here, too? If yes, let's use it.

stefanbirkner: IIRC the namespace URI is *only* a String that is used for namespacing XML tags and attributes. It is not a link and therefore https is not important here and  not really a protocol but just part of the name. I think changing the namespace from "http://" to "https://" changes the namespace (but not cause any problem because it is not important). Therefore I would like to keep it like it is.

sormuras: Alright. Marking this conversation as resolved.

PR #1651, opened 2020-02-11
Theories: only get all data points from enum and boolean types if the…

### LINKED ISSUE #1648: `@FromDataPoints("misspelled") EnumType` shouldn't test all enum values
```
@RunWith(Theories.class)
public class NonExistentTheory {
  @DataPoints("foo") public static final ImmutableSet<MetaSyntacticVariable> FOOS =
      ImmutableSet.of(MetaSyntacticVariable.FOO);

  @Theory
  public void doWithTheFoo(@FromDataPoints("bar") MetaSyntacticVariable whatever) {
    System.err.println(whatever);
  }

  @Theory
  public void doWithTheBoolean(@FromDataPoints("bar") boolean whatever) {
    System.err.println(whatever);
  }

  enum MetaSyntacticVariable {
    FOO, BAR, BAZ
  }
}
```

It is surprising that these `@FromDataPoints`-annotated parameters receive the same values as if the annotation were omitted. This means that typos in data point names aren't flagged as errors.

If the `@FromDataPoints` is present, and nothing matches, it would be helpful to raise "Never found parameters that satisfied method assumptions.".

awturner:
…re is no @FromDataPoints annotation, or the named data points represent a non-empty set of the same type.

Fixes #1648


stefanbirkner (approved):

awturner:
Not familiar with the pull request flow. Do I now have to do something to merge this, or is that something somebody on the project will do?

src/main/java/org/junit/experimental/theories/internal/Assignments.java
```diff
@@ -85,14 +85,16 @@ public Assignments assignNext(PotentialAssignment source) {
...
+
+        FromDataPoints fromDataPoints = unassigned.getAnnotation(FromDataPoints.class);
+        if (fromDataPoints == null) {
+            if (paramType.isEnum()) {
+                return new EnumSupplier(paramType).getValueSources(unassigned);
+            } else if (paramType.equals(Boolean.class) || paramType.equals(boolean.class)) {
+                return new BooleanSupplier().getValueSources(unassigned);
+            }
         }
+        return emptyList();
```
stefanbirkner, line 97: For symmetry it would be nice to keep the `else`

```suggestion
        } else {
            return emptyList();
        }
```

stefanbirkner (dismissed):

src/main/java/org/junit/experimental/theories/internal/Assignments.java
```diff
@@ -85,11 +85,15 @@ public Assignments assignNext(PotentialAssignment source) {
...
-            return new BooleanSupplier().getValueSources(unassigned);
+
+        FromDataPoints fromDataPoints = unassigned.getAnnotation(FromDataPoints.class);
+        if (fromDataPoints == null) {
+            if (paramType.isEnum()) {
+                return new EnumSupplier(paramType).getValueSources(unassigned);
+            } else if (paramType.equals(Boolean.class) || paramType.equals(boolean.class)) {
+                return new BooleanSupplier().getValueSources(unassigned);
+            }
+        }
```
awturner, line 96:
```suggestion
```

src/main/java/org/junit/experimental/theories/internal/Assignments.java
```diff
@@ -85,11 +85,14 @@ public Assignments assignNext(PotentialAssignment source) {
 
     private List<PotentialAssignment> generateAssignmentsFromTypeAlone(ParameterSignature unassigned) {
         Class<?> paramType = unassigned.getType();
-        
-        if (paramType.isEnum()) {
-            return new EnumSupplier(paramType).getValueSources(unassigned);  
-        } else if (paramType.equals(Boolean.class) || paramType.equals(boolean.class)) {
-            return new BooleanSupplier().getValueSources(unassigned);
+
+        FromDataPoints fromDataPoints = unassigned.getAnnotation(FromDataPoints.class);
```
marcphilipp, line 89: ❓ Shouldn't this check for `ParametersSuppliedBy` or, alternatively, shouldn't `SpecificDataPointsSupplier` throw an exception if it couldn't find a matching field?

awturner:
Then please can you fix this yourselves? Sorry, I just don't have enough knowledge of the code base.

src/main/java/org/junit/experimental/theories/internal/Assignments.java
J-Vin: I think so

KikeE36 (approved):

PR #1655, opened 2020-03-06
Document, that the method 'apply' is invoked.

polina-c:


src/main/java/org/junit/ClassRule.java
```diff
@@ -11,7 +11,9 @@
  * a subtype of {@link org.junit.rules.TestRule}.
  * <p>
  * The {@link org.junit.runners.model.Statement} passed
- * to the {@link org.junit.rules.TestRule} will run any {@link BeforeClass} methods,
+ * to the {@link org.junit.rules.TestRule} will run 
+ * the overridable method {@link org.junit.rules.TestRule.apply},
+ * any {@link BeforeClass} methods,
```
marcphilipp: The `Statement` passed to `apply()` will _not_ call `apply()`.

PR #1660, opened 2020-04-18
Deprecated since jdk9 replacing constructor instance of Double and Float

RahulNagekar:


marcphilipp:
Can't we simply rely on autoboxing?

RahulNagekar:
> Can't we simply rely on autoboxing?

Right, good point.

kcooney:
@marcphilipp @RahulNagekar  I personally prefer not using autoboxing here. Having an explicit conversion here makes it clear to future readers that a conversion is happening.

stefanbirkner:
I'm also in favour for Double.valueOf so that the code is the same like in JUnit 4's `Assert`.

RahulNagekar:
I currently not sure if autoboxing would be more beneficial or valueOf (performance-wise).
As valueOf is used at many places in this file, I say keep it the same to preserve consistency in code.
:-)

PR #1669, opened 2020-07-15
Make FrameworkField ctor public. Fixes #1668

### LINKED ISSUE #1668: Make ctor FrameworkField(Field) public?
Hello,

In [junit-quickcheck](https://github.com/pholser/junit-quickcheck) I found myself wanting to make both `FrameworkMethod`s and `FrameworkField`s. `FrameworkMethod`'s constructor is public; `FrameworkField`'s constructor is package-private. I tried to be sneaky and make the class that uses the `FrameworkField` constructor a package cohabitant of `FrameworkField`. Now, however, some of my users are reporting an error when running junit-quickcheck tests, along the lines of [this issue](https://github.com/pholser/junit-quickcheck/issues/267).

In the short term, I'd try to clone `FrameworkField`, put it in a junit-quickcheck package, and move on -- except that `FrameworkField` contains package-private abstract methods that I won't be able to override (better for those to be marked `protected`?)

Longer term -- would it be possible to promote `FrameworkField`'s actor to public?

Thanks for your consideration.

pholser:
Prior to this change, custom runners could make FrameworkMethod
instances, but not FrameworkField instances. This small change
allows for both now, because FrameworkFields constructor has been
promoted to public from package-private.

### DIFF
--- a/src/main/java/junit/runner/Version.java
+++ b/src/main/java/junit/runner/Version.java
@@ -9,7 +9,7 @@ public class Version {
 	}
 
 	public static String id() {
-		return "4.13-SNAPSHOT";
+		return "4.14-SNAPSHOT";
 	}
 	
 	public static void main(String[] args) {
--- a/src/main/java/org/junit/runners/model/FrameworkField.java
+++ b/src/main/java/org/junit/runners/model/FrameworkField.java
@@ -15,7 +15,12 @@ import org.junit.runners.BlockJUnit4ClassRunner;
 public class FrameworkField extends FrameworkMember<FrameworkField> {
     private final Field field;
 
-    FrameworkField(Field field) {
+    /**
+     * Returns a new {@code FrameworkField} for {@code field}.
+     *
+     * <p>Access relaxed to {@code public} since version 4.14.
+     */
+    public FrameworkField(Field field) {
         if (field == null) {
             throw new NullPointerException(
                     "FrameworkField cannot be created without an underlying field.");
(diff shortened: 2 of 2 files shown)


src/main/java/junit/runner/Version.java
```diff
@@ -9,7 +9,7 @@ private Version() {
 	}
 
 	public static String id() {
-		return "4.13-SNAPSHOT";
+		return "4.14-SNAPSHOT";
```
kcooney, line 12: Do you mind moving this change it's own commit (before the other changes)?

src/main/java/org/junit/runners/model/FrameworkField.java
```diff
@@ -15,7 +14,7 @@
 public class FrameworkField extends FrameworkMember<FrameworkField> {
     private final Field field;
 
-    FrameworkField(Field field) {
+    public FrameworkField(Field field) {
```
kcooney, line 17: Let's add JavaDoc and indicate that this constructor is public as of JUnit 4.14.

src/main/java/junit/runner/Version.java
pholser: Sure thing! Done.

src/main/java/org/junit/runners/model/FrameworkField.java
kcooney: @pholser you somehow lost the "change to public" part of the change :-)

I added a commit with suggested edits. If you agree, please squash with the second commit.

pholser:
Whoops. Done. Thanks again!

joe198 (commented):
ok

PR #1673, opened 2020-09-05
Fix tests for floating point assertions

OlivierBlanvillain:
The issue is that both

```java
assertEquals(1.0, 2.0, 0.0);
```

and

```java
fail();
```

throw an `AssertionFailedError`, so these tests were actually not testing anything...


marcphilipp (approved):
Good catch! 👍

stefanbirkner (approved):

PR #1687, opened 2021-01-01
Mark ThreadGroups created by FailOnTimeout as daemon groups

### LINKED ISSUE #1652: Timeout ThreadGroups should not be destroyed
#### Environment:

Eclipse: Version: 2019-09 R (4.13.0), Build id: 20190917-1200
EclEmma 3.1.2.201903112331	org.eclipse.eclemma.feature.feature.group	Eclipse EclEmma
JUnit 4.13
Windows 10 x64
Java 1.8u241

#### Issue:

I've run into an odd error while running coverage on an unit test having some methods use Google Guava's CacheBuilder in Eclipse. The coverage fails with the error code 5013 and error report indicating java.net.SocketException: Connection reset. If I remove these methods, the coverage succeeds.

If I revert back to JUnit 4.12, both the test file and the individual test succeeds in producing the coverage data.

Example method:
https://github.com/ReactiveX/RxJava/blob/5f6aafcbaa97f330ad2007f4ad02c5e33db8914b/src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableGroupByTest.java#L1917

This is where the cache is created:
https://github.com/ReactiveX/RxJava/blob/5f6aafcbaa97f330ad2007f4ad02c5e33db8914b/src/test/java/io/reactivex/rxjava3/internal/operators/flowable/FlowableGroupByTest.java#L2194

#### Additional details:

- If I run "gradlew test jacocoTestReport" on the entire project, it succeeds in creating the report HTML apparently.
- If I run "Coverage As > JUnit test" on the entire project from Eclipse, it fails with the same 5013 error.
- If I run any other test class/package/method in the project, the coverage is created successfully.

Please advise.

kcooney:
Fixes #1652

### DIFF
--- a/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
+++ b/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
@@ -121,27 +121,20 @@ public class FailOnTimeout extends Statement {
         CallableStatement callable = new CallableStatement();
         FutureTask<Throwable> task = new FutureTask<Throwable>(callable);
         ThreadGroup threadGroup = new ThreadGroup("FailOnTimeoutGroup");
-        Thread thread = new Thread(threadGroup, task, "Time-limited test");
-        try {
-            thread.setDaemon(true);
-            thread.start();
-            callable.awaitStarted();
-            Throwable throwable = getResult(task, thread);
-            if (throwable != null) {
-                throw throwable;
-            }
-        } finally {
+        if (!threadGroup.isDaemon()) {
             try {
-                thread.join(1);
-            } catch (InterruptedException e) {
-                Thread.currentThread().interrupt();
-            }
-            try {
-                threadGroup.destroy();
-            } catch (IllegalThreadStateException e) {
-                // If a thread from the group is still alive, the ThreadGroup cannot be destroyed.
+                threadGroup.setDaemon(true);
+            } catch (SecurityException e) {
                 // Swallow the exception to keep the same behavior prior to this change.
             }
+        }
+        Thread thread = new Thread(threadGroup, task, "Time-limited test");
+        thread.setDaemon(true);
+        thread.start();
+        callable.awaitStarted();
+        Throwable throwable = getResult(task, thread);
+        if (throwable != null) {
+            throw throwable;
         }
     }
 
--- a/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
+++ b/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
@@ -125,7 +125,7 @@ public class FailOnTimeout extends Statement {
             try {
                 threadGroup.setDaemon(true);
             } catch (SecurityException e) {
-                // Swallow the exception to keep the same behavior prior to this change.
+                // Swallow the exception to keep the same behavior as in JUnit 4.12.
             }
         }
         Thread thread = new Thread(threadGroup, task, "Time-limited test");
--- a/src/test/java/org/junit/internal/runners/statements/FailOnTimeoutTest.java
+++ b/src/test/java/org/junit/internal/runners/statements/FailOnTimeoutTest.java
@@ -212,7 +212,7 @@ public class FailOnTimeoutTest {
     }
 
     @Test
-    public void threadGroupIsDaemon() throws Throwable {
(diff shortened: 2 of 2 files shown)


marcphilipp (approved):

src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
```diff
@@ -121,28 +121,21 @@ public void evaluate() throws Throwable {
...
-            } catch (InterruptedException e) {
-                Thread.currentThread().interrupt();
-            }
-            try {
-                threadGroup.destroy();
-            } catch (IllegalThreadStateException e) {
-                // If a thread from the group is still alive, the ThreadGroup cannot be destroyed.
+                threadGroup.setDaemon(true);
+            } catch (SecurityException e) {
                 // Swallow the exception to keep the same behavior prior to this change.
```
stefanbirkner, line 128: The term "this change" is a little bit unspecific now. Can you change it to
```suggestion
                // Swallow the exception to keep the same behavior as in JUnit 4.12.
```

stefanbirkner (changes requested):

kcooney:
FYI: I updated the test to verify that the `ThreadGroup` is destroyed (and verified by hand that if I don't mark it as a daemon, the `ThreadGroup` is not destroyed).

kcooney:
So apparently marking the `ThreadGroup` as a daemon will not result in it being destroyed in all versions of the JDK (at least not immediately after the only thread in the group completes). I'm guessing that in some versions of the JDK it would be destroyed after a delay.

marcphilipp:
Which versions did you try?

kcooney:
It was a race condition. The test needed to join on the inner thread before checking the `ThreadGroup` status.

Lots of changes to squash, but leaving as-is for now so you can all see the changes.

stefanbirkner:
@kcooney Can you please have a look at #1690. I changed the test that checks that the thread group is destroyed. IMO it is now checking the behavior that we are looking for. What do you think? It you are ok with the new test then your PR can be reduced to the change in FailOnTimeout and does not need to change the test.

kcooney:
@stefanbirkner the test in this pull now verifies that the thread group is destroyed. It has a few unnecessary assertions (I was trying to determine why it passed on my machine but not in the CI build). I would be happy to remove the now-unnecessary extra assertions.

Even so, the test in this pull seems a bit more straight-forward than the one in yours. WDYT?

stefanbirkner:
I tend to like your approach more. If I change my opinion I can still create another PR.

stefanbirkner (approved):

stefanbirkner:
I changed the explanation in #1690 and removed changing the test.

src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
```diff
@@ -121,28 +121,21 @@ public void evaluate() throws Throwable {
...
-                thread.join(1);
-            } catch (InterruptedException e) {
-                Thread.currentThread().interrupt();
-            }
-            try {
-                threadGroup.destroy();
-            } catch (IllegalThreadStateException e) {
-                // If a thread from the group is still alive, the ThreadGroup cannot be destroyed.
-                // Swallow the exception to keep the same behavior prior to this change.
+                threadGroup.setDaemon(true);
```
j256, line 126: There should be a comment here pointing to this pull request and explaining why they should be daemon.

stefanbirkner: #1691 adds an explanation

j256: Certainly does.  Thanks.

Tuneabq (commented):
👍

src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
shivam11-u: How

kcooney: @shivam11-u I'm not quite sure what your question is, but if you are wondering how this change "keeps the same behavior" code here is catching `SecurityException` it's because it's possible that JUnit is running in an environment

PR #1689, opened 2021-01-02
Fix test "threadGroupNotLeaked"

stefanbirkner:
The FailOnTimeout statement was not executed because the Runnable's
"run" method was not called.


marcphilipp (approved):

src/test/java/org/junit/internal/runners/statements/FailOnTimeoutTest.java
```diff
@@ -212,7 +212,7 @@ private void notTheRealCauseOfTheTimeout() {
     public void threadGroupNotLeaked() throws Throwable {
```
kcooney, line 212: This test is being replaced in pull 1687

Based on the CI run, I suspect that it's flaky anyway.

stefanbirkner:
Obsolete because test changed in #1687.

PR #1690, opened 2021-01-02
Explain why the thread group is destroyed

stefanbirkner:
This information is currently only available in GitHub. It should be
close to the code so that the reader can understand why we destroy the
thread group.

### DIFF
--- a/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
+++ b/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
@@ -121,6 +121,12 @@ public class FailOnTimeout extends Statement {
         CallableStatement callable = new CallableStatement();
         FutureTask<Throwable> task = new FutureTask<Throwable>(callable);
         ThreadGroup threadGroup = new ThreadGroup("FailOnTimeoutGroup");
+        // The thread group must be a daemon thread group so that it is
+        // destroyed when the time-limiting thread finishes. Only when the
+        // thread group is destroyed its parent thread group can be destroyed,
+        // too. This is necessary because data may be stored in the parent
+        // thread group's ThreadGroupContext and this data remains until the
+        // parent thread group is destroyed.
         if (!threadGroup.isDaemon()) {
             try {
                 threadGroup.setDaemon(true);


src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
```diff
@@ -121,6 +121,12 @@ public void evaluate() throws Throwable {
         CallableStatement callable = new CallableStatement();
         FutureTask<Throwable> task = new FutureTask<Throwable>(callable);
         ThreadGroup threadGroup = new ThreadGroup("FailOnTimeoutGroup");
+        // The thread group must be a daemon thread group so that it is
+        // destroyed when the time-limiting thread finishes. Only when the
+        // thread group is destroyed its parent thread group can be destroyed,
```
kcooney, line 126: I'm a bit puzzled by this explanation.

The parent thread group isn't destroyed by any of the code here. The parent thread group will likely live for the life of the thread that ran this Statement, which is likely the life of the test run. 

Looking at the original pull, it sounds like `ThreadGroupContext` is likely `java.beans.ThreadGroupContext`, and Spring uses that to store data, using the identity hash code of the current `ThreadGroup` as a key. So I suggest:

"By ensuring the ThreadGroup is destroyed, any data associated with the ThreadGroup (ex: via java.beans.ThreadGroupContext) is destroyed."

Presumably, any tests not using a timeout would have data stored using the ThreadGroup from the main JUnit thread as the key for the context, and that would stay around until all tests complete. Perhaps that isn't a problem because the data doesn't keep growing as more tests are run.

stefanbirkner: Thanks for digging deeper into it. #1691 makes this PR obsolete, isn't it?

stefanbirkner:
I'm closing this PR because #1691 is changing the code and adds comments with a better explanation.

src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
kcooney: @stefanbirkner I tried to capture the gist of the comments that you added here in my pull. Please let me know if you want me to say more.

Thanks for the reminder to document why the code does what it does

PR #1691, opened 2021-01-03
Only create ThreadGroups if FailOnTimeout.lookForStuckThread is true.

kcooney:
This reduces the differences (where possible) when tests are run with a timeout.
Creating a ThreadGroup can cause noticeable differences, for example in code
that uses java.beans.ThreadGroupContext.

### DIFF
--- a/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
+++ b/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
@@ -120,25 +120,7 @@ public class FailOnTimeout extends Statement {
     public void evaluate() throws Throwable {
         CallableStatement callable = new CallableStatement();
         FutureTask<Throwable> task = new FutureTask<Throwable>(callable);
-
-        ThreadGroup threadGroup = null;
-        if (lookForStuckThread) {
-          // Create the thread in a new ThreadGroup, so if the time-limited thread
-          // becomes stuck, getStuckThread() can find the thread likely to be the
-          // culprit.
-          threadGroup = new ThreadGroup("FailOnTimeoutGroup");
-          if (!threadGroup.isDaemon()) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
```diff
@@ -120,13 +120,24 @@ public FailOnTimeout build(Statement statement) {
...
+              // destroyed after the time-limited thread completes. By ensuring the
+              // ThreadGroup is destroyed, any data associated with the ThreadGroup
+              // (ex: via java.beans.ThreadGroupContext) is destroyed.
+              try {
+                  threadGroup.setDaemon(true);
+              } catch (SecurityException e) {
+                  // Swallow the exception to keep the same behavior as in JUnit 4.12.
+              }
+          }
         }
```
marcphilipp, line 141: Could we extract this into a method, please?

marcphilipp (approved):
Good idea! 👍

stefanbirkner (approved):
Nice catch and thanks for explaining the details in the comments.

src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java
kcooney: Done.

marcphilipp (approved):
Thanks! 👍

stefanbirkner (approved):

kcooney:
I updated this pull so that the release notes documents the change. @marcphilipp and/or @stefanbirkner would you please review the release notes?

Thanks!

marcphilipp (approved):

stefanbirkner (approved):

blaquezzz (approved):

PR #1698, opened 2021-02-17
Create SECURITY.md

Cu2y:


Cu2y (commented):
please review for updated conf.

marcphilipp:
I don't think we should duplicate this information here as we don't have the bandwidth to keep it up to date.

PR #1702, opened 2021-03-02
Don't convert assumption failures to errors in ErrorCollector.addErro…

prbprbprb:
…r().

Provides clearer results when calling sub-methods or when wrapping tests via a Rule.

### DIFF
--- a/src/main/java/org/junit/rules/ErrorCollector.java
+++ b/src/main/java/org/junit/rules/ErrorCollector.java
@@ -43,19 +43,15 @@ public class ErrorCollector extends Verifier {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ZeeZeeMorin (approved):

stefanbirkner:
This PR would revert parts of 88f63daafb9eec076c09f532c9f0bd823a150be0 that fixed #1363.

stefanbirkner:
The code works as expected. From @kcooney's remarks in #1363 

>  The expressions passed into these methods should call production code, not test code, and AssumptionViolationException should generally be thrown by test code.

The Javadoc can be improved but currently we haven't planned to build another JUnit 4 release.

PR #1707, opened 2021-03-23
Remove some code smell

Nepsio:
I remove some code smell  as part of a school project

### DIFF
--- a/src/main/java/junit/runner/BaseTestRunner.java
+++ b/src/main/java/junit/runner/BaseTestRunner.java
@@ -154,7 +154,8 @@ public abstract class BaseTestRunner implements TestListener {
      */
     protected String processArguments(String[] args) {
         String suiteName = null;
-        for (int i = 0; i < args.length; i++) {
+        int i = 0;
+        while( i < args.length) {
             if (args[i].equals("-noloading")) {
                 setLoading(false);
             } else if (args[i].equals("-nofilterstack")) {
@@ -169,6 +170,7 @@ public abstract class BaseTestRunner implements TestListener {
             } else {
                 suiteName = args[i];
             }
+            i++;
         }
         return suiteName;
     }
@@ -297,7 +299,7 @@ public abstract class BaseTestRunner implements TestListener {
     }
 
     protected static boolean showStackRaw() {
-        return !getPreference("filterstack").equals("true") || fgFilterStack == false;
+        return !getPreference("filterstack").equals("true") || !fgFilterStack ;
     }
 
     static boolean filterLine(String line) {
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -603,23 +603,17 @@ public class Assert {
     private static boolean doubleIsDifferent(double d1, double d2, double delta) {
         if (Double.compare(d1, d2) == 0) {
             return false;
+        }else{
+            return  Math.abs(d1 - d2) > delta;
         }
-        if ((Math.abs(d1 - d2) <= delta)) {
-            return false;
-        }
-
-        return true;
     }
 
     private static boolean floatIsDifferent(float f1, float f2, float delta) {
         if (Float.compare(f1, f2) == 0) {
             return false;
+        }else {
+            return  Math.abs(f1 - f2) > delta;
         }
-        if ((Math.abs(f1 - f2) <= delta)) {
-            return false;
-        }
-
-        return true;
     }
 
     /**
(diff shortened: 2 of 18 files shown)


stefanbirkner:
@Nepsio Thank you for your contribution. JUnit 4 is a mature library and there is already a successor with JUnit 5. There is no plan for a new release of JUnit 4 and therefore you PR may not be merged.

src/main/java/junit/runner/BaseTestRunner.java
```diff
@@ -154,7 +154,8 @@ public String elapsedTimeAsString(long runTime) {
      */
     protected String processArguments(String[] args) {
         String suiteName = null;
-        for (int i = 0; i < args.length; i++) {
+        int i = 0;
+        while( i < args.length) {
```
kcooney, line 158: The previous code is the idiomatic way in Java to iterate over an array.

src/main/java/org/junit/Assert.java
```diff
@@ -603,23 +603,17 @@ public static void assertNotEquals(String message, float unexpected,
     private static boolean doubleIsDifferent(double d1, double d2, double delta) {
         if (Double.compare(d1, d2) == 0) {
             return false;
+        }else{
```
kcooney, line 606: This is slightly better. Personally I wouldn't use "else" here (unnecessary line leading to unnecessary indentation.

Note that in Java generally keywords are surrounded by a single space (we considered having auto-formatting or lint warnings for this, but that would introduce a lot of churn)

src/main/java/org/junit/internal/Throwables.java
```diff
@@ -73,7 +73,7 @@ public static String getStacktrace(Throwable exception) {
     public static String getTrimmedStackTrace(Throwable exception) {
         List<String> trimmedStackTraceLines = getTrimmedStackTraceLines(exception);
         if (trimmedStackTraceLines.isEmpty()) {
-            return getFullStackTrace(exception);
+            return Throwables.getStacktrace(exception);
```
kcooney, line 76: Thanks for deleting the redundant `getFullStackTrace()` method. No need to specify `Throwables` here since it's in the same class.

src/main/java/org/junit/internal/requests/MemoizingRequest.java
```diff
@@ -8,7 +8,7 @@
 
 abstract class MemoizingRequest extends Request {
     private final Lock runnerLock = new ReentrantLock();
-    private volatile Runner runner;
```
kcooney, line 11: The fields in JUnit that are volatile are intentionally volatile. https://www.baeldung.com/java-volatile describes a good explanation as to why we do this.

src/main/java/org/junit/runners/RuleContainer.java
```diff
@@ -45,7 +45,10 @@ public int compare(RuleEntry o1, RuleEntry o2) {
         }
 
         private int compareInt(int a, int b) {
-            return (a < b) ? 1 : (a == b ? 0 : -1);
+            if(a < b) {
```
kcooney, line 48: I personally find the previous code to be very readable.

kcooney (commented):
Thanks for your contribution.

As Marc mentioned, we aren't likely to have a new release of JUnit 4, so we aren't likely to accept contributions.

Since this work is apparently part of a class, I've left some comments that I hope are helpful.

Overall, many of your changes are stylistic, and other developers may make other stylistic choices. While I do not agree with some of the original stylistic changes in the code base, at least two people (the author and a reviewer) thought that the code was correct and readable, so I tend to leave things as-is unless there is a way to write it which is more idiomatic in Java.

In short, changing code from one stylistic choice to another introduces unnecessary churn.

In addition, since JUnt has many, many users, we need to be extremely careful when changing APIs, as those changes may break others. We've seen seemly safe changes (like renaming private fields) break people. While unused method parameters are a code smell, if the methods are externally accessible we can't change them.

Also, when making changes, provide good commit messages (sometimes it wasn't clear what code smells you were addressing). This guide on writing useful commit messages looks good: https://chris.beams.io/posts/git-commit/

There are some changes I like (changing "static public" to "public static", using Collection.isEmpty() instead of checking the size). If you'd like to send us a pull with just those changes, in two separate commits with useful commit messages, we might merge that pull (but given we won't make another release, we might not).

I suggest before sending out a set of stylistic changes to a library on Github, you reach out to the owners to discuss the problems you found and the changes you would like to introduce.

Thanks, and happy testing

Kevin

PR #1708, opened 2021-04-13
fixe some code smells

zelleg:
Vous trouverez ici la correction de plusieurs code smells dans le projet JUnit4, ce qui a été fait en détail est expliqué dans le fichier Rendu_Partie2.md

### DIFF
--- a/src/test/java/org/junit/internal/runners/statements/FailOnTimeoutTest.java
+++ b/src/test/java/org/junit/internal/runners/statements/FailOnTimeoutTest.java
@@ -4,30 +4,27 @@ import static java.lang.Long.MAX_VALUE;
 import static java.lang.Math.atan;
 import static java.lang.System.currentTimeMillis;
 import static java.lang.Thread.currentThread;
-import static java.lang.Thread.interrupted;
 import static java.lang.Thread.sleep;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertThrows;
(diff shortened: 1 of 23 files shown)


marcphilipp:
@zelleg Thank you for your contribution. JUnit 4 is a mature library and there is already a successor with JUnit 5. There is no plan for a new release of JUnit 4 and therefore you PR may not be merged.

src/test/java/org/junit/internal/runners/statements/FailOnTimeoutTest.java
```diff
@@ -37,122 +34,176 @@
  */
 @RunWith(Parameterized.class)
 public class FailOnTimeoutTest {
+    private static final long TIMEOUT = 100;
```
kcooney, line 37: These changes appear to come from 2db6394, but show up as differences, which makes it hard to review.

kcooney (commented):
Thanks for your contribution.

As Marc mentioned, we aren't likely to have a new release of JUnit 4, so we aren't likely to accept contributions.

As I indicated in some of the comments, some of your changes are stylistic, and other developers may make other stylistic choices. While I do not agree with some of the original stylistic changes in the code base, at least two people (the author and a reviewer) thought that the code was correct and readable, so I tend to leave things as-is unless there is a way to write it which is more idiomatic in Java.

In short, changing code from one stylistic choice to another introduces unnecessary churn.

In addition, since JUnt has many, many users, we need to be extremely careful when changing APIs, as those changes may break others. We've seen seemly safe changes (like renaming private fields) break people. While unused method parameters are a code smell, if the methods are externally accessible we can't change them.

There are some changes I like (changing "static public" to "public static", using Collection.isEmpty() instead of checking the size). If you'd like to send us a pull with just those changes, in two separate commits with useful commit messages in English, we might merge that pull (but given we won't make another release, we might not).
 
I suggest before sending out a set of stylistic changes to a library on Github, you reach out to the owners to discuss the problems you found and the changes you would like to introduce.

Thanks, and happy testing

- Kevin

PR #1719, opened 2021-10-06
Fix "ReturnValueIgnored" errors flagged by error-prone

kluever:


bhuvanaindukuri:
Which version of java were you able to replicate this issue? Also isn't it better to replace the line 
`e.getMessage();` with either `assertTrue(true);` or just remove the statement altogether as done in other methods like "testAssertNullNotEqualsString"

bhuvanaindukuri (changes requested):
Which version of java were you able to replicate this issue? Also isn't it better to replace the line
e.getMessage(); with either assertTrue(true); or just remove the statement altogether as done in other methods like "testAssertNullNotEqualsString"

src/test/java/junit/tests/framework/AssertTest.java
```diff
@@ -82,15 +82,15 @@ public void testAssertStringNotEqualsNull() {
             assertEquals("foo", null);
             fail();
         } catch (ComparisonFailure e) {
-            e.getMessage(); // why no assertion?
+            String unused = e.getMessage(); // why no assertion?
```
kcooney, line 85: I suggest we use `assertNotNull()` here and below.

kcooney (changes requested):

src/test/java/junit/tests/framework/AssertTest.java
kluever: Sure, done.

kluever:
> Which version of java were you able to replicate this issue?

At Google, we've enabled CheckReturnValue by default, so all non-void APIs that are not annotated with `@CanIgnoreReturnValue` must have their values used. This has caught thousands of bugs so far, and we hope to be open sourcing this soon!

kcooney (approved):

bhuvanaindukuri (approved):
Approving this change but it would be nice to make this change in other tests as well where there is nothing in the catch block.

kcooney:
@bhuvanaindukuri if you send me a pull request doing that I would approve it. Alternately we could have the code assert that the message contained a string.

Note these tests are very old. They verify the JUnit 3.x assertion code

kcooney:
Asking @marcphilipp or @sbrannen to provide a second approval, then I can merge.

sbrannen (approved):
LGTM

PR #1723, opened 2021-10-14
Improve assumeNotNull to allow custom message

kaanmertcakmak:
Added assumeNotNull method that allows custom message, just like other assertions and assumptions

After added

```
        @Test
        public void testAssumeNotNull() {
            Object[] objects = {1, 2, null};
            assumeNotNull("One of the objects is null",objects);
        }

```
Output

`org.junit.AssumptionViolatedException: One of the objects is null: got: <[1, 2, null]>, expected: every item is not null`

In my company, we use junit4 and I tried to add custom message into the assumeNotNull method, just like other assertions and assumption methods of junit4. Then I realized that, assumeNotNull only accepts objects, so I thought that this could be useful

### DIFF
--- a/src/main/java/org/junit/Assume.java
+++ b/src/main/java/org/junit/Assume.java
@@ -80,8 +80,17 @@ public class Assume {
      * the test will halt and be ignored.
      */
     public static void assumeNotNull(Object... objects) {
-        assumeThat(objects, notNullValue());
-        assumeThat(asList(objects), everyItem(notNullValue()));
+        assumeNotNull(null, objects);
+    }
+
+    /**
+     * If called with a {@code null} array or one or more {@code null} elements in {@code objects},
+     * the test will halt and be ignored.
+     * @param message A message to pass to {@link AssumptionViolatedException}.
+     */
+    public static void assumeNotNull(String message, Object... objects) {
+        assumeThat(message, objects, notNullValue());
+        assumeThat(message, asList(objects), everyItem(notNullValue()));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Hjordaan2002:
58f0889ccb32c67ad5361d32b0a7d632c2e964c3

kcooney:
This change would alter the behavior of some tests. The following test would pass in 4.13 but fail with your changes:

```
@Test
public void testPull1723() {
    String object1 = null;
    String object2 = "2";
    String object3 = "3";

    try {
        assumeNotNull(object1, object2, object3);
    } catch (AssumptionViolatedException expected) {
        return;
    }
    fail();
}
```

Why do you want custom messages for `assumeNoNull()`? The message of a `AssumptionViolatedException` doesn't matter; the exception simply indicates that the test should reported as skipped.

kaanmertcakmak:
> This change would alter the behavior of some tests. The following test would pass in 4.13 but fail with your changes:
> 
> ```
> @Test
> public void testPull1723() {
>     String object1 = null;
>     String object2 = "2";
>     String object3 = "3";
> 
>     try {
>         assumeNotNull(object1, object2, object3);
>     } catch (AssumptionViolatedException expected) {
>         return;
>     }
>     fail();
> }
> ```
> 
> Why do you want custom messages for `assumeNoNull()`? The message of a `AssumptionViolatedException` doesn't matter; the exception simply indicates that the test should reported as skipped.

Hi, 
Thanks for answer. I ran the same test and see it is failed. So in my company, we are skipping some scenarios by checking server's version on that environment and if version is higher than some number, then it should be skipped. And also there are some other scenarios that we skip tests conditionally. So displaying the reason of skip in such scenarios, would be useful for our test reports. But I don't think that there is a way to do this in such cases. So we can close this pr

PR #1729, opened 2021-12-23
Make assertEquals platform independent

liyubin117:
we use assertEquals to check the equality of objects in many scenarios, but an annoying issue is that line separator is different in unix and windows, sometimes we encounted the junit failed just only caused by platform dependent, in fact, the problem has been mentioned frequently in Stackoverflow, maybe we can fix it

image
image

### DIFF
--- a/src/main/java/org/junit/Assert.java
+++ b/src/main/java/org/junit/Assert.java
@@ -121,6 +121,13 @@ public class Assert {
         }
     }
 
+    private static Object toPlatformIndependent(Object in) {
+        if (in instanceof String) {
+            return ((String)in).replaceAll("\\n|\\r\\n", System.getProperty("line.separator"));
+        }
+        return in;
+    }
+
     private static boolean equalsRegardingNull(Object expected, Object actual) {
         if (expected == null) {
             return actual == null;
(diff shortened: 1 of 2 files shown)


panchenko:
@liyubin117 There could be different situations. In some cases there should be an exact match, in some cases the line separators are not important.
As this function has been doing the exact composition for decades - it would be better keeping that behaviour. You can implement the platform independent asserts in your code or even make it an open source library.

liyubin117:
@panchenko Thanks for your review.
maybe we can add a function `assertEqualsPlatformIndependent` to avoid inflecting running projects, also give users an choice to check equality ignore platform difference

kcooney:
It sounds like the behavior of your code that generates the expec

kcooney:
It also looks like your IDE is incorrectly showing no differences.

If the following code fails then the problem isn't JUnit:

```
if (expected == null) {
  if (actual != null) {
    throw ...
  }
} else if (expected.toSting().equals(actual.toString())) {
  throw ...
}
```

kcooney:
If you use AssertJ you can use this: http://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/AbstractCharSequenceAssert.html#isEqualToIgnoringNewLines(java.lang.CharSequence)

I don't see an equivalent in Truth.

liyubin117:
@kcooney Thanks for reminding that

PR #1731, opened 2021-12-24
Address some warnings

cpoerschke:
Used `./mvnw compile -Dmaven.compiler.showDeprecation=true` to identify the warnings.

edit: alternatively #1732 splits the changes into separate commits.

### DIFF
--- a/src/main/java/org/junit/internal/Classes.java
+++ b/src/main/java/org/junit/internal/Classes.java
@@ -21,7 +21,6 @@ public class Classes {
      * {@link Classes}.
      *
      * @param className Name of the class.
-     * @throws ClassNotFoundException
      */
     public static Class<?> getClass(String className) throws ClassNotFoundException {
         return getClass(className, Classes.class);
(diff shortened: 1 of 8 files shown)


kcooney:
Thank you for the contribution.

Note that there may not be another release of JUnit 4.x so these changes might never be released.

That being said, keeping things tidy is nice. If you would still like to get this merged would you please split it out into three commits with messages that state what is being done (fix JavaDoc warnings [listing warnings at the end of the commit message], remove usages of deprecated org.junit.Assert.assertThat(), suppress [specific warning] in...). One pull request is fine.

marcphilipp (approved):

cpoerschke:
Thanks for the feedback! #1732 is the multi-commit alternative PR; I'm not keen on `git push --force` on pull requests that are already under review, hence the separate PR rather than updating of this one here. Happy to close out whichever one of them goes unmerged.

kcooney:
Cool. Let's go forward with #1732

PR #1732, opened 2021-12-30
Address some warnings (in 4 parts)

cpoerschke:
Used `./mvnw compile -Dmaven.compiler.showDeprecation=true` to identify the warnings.

### DIFF
--- a/src/main/java/org/junit/runners/model/FrameworkMember.java
+++ b/src/main/java/org/junit/runners/model/FrameworkMember.java
@@ -12,6 +12,7 @@ public abstract class FrameworkMember<T extends FrameworkMember<T>> implements
         Annotatable {
     abstract boolean isShadowedBy(T otherMember);
 
+    @SuppressWarnings("unchecked")
     T handlePossibleBridgeMethod(List<T> members) {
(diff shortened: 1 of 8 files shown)


src/main/java/org/junit/runners/model/FrameworkMember.java
```diff
@@ -12,6 +12,7 @@
         Annotatable {
     abstract boolean isShadowedBy(T otherMember);
 
+    @SuppressWarnings("unchecked")
```
kcooney, line 15: It's preferable to suppress warnings at the smallest scope possible. I suggest:

```
	FrameworkMember<? extends T> thisMember = this;
	@SuppressWarnings("unchecked")
        T result = (T) thisMember;

        return result;
```

kcooney (changes requested):
Thanks for breaking up the changes. One minor nit. Feel free to update in the current branch so we don't need to create a new PR.

Otherwise, LGTM.

To other JUnit mainainers: if you get to this before I do, please use "Merge pull request" not "Squash and merge" to merge it.

kcooney (approved):

kcooney:
Thank you. Happy New Year!

PR #1757, opened 2023-01-22
Recommend generating Eclipse project files with Maven

kcooney:


bhuvanaindukuri (approved):
Surprising to see these files committed to the repo in the first place.

kcooney:
> Surprising to see these files committed to the repo in the first place.

@bhuvanaindukuri they date back from when JUnit was built using Ant. So it's actually not that surprising at all.

marcphilipp (approved):

PR #1761, opened 2023-03-22
ignoring unnecessarily generated surefire report

optimizing-ci-builds:
In our analysis, we observe that this target/surefire-report/ directory is generated but not later used during the CI build. Hence, we propose to disable generating this directory to save runtime.
The generation of this directory can be disabled by simply adding -DdisableXmlReport=true to the mvn command.


marcphilipp:
Thanks for the suggestion but I don't think that will have a noticeable impact on build time but slightly complicate the build setup.

PR #1771, opened 2024-02-21
Migrate off constructors of boxed primitive types.

cpovirk:
The constructors became [deprecated for
removal](https://bugs.openjdk.org/browse/JDK-8254324) in [JDK
16](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/Integer.html#%3Cinit%3E(int)).


kcooney:
Thanks!

Note there are no planned future releases of JUnit 4.x, but merging because 1) it only affects tests, and 2) this allows us to verify that things still work with future JDKs.

PR #1772, opened 2024-04-02
pom.xml: Fix build for Java 20 and above

Leo3418:
Java 20 has [dropped](https://www.oracle.com/java/technologies/javase/20-relnote-issues.html#JDK-8173605) support for 1.7 as the source and target versions.  This pull request does something similar to commit 66083734d13aa67e616f3c4b429c019c5992bba0 to allow the project to be built with e.g. JDK 21.

This is perhaps also a prerequisite of #1767.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -591,7 +591,7 @@
         <profile>
             <id>java12</id>
             <activation>
-                <jdk>[12,)</jdk>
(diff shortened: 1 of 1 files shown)


marcphilipp (approved):

marcphilipp:
@Leo3418 Thanks!

